PR #1, opened 2021-11-03
CI: Initial GitHub Action to build examples

david-cermak:
Added build jobs for building target examples for ESP32 (only) against IDF (master, `v4.4`, `v4.3`, `v4.2` and `v4.1`)


david-cermak:
@kumekay Thanks for your help with creating this initial build test! Could you please take a look at this PR?

.github/workflows/build.yml
```diff
@@ -0,0 +1,151 @@
...
+        with:
+          repository:  espressif/esp-idf
+          path: esp-idf
+          ref: release/v4.4
+
+      - name: Build-examples-with-IDF
+        run: |
+          cd $GITHUB_WORKSPACE/esp-idf
+          git submodule update --init --recursive
+          ./install.sh
```
igrr, line 57: Instead of running `install.sh` in the build job, have you considered using versioned IDF Docker images?

See for example: https://github.com/igrr/libnmea-esp32/blob/ff80ce536dfa3f1be40fa2c8c1f37d542d958cdf/.github/workflows/build.yml#L8-L17

(https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-docker-image.html)

kumekay:
Looks good to me, thank you David!

.github/workflows/build.yml
david-cermak: Thanks for the suggestion! I've rewritten the yaml based on your example.

david-cermak:
Thanks for the review and suggestions.

.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+          - idf_ver: "v4.2"
+            examples: simple_cmux_client
+
+    runs-on: ubuntu-20.04
+    container: espressif/idf:release-${{ matrix.idf_ver }}
+    steps:
+      - name: Checkout IDF
+        uses: actions/checkout@v1
+        with:
+          submodules: recursive
```
igrr, line 58: This part is probably not needed anymore, since IDF is already present in the docker image under `/opt/esp/idf`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Checkout esp-protocols
+        uses: actions/checkout@master
+        with:
+          path: esp-protocols
+      - name: Build for ESP32
```
igrr, line 63: You can use `Build for ${{ matrix.idf_target }}` here — i should fix this in libnmea-esp32 as well...

igrr (approved):

david-cermak:
Thanks again, Ivan for the review! Applied the little fixes you suggested and moved the master job to the build loop.

PR #2, opened 2021-11-08
esp_modem: Update Task handle and Event Group handle to match freertos v10.4.3

sudeep-mohanty:
This commit updates the default handles for Task type and Signal Group
type to match the struct type handle from freertos v10.4.3.

Signed-off-by: Sudeep Mohanty <sudeep.mohanty@espressif.com>


david-cermak (approved):
LGTM!

PR #4, opened 2021-11-12
CI/Docs: Add jobs to deploy component docs

david-cermak:
Added the build-docs job to:
* generate `breathe` style documentation
* deploy it to `gh-pages` branch

To be published as GitHub pages.


david-cermak:
@kumekay PTAL, example of the generated docs is available [here](https://david-cermak.github.io/esp-protocols/esp_modem/index.html)

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+          sudo apt-get update
+          sudo apt-get -y install doxygen clang python3-pip
+          python -m pip install breathe recommonmark
+          cd $GITHUB_WORKSPACE/components/esp_modem/docs
+          ./generate_docs
+          cd $GITHUB_WORKSPACE/docs
+          touch .nojekyll
+          echo '<a href="esp_modem/index.html">esp-modem</a>' > index.html
+
+      - name: Commit generated docs
```
kumekay, line 31: There is a separate action for publishing on gh-pages in one step:
https://github.com/JamesIves/github-pages-deploy-action

It should make the job a bit cleaner.

kumekay:
Docs look good to me, a short not on GitHub pages deployment

.github/workflows/build-docs.yml
david-cermak: Thanks, used the two in one action!

david-cermak:
Thanks for the suggestions, Sergei.

Also added a docs build step before the component deployment to the registry, so the documentation is generated also locally when component is downloaded.

kumekay (approved):
Thank you for update, LGTM

.github/workflows/upload_component.yml
```diff
@@ -12,10 +9,10 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Upload components to component service
-        uses: espressif/github-actions/upload_components@master
+        uses: david-cermak/github-actions/upload_components@feature/upload_component_hooks
```
kumekay, line 12: Just don't forget to revert this, before merging

david-cermak: :-) Thanks for reminder, I will. Forgot to mark it *Draft* while some WIP commits still present.

david-cermak: Done.

Did it as @igrr suggested in the referenced PR. This way it's less code and runs faster (builds the docs just once and uses it for both the `gh-pages` and the component registry), so I think it's okay as we have just one component.

kumekay (approved):

david-cermak:
Thanks for the review!

PR #8, opened 2021-11-18
esp_modem: Update FreeRTOS EventQueueHandle_t forward declaration

sudeep-mohanty:
This commit includes freertos/event_groups.h header and removes the
forward declaration for EventGroupHandle_t.

Signed-off-by: Sudeep Mohanty <sudeep.mohanty@espressif.com>


david-cermak (approved):
LGTM!

PR #9, opened 2021-11-25
fix(esp_modem): Read module name with AT commands

david-cermak:
Module names were hard-coded strings defined as `override` method for different modules.
This fixes basic IDF test case which supplied dummy module name.


david-cermak:
@lhespress PTAL

PR #10, opened 2021-12-06
 fix(esp_modem): Fix battery status parser

david-cermak:
Variable `pos`  was meant to hold position of first `,` (comma) appearance in the parsed string, but due to wrong parentheses it contained the boolean representation of not equal to `string::npos`

It wasn't caught by the tests, since the case used one-letter properties (where the expected position equals to 1/`true`)
```
+CBC: 1,2,123456
```

Updated the tests, too.

### DIFF
--- a/components/esp_modem/idf_component.yml
+++ b/components/esp_modem/idf_component.yml
@@ -1,4 +1,4 @@
-version: "0.1.12"
+version: "0.1.13"
 description: esp modem
 dependencies:
   # Required IDF version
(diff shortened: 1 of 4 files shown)


david-cermak:
@tore-espressif PTAL

tore-espressif:
LGTM!
What is the release strategy? Do you plan to bump `idf_component` version?

david-cermak:
good point, thanks. Added as a new commit.

> What is the release strategy?

I don't know yet, but will likely follow the IDF release cycle after `v5.0`

PR #14, opened 2022-01-28
fix: add virtual destructors to ModuleIf and CommandableIf

### LINKED ISSUE #13: BUG: The `ModuleIf` and `CommandableIf` do not define `virtual` destructors (IDFGH-6677)
In the following snippet, the `ModuleIf` abstract class defines an interface from which the `GenericModule` and other implementations derive. However, it does not define a `virtual` destructor. This is a bug that makes it impossible to destroy the derived object via a base class pointer.
https://github.com/espressif/esp-protocols/blob/5d27b2681a3c70c059cca3d274e04bee4791cbe0/components/esp_modem/include/cxx_include/esp_modem_types.hpp#L81-L98

Same for the `CommandableIf` abstract class.
https://github.com/espressif/esp-protocols/blob/5d27b2681a3c70c059cca3d274e04bee4791cbe0/components/esp_modem/include/cxx_include/esp_modem_types.hpp#L65-L79

b1ackviking:
Closes #13 

The PR adds explicitly defaulted virtual destructors and explicitly deleted other special member functions for the `ModuleIf` and `CommandableIf` classes to follow the [rule of 5](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five) and [this guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-virtual).

The second commit also fixes a compilation error from in the `modem_console` example with `-Werror -Wswitch-default` compile options.

### DIFF
--- a/components/esp_modem/examples/modem_console/main/httpget_handle.c
+++ b/components/esp_modem/examples/modem_console/main/httpget_handle.c
@@ -46,6 +46,7 @@ static esp_err_t http_event_handler(esp_http_client_event_t *evt)
         case HTTP_EVENT_DISCONNECTED:
             ESP_LOGI(TAG, "HTTP_EVENT_DISCONNECTED");
             break;
+        default: break;
     }
     return ESP_OK;
 }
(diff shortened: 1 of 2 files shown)


david-cermak (approved):
Thanks for the fixes!

david-cermak:
@b1ackviking Sorry it took so long to accept your PR, just wanted to make sure we don't have any undef-behavior in the current code and add the unit tests I mentioned in https://github.com/espressif/esp-protocols/pull/16

PR #16, opened 2022-02-23
add(esp_modem): Add unit test to check polymorphic delete

david-cermak:
Checks for potential undefined behaviour if we create custom device or a DTE using `new` and `delete` operators.


david-cermak:
@gabsuren PTAL

PR #17, opened 2022-02-23
Implement esp_modem_sync for the C API

mensi:
Implements sync() in the C API


david-cermak (approved):
Thanks for posting this fix! LGTM

PR #18, opened 2022-03-01
Esp websocket client migration with history

gabsuren:


docs/en/api-reference/protocols/esp_websocket_client.rst
```diff
@@ -0,0 +1,127 @@
+ESP WebSocket Client
```
david-cermak, line 1: Shouldn't this be located under `components/esp_websocket_client/docs` ?

david-cermak (changes requested):

david-cermak (approved):
Thanks @suren-gabrielyan-espressif  for adding this component!

PR #20, opened 2022-03-04
build: fix issue with passing cxx_std option, a common workaround (IDFGH-6876)

antmak:
issue: https://gitlab.kitware.com/cmake/cmake/-/issues/23297

It should work since CMake 3.1 version (https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

### DIFF
--- a/components/esp_modem/CMakeLists.txt
+++ b/components/esp_modem/CMakeLists.txt
@@ -34,16 +34,11 @@ idf_component_register(SRCS "${srcs}"
                     PRIV_INCLUDE_DIRS private_include
                     REQUIRES ${dependencies})
 
-
-# If CMake doesn't know C++17 features, set it manually via comile options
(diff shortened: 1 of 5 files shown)


antmak:
Another way like
```cmake
if (NOT "cxx_std_17" IN_LIST cxx_known_features
    OR (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"
        AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "11.0.0"
       )
    )
    target_compile_options(${COMPONENT_LIB} PRIVATE "-std=gnu++17")
else()
    target_compile_features(${COMPONENT_LIB} PRIVATE cxx_std_17)
endif()
```
doesn't make sense for me

antmak:
@david-cermak should we bump the component version?

david-cermak (approved):

PR #21, opened 2022-03-16
Fix IDFv5.0 include paths (IDFGH-6972)

tore-espressif:
`esp_system.h` does not longer include `esp_mac.h` so it must be included explicitly.

See https://github.com/espressif/esp-idf/commit/a9fda54d39d1321005c3bc9b3cc268d0b7e9f052


tore-espressif:
@david-cermak PTAL

david-cermak (approved):

PR #25, opened 2022-03-17
Add USB DTE example

tore-espressif:
This PR extends `modem_console` example with USB DTE on supported targets (ESP32-S2/3) and IDF versions (>=4.4).

Added `esp_modem_usb_dte` component mimics structure of original `esp_modem`.

Tested on ESP32-S2 and Quactel BG96 modem.

---
This PR also contains #21 so the CI would pass


david-cermak (approved):
LGTM

PR #28, opened 2022-04-07
esp_modem fix for SIM800 modem devices

bitron:
This PR is meant to fix an issue experienced with the ‘pppos_client’ example in the ‘esp_modem’ library running on an ESP32 connected to a SIM800C module.

The ‘pppos_client’ example of the ‘esp_modem’ library was not working because of two reasons:

a) the AT command to start the PPP communication in the ‘esp_modem’ library was different from the one in the SIM800 Series AT Command Manual
b) the ‘pppos_client’ example was ignoring the type of modem device set through menuconfig

This fix should be compatible with all ESP32 chips and SIM800 series modules.


david-cermak (approved):

david-cermak:
Thanks for the fix @bitron !

bitron:
You're welcome @david-cermak !

PR #29, opened 2022-04-15
esp_websocket_client: Expose frame fin flag in websocket event

MacDue:
Follow up from https://github.com/espressif/esp-idf/issues/8367 and https://github.com/espressif/esp-idf/pull/8395 :)


david-cermak (approved):
Thanks for this additional PR on the websocket FIN flag!

PR #30, opened 2022-04-20
Feature/extended simcom support (IDFGH-7243)

Sjurinho:
## Changes
- Added lists to esp_modem_command_declare.inc
- Extended command library to include many commonly used AT commands (using simcom modules as a base)
- Added SIM7070, SIM7000 and extended SIM7600 classes
  - Tested both C and C++ APIs
- Changed initial modem_mode to be set to UNDEF
  - In case of mcu restart without modem restart, which could make the modem be stuck in data mode

### DIFF
--- a/components/esp_modem/include/generate/esp_modem_command_declare.inc
+++ b/components/esp_modem/include/generate/esp_modem_command_declare.inc
@@ -28,7 +28,7 @@
 #define BOOL_IN(param, name) const bool  _ARG(param, name)
 #define BOOL_OUT(param, name) bool&  _ARG(param, name)
 #define INT_OUT(param, name) int&  _ARG(param, name)
-#define LIST_IN(param, name) const int* _ARG(param, name)
+#define INTEGER_LIST_IN(param, name) const int* _ARG(param, name)
 
 #define STRUCT_OUT(struct_name, p1)  struct_name& p1
 #else
@@ -37,7 +37,7 @@
 #define BOOL_IN(param, name) const bool _ARG(param, name)
 #define BOOL_OUT(param, name) bool* _ARG(param, name)
 #define INT_OUT(param, name) int* _ARG(param, name)
-#define LIST_IN(param, name) const int* _ARG(param, name)
+#define INTEGER_LIST_IN(param, name) const int* _ARG(param, name)
 #define STRUCT_OUT(struct_name, p1)  struct struct_name* p1
 #endif
 
@@ -252,11 +252,6 @@ ESP_MODEM_DECLARE_DCE_COMMAND(get_radio_state, command_result, 1, INT_OUT(p1, st
 /**
  * @brief Set network mode
  * @param[in] mode preferred mode
- * -- SIM7xxx --
- * mode=2 - Automatic
- * mode=13 - GSM only
- * mode=38 - LTE only
- * mode=51 - GSM and LTE only
  * @return OK, FAIL or TIMEOUT
  */ \
 ESP_MODEM_DECLARE_DCE_COMMAND(set_network_mode, command_result, 1, INT_IN(p1, mode)) \
@@ -273,7 +268,7 @@ ESP_MODEM_DECLARE_DCE_COMMAND(set_preferred_mode, command_result, 1, INT_IN(p1, 
  * @param[in] mode CAT-M or NB-IoT
  * @return OK, FAIL or TIMEOUT
  */ \
-ESP_MODEM_DECLARE_DCE_COMMAND(set_network_bands, command_result, 3, STRING_IN(p1, mode), LIST_IN(p2, bands), INT_IN(p3, size)) \
+ESP_MODEM_DECLARE_DCE_COMMAND(set_network_bands, command_result, 3, STRING_IN(p1, mode), INTEGER_LIST_IN(p2, bands), INT_IN(p3, size)) \
     \
 /**
  * @brief Show network system mode
--- a/components/esp_modem/include/generate/esp_modem_command_declare.inc
+++ b/components/esp_modem/include/generate/esp_modem_command_declare.inc
@@ -28,6 +28,7 @@
 #define BOOL_IN(param, name) const bool  _ARG(param, name)
 #define BOOL_OUT(param, name) bool&  _ARG(param, name)
 #define INT_OUT(param, name) int&  _ARG(param, name)
+#define LIST_IN(param, name) const int* _ARG(param, name)
 
 #define STRUCT_OUT(struct_name, p1)  struct_name& p1
 #else
@@ -36,6 +37,7 @@
 #define BOOL_IN(param, name) const bool _ARG(param, name)
 #define BOOL_OUT(param, name) bool* _ARG(param, name)
 #define INT_OUT(param, name) int* _ARG(param, name)
+#define LIST_IN(param, name) const int* _ARG(param, name)
 #define STRUCT_OUT(struct_name, p1)  struct struct_name* p1
 #endif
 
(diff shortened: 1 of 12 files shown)


components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -18,6 +18,7 @@
 #include "cxx_include/esp_modem_dte.hpp"
 #include "cxx_include/esp_modem_dce_module.hpp"
 #include "cxx_include/esp_modem_command_library.hpp"
+#include <iomanip>
```
david-cermak, line 21: Nitpick: Please move up, next to the standard lib `#include <>`'s

components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -399,4 +406,138 @@ command_result get_signal_quality(CommandableIf *t, int &rssi, int &ber)
...
+// any_mode = "0xFFFFFFFF7FFFFFFF";
+command_result set_network_bands_sim76xx(CommandableIf *t, const std::string& mode, const int* bands, int size)
+{
+    ESP_LOGV(TAG, "%s", __func__ );
+    uint64_t band_bits = 0;
+    for (int i = 0; i<size; ++i) {
+        // OR-operation to add bands
+        band_bits |= 1 << bands[i];
+    }
+    std::stringstream stream;
```
david-cermak, line 503: Could you please check if using the `stringstream` doesn't have any big influence on flash size or static memory usage? I recall we had some issues in the past when using C++ *streams* on ESP platform.

components/esp_modem/include/generate/esp_modem_command_declare.inc
```diff
@@ -36,6 +37,7 @@
 #define BOOL_IN(param, name) const bool _ARG(param, name)
 #define BOOL_OUT(param, name) bool* _ARG(param, name)
 #define INT_OUT(param, name) int* _ARG(param, name)
+#define LIST_IN(param, name) const int* _ARG(param, name)
```
david-cermak, line 40: Could we make the `LIST_IN` type array instead of ptr? (we should also call in *integer-list* or similar...)

david-cermak (approved):
Nice work, thanks! I don't have the devices to test your changes, but the PR looks very good to me!

david-cermak:
Please check the unit test failure in https://github.com/espressif/esp-protocols/runs/6170758917?check_suite_focus=true

components/esp_modem/include/generate/esp_modem_command_declare.inc
Sjurinho: We probably could, I just wasn't sure exactly how to navigate the _ARG definitions. Although I think it might be better to use pointers and rather have _ARG take in pointer + size?

Sjurinho:
@david-cermak Could you elaborate where the host test failed? 
I see that the last console print was during the CMUX test, but as far as I can see I haven't altered the CMUX part of the code anywhere, except for changing the initial mode to be `UNDEF`

components/esp_modem/src/esp_modem_command_library.cpp
david-cermak: I would just check the static sizes using [idf.py size/size-files](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/performance/size.html#size-summary-idf-py-size) before and after your changes (i.e. using the stream)
I don't think the actual creation of the `stringstream` would have a dramatic impact on heap (was just concerned about the statics)

components/esp_modem/include/generate/esp_modem_command_declare.inc
david-cermak: Would this work?
```suggestion
#define LIST_IN(param, name) const int _ARG(param[], name[])
```
it still uses pointer and sizes, but is syntactically closer to an array.

but very minor point, feel free to resolve.

david-cermak:
> @david-cermak Could you elaborate on where the host test failed?

I don't see much in the logs either. I'm sorry, will have to improve the error reporting in the CI!

Running it locally makes it clear:
```
-------------------------------------------------------------------------------
DCE modes
-------------------------------------------------------------------------------
../main/test_modem.cpp:120
...............................................................................

../main/test_modem.cpp:131: FAILED:
  CHECK( dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == false )
with expansion:
  true == false
```

so your assumption about the initial state was correct. This change makes the unit test pass:
```patch
diff --git a/components/esp_modem/test/host_test/main/test_modem.cpp b/components/esp_modem/test/host_test/main/test_modem.cpp
index f2305e9..ff46808 100644
--- a/components/esp_modem/test/host_test/main/test_modem.cpp
+++ b/components/esp_modem/test/host_test/main/test_modem.cpp
@@ -128,7 +128,7 @@ TEST_CASE("DCE modes", "[esp_modem]")
     auto dce = create_SIM7600_dce(&dce_config, dte, &netif);
     CHECK(dce != nullptr);
 
-    CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == false);
+    CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == true);
     CHECK(dce->set_mode(esp_modem::modem_mode::DATA_MODE) == true);
     CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == true);
 }
```

diplfranzhoepfinger:
i want to test this ! i have a SIM7070 !

cj8scrambler (commented):
I was able to verify this on a SIM7000G (LilyGo board).  This solution is much more elegant than the hacks I was trying to wedge in.  Just a few documentation suggestions.

Sjurinho:
Hey @cj8scrambler Thanks for the feedback. Haven't had the time to work on this for a while, but hopefully will by the end of this week. Can add the notes, however they are related to the specific modem and I am therefore not sure if it is helpful as they might change depending on what modem one uses.

Sjurinho:
> i want to test this ! i have a SIM7070 !

Yeah, that would be great. I have tested it, but don't hesitate to report any errors/unclarities you might stumble upon:)

Sjurinho:
> > @david-cermak Could you elaborate on where the host test failed?
> 
> I don't see much in the logs either. I'm sorry, will have to improve the error reporting in the CI!
> 
> Running it locally makes it clear:
> 
> ```
> -------------------------------------------------------------------------------
> DCE modes
> -------------------------------------------------------------------------------
> ../main/test_modem.cpp:120
> ...............................................................................
> 
> ../main/test_modem.cpp:131: FAILED:
>   CHECK( dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == false )
> with expansion:
>   true == false
> ```
> 
> so your assumption about the initial state was correct. This change makes the unit test pass:
> 
> ```diff
> diff --git a/components/esp_modem/test/host_test/main/test_modem.cpp b/components/esp_modem/test/host_test/main/test_modem.cpp
> index f2305e9..ff46808 100644
> --- a/components/esp_modem/test/host_test/main/test_modem.cpp
> +++ b/components/esp_modem/test/host_test/main/test_modem.cpp
> @@ -128,7 +128,7 @@ TEST_CASE("DCE modes", "[esp_modem]")
>      auto dce = create_SIM7600_dce(&dce_config, dte, &netif);
>      CHECK(dce != nullptr);
>  
> -    CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == false);
> +    CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == true);
>      CHECK(dce->set_mode(esp_modem::modem_mode::DATA_MODE) == true);
>      CHECK(dce->set_mode(esp_modem::modem_mode::COMMAND_MODE) == true);
>  }
> ```

Did you want me to change the tests in this PR, @david-cermak ?

components/esp_modem/include/generate/esp_modem_command_declare.inc
Sjurinho: If we want to make it more general, perhaps we could make it a `const void*` ?

david-cermak: > If we want to make it more general, perhaps we could make it a const void* ? 

I personally prefer concrete types, so the list of integers looks good to me. Resolving, as I'm finding this okay.

Just to explain why I commented:
1) Types `int *` and `int []` are *almost* the same, but the latter suggest that it's an array (while the former is just a pointer)
2) Naming it `LIST_IN()` may suggest any kind of list, but it is a list of **integers** per your definition.

(Optionally you can just rename it to `INT_LIST_IN()` if that sounds good to you)

david-cermak:
> Did you want me to change the tests in this PR, @david-cermak ?

Yes, please. Would be great If you can make it in a separate commit. The reason is that you've actually changed the state-machine, i.e. that the modem is in a different state after init. This change makes perfect sense, but we'd have to update the basic behavioral tests, too.

components/esp_modem/src/esp_modem_command_library.cpp
diplfranzhoepfinger: hi, i can confirm that the used Flash size siginificantly increase. 
before i got no Message about Partition too small, now i get one.

diplfranzhoepfinger: image

diplfranzhoepfinger: before: 

image

components/esp_modem/include/generate/esp_modem_command_declare.inc
```diff
@@ -200,7 +202,93 @@ ESP_MODEM_DECLARE_DCE_COMMAND(reset, command_result, 0) \
...
+ * mode=13 - GSM only
+ * mode=38 - LTE only
+ * mode=51 - GSM and LTE only
+ * @return OK, FAIL or TIMEOUT
+ */ \
+ESP_MODEM_DECLARE_DCE_COMMAND(set_network_mode, command_result, 1, INT_IN(p1, mode)) \
+    \
+/**
+ * @brief Preferred network mode (CAT-M and/or NB-IoT)
+ * @param[in] mode preferred selection
```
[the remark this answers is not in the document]
diplfranzhoepfinger: this should be really in the Sourcecode or in the Documantation.

components/esp_modem/include/generate/esp_modem_command_declare.inc
```diff
@@ -200,7 +202,93 @@ ESP_MODEM_DECLARE_DCE_COMMAND(reset, command_result, 0) \
...
+ * @return OK, FAIL or TIMEOUT
+ */ \
+ESP_MODEM_DECLARE_DCE_COMMAND(set_preferred_mode, command_result, 1, INT_IN(p1, mode)) \
+    \
+/**
+ * @brief Set network bands for CAT-M or NB-IoT
+ * @param[in] mode CAT-M or NB-IoT
```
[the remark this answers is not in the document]
diplfranzhoepfinger: EGSM_MODE
DCS_MODE
ALL_MODE

Sjurinho:
It took a while, but I've finally come around to update the code. I've removed stringstream and renamed the list-in parameter to be integer_list_in, @david-cermak @diplfranzhoepfinger . I agree with @diplfranzhoepfinger that the documentation should not mention modem-specific comments, so I've removed those, and will not be adding it. Thank you to everyone who has helped test the implementation

components/esp_modem/include/generate/esp_modem_command_declare.inc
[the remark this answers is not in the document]
Sjurinho: Yes, that is correct.

david-cermak:
@Sjurinho Thanks for the updates! Could you please click though the [CLA assistant](https://github.com/espressif/esp-protocols/pull/30#issuecomment-1111176354)? Need your agreement to accept the PR.

Sjurinho:
I believe I have already, @david-cermak

david-cermak (approved):

david-cermak:
> Can we get the package version bumped so this feature is available?

Good point, thanks. we still have some CMUX related issues, so let me quickly fix those and I'll bump the version after that.

>  Or is there some way to tell idf to pull this package from HEAD?

You can use
* `^1.0.0` notation to get the latest package with `MAJOR_VERSION==1`
* `~0.1.0` notation to get the latest package starting with `0.1`

franz-ms-muc:
can you tell a Status about the "some CMUX related issues" ?

david-cermak:
> can you tell a Status about the "some CMUX related issues" ?

turns out to be multiple related issues, that would require some minor refactor.

I would try to complete at least this https://github.com/espressif/esp-protocols/pull/53 simple fix and will update the version. Will do by the end of this week.

Issue #33, opened 2022-05-06
Unable to use the CMUX example (with BG96) (IDFGH-7328)


jonathandreyer:
Hi everyone,

I have tried to make the *simple_cmux_client* example work, but that is not successful. The log is (with log set as verbose):
```
.
..
...
D (888) event: running task for loop 0x3fce0814
D (898) event: created task for loop 0x3fce0814
D (898) event: created event loop 0x3fce0814
D (908) esp_netif_lwip: LwIP stack has been initialized
D (908) esp_netif_lwip: esp-netif has been successfully initialized
I (918) uart: queue free spaces: 30
V (918) intr_alloc: esp_intr_alloc_intrstatus (cpu 0): checking args
V (918) intr_alloc: esp_intr_alloc_intrstatus (cpu 0): Args okay. Resulting flags 0xE
D (918) intr_alloc: Connected src 28 to int 13 (cpu 0)
D (938) esp_netif_objects: esp_netif_add_to_list 0x3fce5b84
D (948) esp_netif_objects: esp_netif_add_to_list netif added successfully (total netifs: 1)
D (958) esp-netif_lwip-ppp: esp_netif_new_ppp: PPP connection created: 0x3fce5eb4
D (958) esp-netif_lwip-ppp: Phase Dead
V (968) command_lib: set_cmux
V (968) command_lib: generic_command_common
V (968) command_lib: generic_command
D (978) command_lib: generic_command command AT+CMUX=0

D (988) command_lib: Response: 
OK


V (2008) intr_alloc: esp_intr_free: Disabling int, killing handler
E (2008) cmux_example: Failed to configure desired mode... exiting
D (2008) event: deleted loop 0x3fce0814
```

As the problem seems to come from the line (below), I have also tried to separate both actions. From:
https://github.com/espressif/esp-protocols/blob/38149c8d9bdf469806fe0685558eba74f3366d73/components/esp_modem/examples/simple_cmux_client/main/simple_cmux_client_main.cpp#L90

to:
```
bool cmux_mode = dce->set_mode(esp_modem::modem_mode::CMUX_MODE);
ESP_LOGD(TAG, "set_mode in CMUX : %d", cmux_mode);
bool data_mode = dce->set_mode(esp_modem::modem_mode::DATA_MODE);
ESP_LOGD(TAG, "set_mode in DATA : %d", data_mode);

if (cmux_mode && data_mode) {
```

In this case, the log is:
```
.
..
...
[...]

PR #34, opened 2022-05-10
fix(esp_modem): Correction of switching to CMUX mode

david-cermak:
* BG96 needs a small pause between CMUX command reception and actual use
of the CMUX mode
* Correct CMUX payload reception to check FT_UIH message type (P/F flag
could be cleared or set)

Closes https://github.com/espressif/esp-protocols/issues/33


david-cermak:
@lhespress PTAL, too

gabsuren (approved):

PR #35, opened 2022-05-12
feat(websocket): Optimize memory size for websocket client init (IDFGH-7357)

ESP-YJM:
- Remove the unused transport. For example, if the connection is Websocket over TCP, we only need ws and tcp transport, no need wss and ssl transport.
- Make the tx buffer and rx buffer malloc dynamic.

This can save about 3 KB memory when no data send and receive.

### DIFF
--- a/components/esp_websocket_client/Kconfig
+++ b/components/esp_websocket_client/Kconfig
@@ -1,0 +1,10 @@
+menu "ESP WebSocket client"
+
+    config ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER
+        bool "Enable websocket client dynamic buffer for send and receive data"
+        default n
+        help
+            Enable this option will reallocated buffer when send or receive data and free them when end of use.
+            This can save about 2 KB memory when no websocket data send and receive.
+
+endmenu
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -54,6 +54,9 @@ static const char *TAG = "WEBSOCKET_CLIENT";
         action;                                                                                     \
         }
 
+#define WS_OVER_TCP_SCHEME  "ws"
+#define WS_OVER_TLS_SCHEME  "wss"
+
 const static int STOPPED_BIT = BIT0;
 const static int CLOSE_FRAME_SENT_BIT = BIT1;   // Indicates that a close frame was sent by the client
                                         // and we are waiting for the server to continue with clean close
@@ -78,6 +81,15 @@ typedef struct {
     char                        *headers;
     int                         pingpong_timeout_sec;
     size_t                      ping_interval_sec;
+    const char                  *cert;
+    size_t                      cert_len;                
+    const char                  *client_cert;
+    size_t                      client_cert_len;
+    const char                  *client_key;
+    size_t                      client_key_len;
+    bool                        use_global_ca_store;
+    bool                        skip_cert_common_name_check;
+    esp_err_t                   (*crt_bundle_attach)(void *conf);
 } websocket_config_storage_t;
 
 typedef enum {
@@ -120,6 +132,45 @@ struct esp_websocket_client {
 static uint64_t _tick_get_ms(void)
 {
     return esp_timer_get_time()/1000;
+}
+
+static esp_err_t esp_websocket_new_buf(esp_websocket_client_handle_t client, bool is_tx)
+{
+#ifdef CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER
+    if (is_tx) {
+        if (client->tx_buffer) {
+            free(client->tx_buffer);
+        }
+
+        client->tx_buffer = calloc(1, client->buffer_size);
+        ESP_WS_CLIENT_MEM_CHECK(TAG, client->tx_buffer, return ESP_ERR_NO_MEM);
+    } else {
+        if (client->rx_buffer) {
+            free(client->rx_buffer);
+        }
+
+        client->rx_buffer = calloc(1, client->buffer_size);
+        ESP_WS_CLIENT_MEM_CHECK(TAG, client->rx_buffer, return ESP_ERR_NO_MEM);
+    }
+#endif
+    return ESP_OK;
+}
+
+static void esp_websocket_free_buf(esp_websocket_client_handle_t client, bool is_tx)
+{
+#ifdef CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER
+    if (is_tx) {
+        if (client->tx_buffer) {
+            free(client->tx_buffer);
+            client->tx_buffer = NULL;
+        }
+    } else {
+        if (client->rx_buffer) {
+            free(client->rx_buffer);
+            client->rx_buffer = NULL;
+        }
+    }
+#endif
 }
 
 static esp_err_t esp_websocket_client_dispatch_event(esp_websocket_client_handle_t client,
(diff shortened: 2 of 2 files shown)


ESP-YJM:
@david-cermak @lhespress PTAL.

david-cermak:
@ESP-YJM Thanks for the improvements! Do you think we can make the dynamic buffer configurable? While this is a perfect option for some user apps, others (that communicate over ws more frequently) may prefer one time allocation.

@euripedesrocha PTAL, too.

ESP-YJM:
> @ESP-YJM Thanks for the improvements! Do you think we can make the dynamic buffer configurable? While this is a perfect option for some user apps, others (that communicate over ws more frequently) may prefer one time allocation.
> 
> @euripedesrocha PTAL, too.

Hi David, it's a good ideal that make dynamic buffer configurable. I have add the option. PTAL.

components/esp_websocket_client/Kconfig
```diff
@@ -0,0 +1,9 @@
+menu "ESP WebSocket client"
+
+    config ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER
+        bool "Enable websocket client dynamic buffer for tx/rx"
+        default n
+        help
+            This option will enable dynamic buffer malloc for tx/rx, it can save about tx and rx memory when no websocket data send and receive.
```
euripedesrocha, line 7: I suggest that we add that the buffer is reallocated for each use, since this is relevant to our context.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -78,6 +81,14 @@ typedef struct {
     char                        *headers;
     int                         pingpong_timeout_sec;
     size_t                      ping_interval_sec;
+    const char                  *cert_pem;
```
euripedesrocha, line 84: Rename it to remove the `_pem` part, since It's possible to have certificates in DER format.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
...
+            if (client->keep_alive_cfg.keep_alive_enable) {
+                esp_transport_tcp_set_keep_alive(tcp, &client->keep_alive_cfg);
+            }
+            if (client->if_name) {
+                esp_transport_tcp_set_interface_name(tcp, client->if_name);
+            }
+
+            esp_transport_handle_t ws = esp_transport_ws_init(tcp);
+            ESP_WS_CLIENT_MEM_CHECK(TAG, ws, {
+                esp_transport_destroy(tcp);
```
euripedesrocha, line 393: Must call `esp_transport_list_destroy` to destroy both `tcp` and the list.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
...
+                    esp_transport_ssl_set_client_key_data_der(ssl, client->config->client_key, client->config->client_key_len);
+                }
+            }
+            if (client->config->skip_cert_common_name_check) {
+                esp_transport_ssl_skip_common_name_check(ssl);
+            }
+
+            esp_transport_handle_t wss = esp_transport_ws_init(ssl);
+            ESP_WS_CLIENT_MEM_CHECK(TAG, wss, {
+                esp_transport_destroy(ssl);
```
euripedesrocha, line 435: Must call `esp_transport_list_destroy`.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
...
+                return ESP_ERR_NO_MEM;
+            });
+
+            esp_transport_set_default_port(wss, WEBSOCKET_SSL_DEFAULT_PORT);
+
+            esp_transport_list_add(client->transport_list, wss, WS_OVER_TLS_SCHEME);
+            ESP_WS_CLIENT_ERR_OK_CHECK(TAG, set_websocket_transport_optional_settings(client, WS_OVER_TLS_SCHEME), return ESP_FAIL;)
+        } else {
```
euripedesrocha, line 443: Nitpick: Since you are returning on error, you could have handled this cases first and return avoiding the else.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
...
+            ESP_LOGE(TAG, "Not support this websocket scheme %s", client->config->scheme);
```
euripedesrocha, line 444: Can we point to available schemes on the error message?

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -122,6 +133,52 @@ static uint64_t _tick_get_ms(void)
     return esp_timer_get_time()/1000;
 }
 
+static esp_err_t esp_websocket_setup_rx_buf(esp_websocket_client_handle_t client)
```
euripedesrocha, line 136: Check for a valid client in all of these functions.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
     return ESP_ERR_INVALID_ARG;
 }
 
+static esp_err_t esp_websocket_client_create_transport(esp_websocket_client_handle_t client)
+{
+    if (client->transport_list) {
```
euripedesrocha, line 370: Check for a valid client

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -308,6 +365,92 @@ static esp_err_t set_websocket_transport_optional_settings(esp_websocket_client_
...
+
+            esp_transport_set_default_port(wss, WEBSOCKET_SSL_DEFAULT_PORT);
+
+            esp_transport_list_add(client->transport_list, wss, WS_OVER_TLS_SCHEME);
+            ESP_WS_CLIENT_ERR_OK_CHECK(TAG, set_websocket_transport_optional_settings(client, WS_OVER_TLS_SCHEME), return ESP_FAIL;)
+        } else {
+            ESP_LOGE(TAG, "Not support this websocket scheme %s", client->config->scheme);
+            return ESP_FAIL;
+        }
+    } else {
```
euripedesrocha, line 447: Is it possible to have scheme not set for a client that was initialized?

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -122,6 +133,52 @@ static uint64_t _tick_get_ms(void)
...
+        free(client->rx_buffer);
+    }
+
+    client->rx_buffer = calloc(1, client->buffer_size);
+    ESP_WS_CLIENT_MEM_CHECK(TAG, client->rx_buffer, return ESP_ERR_NO_MEM);
+#endif
+    return ESP_OK;
+}
+
+static void esp_websocket_free_rx_buf(esp_websocket_client_handle_t client)
```
euripedesrocha, line 149: We could merge the `setup` and `free` functions and let the call site to select the buffer:
```cpp
esp_websocket_free_buffer(client->rx_buffer)
```

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: I think this is an internal function and the client must not be NULL. So the check could be ignored.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: I think this is an internal function and the client must not be NULL. So the check could be ignored.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Good idea.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Thanks for pointing out.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: I think we can return `ESP_ERR_NO_MEM` directly and `esp_websocket_client_destory` will call `esp_transport_list_destroy` to free all the transport that added to list.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: The same as https://github.com/espressif/esp-protocols/pull/35#discussion_r874290728

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Yes, i will change it.

components/esp_websocket_client/Kconfig
ESP-YJM: Sorry, I didn't understand what you meant, can you share your thoughts in detail? Do you mean make tx/rx buffer size seperate?

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Yeah, we could merge setup and free to one function, but i think seperate them will make sense.

components/esp_websocket_client/Kconfig
euripedesrocha: No, I mean to clarify what "dynamic buffer" mean on operation.

components/esp_websocket_client/esp_websocket_client.c
euripedesrocha: I'm not sure if we are talking about the same thing :) . What I meant was to have:
`esp_websocket_free_buffer` and `esp_websocket_setup_buffer`.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -122,6 +133,52 @@ static uint64_t _tick_get_ms(void)
...
+{
+#ifdef CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER
+    if (client->rx_buffer) {
+        free(client->rx_buffer);
+        client->rx_buffer = NULL;
+    }
+#endif
+}
+
+static esp_err_t esp_websocket_setup_tx_buf(esp_websocket_client_handle_t client)
```
euripedesrocha, line 159:
```suggestion
static esp_err_t esp_websocket_setup_new_buf(esp_websocket_client_handle_t client)
```

components/esp_websocket_client/Kconfig
ESP-YJM: Ok, i have changed it. Please help review it again. Thanks.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Yes, i have changed it.

components/esp_websocket_client/esp_websocket_client.c
ESP-YJM: Thanks for your reviewing. I have change it. PTAL.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -122,6 +133,45 @@ static uint64_t _tick_get_ms(void)
     return esp_timer_get_time()/1000;
 }
 
+static esp_err_t esp_websocket_new_buf(esp_websocket_client_handle_t client, bool is_tx)
```
euripedesrocha, line 136: We can avoid the boolean if we have in the call site:
```cpp
esp_websocket_new_buf(client->tx_buffer);
esp_websocket_new_buf(client->rx_buffer);
```

ESP-YJM: We need `client` handle to know `client->buffer_size` value to allocate property size memory. Only tx_buffer or rx_buffer point is not enough.

euripedesrocha (approved):

david-cermak (approved):

david-cermak:
@euripedesrocha Thanks for the review!

@ESP-YJM Could you please rebase the PR? It picked up a conflict.

ESP-YJM:
> @euripedesrocha Thanks for the review!
> 
> @ESP-YJM Could you please rebase the PR? It picked up a conflict.

Sure. I have rebased and solved the conflict.

david-cermak:
@ESP-YJM Thanks for the update and for your help in addressing memory usage!

PR #49, opened 2022-06-02
feat(websocket): allow users to attach CA bundle (IDFGH-7520)

### LINKED ISSUE #48: esp_websocket_client does not allow users to include bundled CA certificates (IDFGH-7519)
I would like to use the bundled CA certificates in TLS WebSocket. The current code does not include `esp_transport_ssl_crt_bundle_attach()` in the code. other protocol clients, such as `esp_http_client`, have a member of function pointer to pass to `esp_transport_ssl_crt_bundle_attach()`.

`esp_websocket_client` should do the same.

trombik:
with esp_transport_ssl_crt_bundle_attach(), it is not necessary to
include a specific certificate. other protocol clients, such as
esp_http_client, do the same.

fixes #48

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -392,6 +392,15 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
             esp_transport_ssl_set_client_key_data_der(ssl, config->client_key, config->client_key_len);
         }
     }
+
+    if (config->crt_bundle_attach) {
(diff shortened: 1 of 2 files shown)


trombik:
is it possible to back-port this PR to version 4.x?

david-cermak (approved):
Thanks for adding the support for CA bundle!

david-cermak:
> is it possible to back-port this PR to version 4.x?

@trombik Thanks for the contribution. Yes, we can backport this to IDF `v4.4` (but not to `v4.3`)

PR #51, opened 2022-06-03
Minor fixes here and there

gabsuren:
1. mDNS: Added idf_component.yml
2. mDNS: Updated document generation paths
2. Websocket: Updated conf_common.py to use internal
3. Websocket: Updated CmakeList.txt to use internal


david-cermak (approved):
Thanks @gabsuren for the fixes!

Issue #46, opened 2022-05-30
About the problem of CMUX protocol mismatch (IDFGH-7490)


hqs1:
Hello, this warning will be reported when running the esp_modem_cmux example，I don't know if it's a cache size issue.
module:ESP32 S2 WROVER
4G module:合宙Air724UG

**W (7798) CMUX: Protocol mismatch: Missed leading SOF, recovering...**

> ESP-ROM:esp32s2-rc4-20191025
Build:Oct 25 2019
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3ffe6100,len:0x1768
load:0x4004c000,len:0xa48
load:0x40050000,len:0x2e58
entry 0x4004c204
I (21) boot: ESP-IDF v4.4.1 2nd stage bootloader
I (21) boot: compile time 13:03:43
I (21) boot: chip revision: 0
I (24) boot.esp32s2: SPI Speed      : 80MHz
I (29) boot.esp32s2: SPI Mode       : DIO
I (33) boot.esp32s2: SPI Flash Size : 4MB
I (38) boot: Enabling RNG early entropy source...
I (43) boot: Partition Table:
I (47) boot: ## Label            Usage          Type ST Offset   Length
I (54) boot:  0 nvs              WiFi data        01 02 00009000 00004000
I (62) boot:  1 otadata          OTA data         01 00 0000d000 00002000
I (69) boot:  2 phy_init         RF data          01 01 0000f000 00001000
I (77) boot:  3 factory          factory app      00 00 00010000 00100000
I (84) boot:  4 ota_0            OTA app          00 10 00110000 00100000
I (92) boot:  5 ota_1            OTA app          00 11 00210000 00100000
I (99) boot: End of partition table
I (103) boot: Defaulting to factory image
I (108) esp_image: segment 0: paddr=00010020 vaddr=3f000020 size=25924h (153892) map
I (147) esp_image: segment 1: paddr=0003594c vaddr=3ffbe710 size=01e04h (  7684) load
I (149) esp_image: segment 2: paddr=00037758 vaddr=40022000 size=088c0h ( 35008) load
I (162) esp_image: segment 3: paddr=00040020 vaddr=40080020 size=7056ch (460140) map
I (254) esp_image: segment 4: paddr=000b0594 vaddr=4002a8c0 size=03e48h ( 15944) load
[...]

PR #53, opened 2022-06-06
esp_modem: Support 2 byte size packets

david-cermak:
Support CMUX with extended (2-byte coded) payload.

Closes https://github.com/espressif/esp-protocols/issues/46


david-cermak:
Thanks @euripedesrocha for the review!

Issue #59, opened 2022-06-17
Suggestion for custom commands (IDFGH-7621)


MisterScience5:
I am currently implementing the modem protocol on a SIM7070, and in my use case I would like for it to use the modems built in feature to check received APN on network connection, however this is not available in the library. I have been trying to add a custom command, but I am having some trouble with it. Therefore it would be nice if the library exposed a simple "custom command" that sends a desired string as a command, and simply returns the received data in a string, something like:

char return_string[50];
esp_modem_send_custom_command(dce, "AT+CGNAPN", &return_string);

Which would return for example "+CGNAPN: 1,telenor.iot".

This would make running a few custom commands very easy, at the expense of the user creating his own custom code to deal with the results.

bbinet:
I had the same needs, and I've implemented a "raw" command which may be useful to you.
I will submit a PR soon.

bbinet:
See PR #63

MisterScience5:
Excellent, thank you very much @bbinet :)

david-cermak:
https://github.com/espressif/esp-protocols/pull/63 has been merged

PR #63, opened 2022-06-22
Add raw command (IDFGH-7658)

bbinet:
See #59


david-cermak (approved):
Thanks for addressing the issue!

The only thing I'm not sure about is the name of the API: `raw` (?)
Would suggest something like `command`, but that already exists in `DCE` context; `custom_command` is too long, maybe `at_command` or simply `at`?

david-cermak:
@bbinet Thanks for the contribution!

PR #64, opened 2022-06-22
Allow to customize uart source_clk (IDFGH-7659)

bbinet:
This is for example needed when using an external crystal to select
UART_SCLK_XTAL, which fixes uart comm issue when using PM with light
sleep enabled.


david-cermak (approved):

david-cermak:
@bbinet Make sense!  And thanks again for your help.

Host test fails, as our linux port doesn't know these clock sources. Please add a stub typedef next to:

https://github.com/espressif/esp-protocols/blob/9905fe4cf62af0ae2ef5beb4bd9bf3596072bbdd/components/esp_modem/port/linux/esp_system_protocols_linux/include/driver/uart.h#L17-L20

```cpp
typedef int uart_sclk_t;
```

bbinet:
I've updated this PR as requested. See commit [d723fb7](https://github.com/espressif/esp-protocols/pull/64/commits/d723fb7a5a59850b53f876ad3279ca4837155396)

Thank you for the great development framework!

PR #32, opened 2022-05-02
Add websocket client error message string

0xFEEDC0DE64:
Makes debugging an faulty esp32 websocket client connection at customer's site easier.

PR #84, opened 2022-07-11
asio: fix exampels build (IDFGH-7792)

Lapshin:
Found that asio examples can't be built because can not find asio include files (https://github.com/espressif/esp-protocols/runs/7277980358?check_suite_focus=true)

This change fixes that

### DIFF
--- a/.github/workflows/build_asio.yml
+++ b/.github/workflows/build_asio.yml
@@ -16,6 +16,7 @@ jobs:
       - name: Checkout esp-protocols
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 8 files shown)


Lapshin:
@david-cermak @gabsuren , I picked commit from #32 to check how it works now

Lapshin:
@david-cermak , also mdns test_app was affected. Added commit with a fix

david-cermak (approved):

david-cermak:
@Lapshin Thanks for the fixes!

PR #89, opened 2022-07-13
Use USB CDC driver from component registry (IDFGH-7816)

tore-espressif:
Hi @david-cermak ,
This PR updates USB DTE example.

There are two outstanding items:
1. Is there a way to add the (non-default) USB-DTE example into build job in CI? (something like `sdkconfig.ci.usb` ?)
2. It would be better to publish `esp_modem_usb_dte` component as separate component registry. Then we could remove ugly checks [here](components/esp_modem/examples/modem_console/components/esp_modem_usb_dte/CMakeLists.txt), and use usb_dte in other esp_modem applications too.

What do you think?


david-cermak:
Hi @tore-espressif 

> 1. Is there a way to add the (non-default) USB-DTE example into build job in CI? (something like `sdkconfig.ci.usb` ?)

You can add  `sdkconfig.ci.usb`, but we don't have a mechanism to iterate over different configs (yet, might use `build_apps` in future). have to add it manually as we do e.g. here:

https://github.com/espressif/esp-protocols/blob/85c72826411181183a452f22245a5a1906717f0c/.github/workflows/target-test.yml#L63-L64


> 2. It would be better to publish `esp_modem_usb_dte` component as separate component registry. Then we could remove ugly checks [here](components/esp_modem/examples/modem_console/components/esp_modem_usb_dte/CMakeLists.txt), and use usb_dte in other esp_modem applications too.

I think it's a good idea to create a separate component that's pre-configured to use the USB DTE. I would just prefer to keep one source here and publish it into two components in CI.

tore-espressif:
@david-cermak @leeebo Modem_console USB example is extended with hot-plugging feature and added to CI, PTAL

tore-espressif:
Changes since last review:
1. esp_modem_usb_dte is now a separate component https://components.espressif.com/components/espressif/esp_modem_usb_dte
2. USB-DTE now offers C API too
3. Simple PPPoS example now supports USB-DTE
4. Sudden disconnection supported
5. USB example builds added to CI

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -157,9 +175,25 @@ void app_main(void)
...
+
+#elif defined(CONFIG_EXAMPLE_SERIAL_CONFIG_USB)
+    while (1) {
+        ESP_LOGI(TAG, "Initializing esp_modem for the BG96 module...");
+        struct esp_modem_usb_term_config usb_config = ESP_MODEM_DEFAULT_USB_CONFIG(0x2C7C, 0x0296, 2); // VID, PID and interface num of BG96 modem
+        const esp_modem_dte_config_t dte_usb_config = ESP_MODEM_DTE_DEFAULT_USB_CONFIG(usb_config);  
+        ESP_LOGI(TAG, "Waiting for USB device connection...");
+        esp_modem_dce_t *dce = esp_modem_new_dev_usb(ESP_MODEM_DCE_BG96, &dte_usb_config, &dce_config, esp_netif);
+        esp_modem_set_error_cb(dce, usb_terminal_error_handler);
+        vTaskDelay(1000);
```
david-cermak, line 188: Nitpick:
```suggestion
        vTaskDelay(pdMS_TO_TICKS(...));
```

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -157,9 +175,25 @@ void app_main(void)
 #else
     ESP_LOGI(TAG, "Initializing esp_modem for a generic module...");
     esp_modem_dce_t *dce = esp_modem_new(&dte_config, &dce_config, esp_netif);
+#endif
+
+#elif defined(CONFIG_EXAMPLE_SERIAL_CONFIG_USB)
+    while (1) {
```
david-cermak, line 181: Could we make the while loop "bigger", so it's better readable for both UART and USB?

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -134,17 +163,6 @@ void app_main(void)
     dte_config.task_priority = CONFIG_EXAMPLE_MODEM_UART_EVENT_TASK_PRIORITY;
     dte_config.dte_buffer_size = CONFIG_EXAMPLE_MODEM_UART_RX_BUFFER_SIZE / 2;
 
-    /* Configure the DCE */
-    esp_modem_dce_config_t dce_config = ESP_MODEM_DCE_DEFAULT_CONFIG(CONFIG_EXAMPLE_MODEM_PPP_APN);
-
-    /* Configure the PPP netif */
-    esp_netif_config_t netif_ppp_config = ESP_NETIF_DEFAULT_PPP();
```
david-cermak, line 141: Here I wanted to demonstrate the sequence of config and init, so configure the DTE, then DCE and then PPP. And after that we would start creating objects (in C we do that in one step). I'm not sure if we could keep this order, maybe if we make the UART part destroyable, too?

david-cermak (approved):
Nice work! LGTM.
left just some questions about readability of the most basic example. But that could be addressed later by me (Also wanted to show the cleanup and reinit of the DCE using C-API)

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
tore-espressif: Yes, the UART and netif objects should be destroyed and then recreated. But I didn't want to make this PR too big. And since you already have an idea about doing this, I'd prefer doing it in follow-up PR.

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
tore-espressif: I agree, this change does not improve readability :(

I plan to follow up this PR for support of multiple types of DCE. Currently the USB examples support only BG96. So the USB initialization sequence could then by similar to UART. But again, would you agree with follow-up PR?

tore-espressif (commented):
@david-cermak Thank you very much for the review.

In the recent push I only resolved conflicts and provided explanation to the artificial delay.

Overall, IMO, this PR is reasonably large and is open for almost 3 months(!). So if you agree, we could merge it and follow up with another PR.

PR #92, opened 2022-07-18
feat(mdns): Define explicit dependencies on esp-wifi

david-cermak:
`mdns` depends on `esp-wifi`, but currently it's is transiently brought by `esp-netif` (Which doesn't have to depend on esp-wifi)


gabsuren (approved):

david-cermak:
> To do this, we could use idf_component_optional_requires in CMake, and #if __has_include("esp_wifi.h") in C source code.

It's a bit more complicated than that, as we'd have to rewrite our current (default) event handlers that use (many) events from the WiFi library.

But nice suggestion, noted!

PR #95, opened 2022-07-22
get_gnss_information_sim70xx added ! (IDFGH-7903)

diplfranzhoepfinger:
this enables to read out the Location information from GPS, this works for SIM70xx, tested on SIM7070.

### DIFF
--- a/common_components/protocol_examples_common/CMakeLists.txt
+++ b/common_components/protocol_examples_common/CMakeLists.txt
@@ -1,4 +1,4 @@
 idf_component_register(SRCS "connect.c" "stdin_out.c" "addr_from_stdin.c"
                     INCLUDE_DIRS "include"
-                    PRIV_REQUIRES esp_netif driver esp_eth esp_wifi
+                    PRIV_REQUIRES esp_netif driver esp_eth esp_wifi vfs
                     )
--- a/common_components/protocol_examples_common/CMakeLists.txt
+++ b/common_components/protocol_examples_common/CMakeLists.txt
@@ -1,4 +1,4 @@
 idf_component_register(SRCS "connect.c" "stdin_out.c" "addr_from_stdin.c"
                     INCLUDE_DIRS "include"
-                    PRIV_REQUIRES esp_netif driver esp_eth
+                    PRIV_REQUIRES esp_netif driver esp_eth esp_wifi
                     )
--- a/components/esp_modem/examples/pppos_client/main/pppos_client_main.c
+++ b/components/esp_modem/examples/pppos_client/main/pppos_client_main.c
@@ -209,10 +209,11 @@ void app_main(void)
             .broker = {
                 .address.uri = BROKER_URL
             },
+			//TODO !! Event Handler !
 #else
             .uri = BROKER_URL,
-#endif
             .event_handle = mqtt_event_handler,
+#endif
     };
     esp_mqtt_client_handle_t mqtt_client = esp_mqtt_client_init(&mqtt_config);
     esp_mqtt_client_start(mqtt_client);
--- a/components/esp_modem/examples/pppos_client/main/pppos_client_main.c
+++ b/components/esp_modem/examples/pppos_client/main/pppos_client_main.c
@@ -16,6 +16,7 @@
 #include "mqtt_client.h"
 #include "esp_modem_api.h"
 #include "esp_log.h"
+#include "esp_idf_version.h"
 
 
 #define BROKER_URL "mqtt://mqtt.eclipseprojects.io"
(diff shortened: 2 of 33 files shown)


diplfranzhoepfinger:
so, 

get\_gnss\_information\_sim70xx makes it possible to request GPS Position and other Info like speed and course from SIMCOM SIM7070 and possibly other Modules.

diplfranzhoepfinger:
@david-cermak

common_components/protocol_examples_common/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 idf_component_register(SRCS "connect.c" "stdin_out.c" "addr_from_stdin.c"
                     INCLUDE_DIRS "include"
-                    PRIV_REQUIRES esp_netif driver esp_eth
+                    PRIV_REQUIRES esp_netif driver esp_eth esp_wifi vfs
```
david-cermak, line 3: :+1:

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -201,8 +205,15 @@ void app_main(void)
     xEventGroupWaitBits(event_group, CONNECT_BIT, pdTRUE, pdTRUE, portMAX_DELAY);
     /* Config MQTT */
     esp_mqtt_client_config_t mqtt_config = {
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
+            .broker = {
+                .address.uri = BROKER_URL
+            },
+			//TODO !! Event Handler !
```
david-cermak, line 212: Should use event handler registration instead:

https://github.com/espressif/esp-idf/blob/36f49f361c001b49c538364056bc5d2d04c6f321/examples/protocols/mqtt/tcp/main/app_main.c#L142

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -201,8 +205,15 @@ void app_main(void)
     xEventGroupWaitBits(event_group, CONNECT_BIT, pdTRUE, pdTRUE, portMAX_DELAY);
     /* Config MQTT */
     esp_mqtt_client_config_t mqtt_config = {
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
```
david-cermak, line 208: :+1:

components/esp_modem/include/nmea_parser.h
```diff
@@ -0,0 +1,137 @@
+/*
+ * SPDX-FileCopyrightText: 2015-2022 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Origin: https://github.com/espressif/esp-idf/blob/master/examples/peripherals/uart/nmea0183_parser/main/nmea_parser.h
```
david-cermak, line 6: Let's just reference the exact commit instead of master, e.g. 
```suggestion
 * Origin: https://github.com/espressif/esp-idf/blob/36f49f36/examples/peripherals/uart/nmea0183_parser/main/nmea_parser.h
```

(agree that it's probably better to simply copy this header here, than reference from IDF/examples)

components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -560,4 +562,280 @@ command_result set_gnss_power_mode_sim76xx(CommandableIf *t, int mode)
...
+        return command_result::FAIL;
+    }
+    gps.run = (gps_run_t)GNSS_run_status;
+    out = out.substr(pos + 1);
+    if ((pos = out.find(',')) == std::string::npos) {
+    	return command_result::FAIL;
+    }
+    //Fix status
+    {
+		std::string_view fix_status = out.substr(0 , pos);
```
david-cermak, line 618: This change/file needs a cleanup in general, at least for formatting and mixed spaces+tabs

david-cermak (commented):
Thanks for posting this useful update, Nice work!

And thanks for addressing the recent breaking changes in IDFv5.0
* mqtt client's config struct
* explicit dependencies on network interface/driver components

As for the gnss information, my general comment would be to keep it a little separate from the rest of modem APIs. This doesn't belong to the most generic modem functionalities and is not supported on certain devices.
I would suggest removing the API from `generate/esp_modem_commands.inc` (which in turn helps decoupling most of the modules from the `nmea_parser.h`) and introduce an additional header that could be included only if users need this API. The same suggestion goes for the implementation. If possible, please try to add as a separate module (might be that some internal calls like `generic_get_string()`  might need to be promoted to the actutal/private APIs).

PR #96, opened 2022-07-22
fix(mdns): ignore authoritative flag on reception (IDFGH-7891)

gjc13:
According to [RFC 6762](https://datatracker.ietf.org/doc/html/rfc6762#section-18.4):

   In response messages for Multicast domains, the Authoritative Answer
   bit MUST be set to one (not setting this bit would imply there's some
   other place where "better" information may be found) and MUST be
   ignored on reception.

The PR removes the check on the AA bit.


Alvin1Zhang:
Thanks for your contribution.

david-cermak (approved):
Thanks for the fix!

PR #98, opened 2022-07-26
mdns: fix bit order issue in DNS header flags

gjc13:
The bit field in the header flags does not match its intended order on
little endian machines. The PR removes the bit field and uses bit
operations instead.

### DIFF
--- a/components/mdns/private_include/mdns_private.h
+++ b/components/mdns/private_include/mdns_private.h
@@ -71,7 +71,9 @@
 #define MDNS_ANSWER_A_TTL           120
 #define MDNS_ANSWER_AAAA_TTL        120
 
-#define MDNS_FLAGS_AUTHORITATIVE    0x8400
+#define MDNS_FLAGS_QUERY_REPSONSE   0x8000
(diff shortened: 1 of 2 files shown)


components/mdns/private_include/mdns_private.h
```diff
@@ -212,18 +214,6 @@ typedef enum {
 typedef struct {
     uint16_t id;
     union {
```
david-cermak, line 216: Nitpick: Could we also remove the union here? (so instead of `header.flags.value` we'd just say `header.flags` ?)

david-cermak (approved):
LGTM, Thanks for addressing this problem!

components/mdns/private_include/mdns_private.h
david-cermak: (if you're addressing this, could you please rebase as well? we've just merged some fixes to CI failures against IDF master)

david-cermak:
Closing in favour of https://github.com/espressif/esp-protocols/pull/114

PR #106, opened 2022-08-08
get_gnss_information_sim70xx added ! (IDFGH-8039)

diplfranzhoepfinger:
this enables to read out the Location information from GPS, this works for SIM70xx, tested on SIM7070.
not finished. will get updates.


franz-ms-muc:
@david-cermak i need your Help a bit.

Issue #103, opened 2022-08-01
esp_modem - cant exit CMUX on BG96 (IDFGH-7954)


chegewara:
Hi,
esp_modem example works fine to enter CMUX on BG96, but i am having problem to exit CMUX and switch to command mode.
For some reason my BG96 is not responding on close DLCI0 command.
```
V (543) command_lib: set_cmux
V (543) command_lib: generic_command_common
V (543) command_lib: generic_command
D (543) command_lib: generic_command command AT+CMUX=0

D (553) command_lib: Response: 
OK

V (653) CMUX Received: 0x3ffc0d10   f9 03 73 01 d7 f9                                 |..s...|
D (653) CMUX: Payload frame: dlci:00 type:73 payload:0 available:2
V (663) CMUX Received: 0x3ffc0d10   f9 07 73 01 15 f9                                 |..s...|
D (663) CMUX: Payload frame: dlci:01 type:73 payload:0 available:2
V (673) CMUX Received: 0x3ffc0d10   f9 0b 73 01 92 f9                                 |..s...|
D (673) CMUX: Payload frame: dlci:02 type:73 payload:0 available:2
V (683) command_lib: set_echo
V (683) command_lib: generic_command_common
V (693) command_lib: generic_command
D (693) command_lib: generic_command command ATE0

V (703) CMUX Received: 0x3ffc0d10   f9 05 ef 0d 0d 0a 4f 4b  0d 0a 5f f9              |......OK.._.|
D (713) CMUX: Payload frame: dlci:01 type:ef payload:6 available:8
D (713) command_lib: Response: 
OK

V (723) command_lib: set_pdp_context
V (723) command_lib: generic_command_common
V (723) command_lib: generic_command
D (733) command_lib: generic_command command AT+CGDCONT=1,"IP","plus"

V (763) CMUX Received: 0x3ffc0d10   f9 05 ef 0d 0d 0a 4f 4b  0d 0a 5f f9              |......OK.._.|
D (763) CMUX: Payload frame: dlci:01 type:ef payload:6 available:8
D (773) command_lib: Response: 
OK

V (773) command_lib: set_data_mode
V (773) command_lib: generic_command
D (783) command_lib: generic_command command ATD*99##

V (793) CMUX Received: 0x3ffc0d10   f9 05 ef 2b 0d 0a 43 4f  4e 4e 45 43 54 20 31 35  |...+..CONNECT 15|
[...]

PR #115, opened 2022-08-19
fix(esp_modem): Correct exit of CMUX mode

david-cermak:
* Fix DISC message's control field (0xFF->0xFE) to keep Poll bit cleared
* Accept UIH resp on DISC for both (0xFF, 0xFE) ignoring P/F bit
* Fix timeout processing when handling DISC message

Closes https://github.com/espressif/esp-protocols/issues/103


euripedesrocha (approved):

david-cermak:
Thanks for the review!

PR #116, opened 2022-08-19
mdns: Example tests integration

gabsuren:


david-cermak (approved):
looks very good to me!

Issue #110, opened 2022-08-11
abort() was called at PC 0x420c4603 on core 1


franz-ms-muc:
it seems there is an exception somwhere, 

i assume from "throw" ... 

```

ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x2f (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd03e0,len:0x32b0
load:0x403b6000,len:0xbb8
load:0x403b9000,len:0x4c0c
entry 0x403b6260
W (30) boot.esp32s3: eFuse virtual mode is enabled. If Secure boot or Flash encryption is enabled then it does not provide any security. FOR TESTING ONLY!
I (46) boot: ESP-IDF 01d014c-dirty 2nd stage bootloader
I (46) boot: compile time 18:11:37
I (47) boot: chip revision: 0
I (49) boot.esp32s3: Boot SPI Speed : 80MHz
I (54) boot.esp32s3: SPI Mode       : DIO
I (58) boot.esp32s3: SPI Flash Size : 8MB
I (63) boot: Enabling RNG early entropy source...
I (68) boot: Partition Table:
I (72) boot: ## Label            Usage          Type ST Offset   Length
I (79) boot:  0 otadata          OTA data         01 00 0000d000 00002000
I (87) boot:  1 reserved         WiFi data        01 02 0000f000 00001000
I (94) boot:  2 ota_0            OTA app          00 10 00010000 00280000
I (102) boot:  3 ota_1            OTA app          00 11 00290000 00280000
I (109) boot:  4 prod             Unknown data     01 ff 00610000 00003000
I (117) boot:  5 safeNVrb0        Unknown data     01 ff 00613000 00003000
I (124) boot:  6 safeNVrb1        Unknown data     01 ff 00616000 00003000
I (132) boot:  7 safeNVrb2        Unknown data     01 ff 00619000 00003000
I (139) boot:  8 config           WiFi data        01 02 0061c000 00046000
I (147) boot:  9 appl             WiFi data        01 02 00662000 00004000
I (154) boot: 10 shared           WiFi data        01 02 00666000 00003000
W (162) efuse: Loading virtual efuse blocks from flash
EFUSE_BLKx:
0) 0x00800101 0x000f0000 0x09c00000 0x00500000 0x00000022 0x00000000
[...]

PR #125, opened 2022-08-24
fix(esp_modem): DTE command race of timeout vs reply's signal

david-cermak:
Race condtion:
* First command timeouted, but the reply came just after evaluation and
set signal variable to GOT_LINE
* Second command should timeout too, but a consistency check validates
that it timeouted and at the same time GOT_LINE (from previous step) and
throws an exception

STR:
* Revert change in esp_modem_dte.cpp
* Run TEST_CASE("DTE command races", "[esp_modem]")

Closes https://github.com/espressif/esp-protocols/issues/110


euripedesrocha (approved):

david-cermak:
Thanks for the review @euripedesrocha !

PR #127, opened 2022-08-30
FlowControlSwitchable (IDFGH-8452)

diplfranzhoepfinger:
FlowControlSwitchable

this enables to make a Autobauding, (in 2-wire mode)
and then make a switch to 4-wire mode. 

for it it work you need to init the DCE in 4-wire mode, 
switch then to 2-wire, and later back to 4-wire.


david-cermak:
@diplfranzhoepfinger Sorry for such a late reply.

What is the motivation for switching the control flow runtime? I mean In general it's always good to extend the flexibility and add tests to exercise it, but is there any particular practical reason for that? Since we usually design boards with either two or four wire connection. Also adding a virtual method in the generic terminal class takes resources (few bytes, but still...) and expects other, USB and future terminals to define that. I think we should like to keep the abstract classes simple, so maybe adding the method to the uart-terminal only and test the switching before we construct DCE?

david-cermak:
As described in https://github.com/espressif/esp-protocols/pull/127#issuecomment-1268018632 I don't think that the runtime configurable UART flow control (on DCE level) brings much advantage.  Closing for now, please feel free to comment or reopen or rework to make the terminal only switchable.

franz-ms-muc:
> @diplfranzhoepfinger Sorry for such a late reply.
> 
> What is the motivation for switching the control flow runtime? I mean In general it's always good to extend the flexibility and add tests to exercise it, but is there any particular practical reason for that? Since we usually design boards with either two or four wire connection. Also adding a virtual method in the generic terminal class takes resources (few bytes, but still...) and expects other, USB and future terminals to define that. I think we should like to keep the abstract classes simple, so maybe adding the method to the uart-terminal only and test the switching before we construct DCE?

the Reason is: 

Autobaud. 

if the SIM-Module, e.g. the SIM7070 is in Autobaud Mode, you must send AT AT AT as long as it answers with OK. 

then, AFTER this you can activate the 4-wire Mode on the ESP32 side and on the Module Side. (Module side you send the set-Flow-control Message, therefore it is required to be out of Autobaud before this)

franz-ms-muc:
and yes: the Reason for Support Autobaud is the SIM7070 has this Setting as Default. not a Fixed Baudrate but Autobaud.

david-cermak:
Oh, I see. Thanks for explaining, makes perfect sense. Let me reopen and maybe suggest some alternatives...

diplfranzhoepfinger:
@david-cermak the [https://github.com/espressif/esp-protocols/CONTRIBUTING.md](https://github.com/espressif/esp-protocols/CONTRIBUTING.md) seems not to exist.

diplfranzhoepfinger:
@david-cermak have updated this PR, as we need it for Autobauding.

david-cermak:
As mentioned previously, I'm quite confident that we should propagate this UART specific features up to DTE and DCE layers, that could be constructed using other terminals.

I was thinking about alternatives, and one option would be to support detaching a terminal from DTE. The change is sketched here:
https://github.com/espressif/esp-protocols/compare/master...david-cermak:esp-protocols:feat/modem_dte_detach_term

This is not perfect either, as we have to *downcast* the generic `Terminal` back to `UartTerminal` and that's not nice without RTTI.

Another alternative is to use UART API directly in the application code. This is safe as the resource is being locked internally and probably okay.

```patch
diff --git a/components/esp_modem/examples/modem_console/main/modem_console_main.cpp b/components/esp_modem/examples/modem_console/main/modem_console_main.cpp
index 2f8052e..85d8083 100644
--- a/components/esp_modem/examples/modem_console/main/modem_console_main.cpp
+++ b/components/esp_modem/examples/modem_console/main/modem_console_main.cpp
@@ -192,6 +192,16 @@ extern "C" void app_main(void)
 
     assert(dce != nullptr);
 
+    uart_set_sw_flow_ctrl(dte_config.uart_config.port_num, true, 8, UART_FIFO_LEN - 8);
+
+    dce->sync();
+    //....
+
+    uart_set_hw_flow_ctrl(dte_config.uart_config.port_num, UART_HW_FLOWCTRL_CTS_RTS, UART_FIFO_LEN - 8);
+
+    dce->sync();
+    //....
+
     if (dte_config.uart_config.flow_control == ESP_MODEM_FLOW_CONTROL_HW) {
         if (command_result::OK != dce->set_flow_control(2, 2)) {
             ESP_LOGE(TAG, "Failed to set the set_flow_control mode");
```

Also, as mentioned previously, Isn't is possible to store the current configuration. so you don't have to go over the autobauding over again after each startup?

diplfranzhoepfinger:
> 
> Also, as mentioned previously, Isn't is possible to store the current configuration. so you don't have to go over the autobauding over again after each startup?

Hi, 

the Baudrate *can* be stored, then the Autobauding is not needed anymore. 

 but the Customer wanted a Firmware which is running on a Factory-blank SIMCOM Module. 
and the Default Setting is Autobaud. 

i will talk to the Customer again.

diplfranzhoepfinger:
Hi, 

here https://github.com/diplfranzhoepfinger/esp-protocols/commit/cb569065cdaff6186df9f8be250e959ce191978c i tested ESP-Modem Console without any Autobauding, 
on a SIMCOM-A7672E-FASE

even it is Fixed Baudrate 115200 it does not work without the "sync" Period at the Beginning.

diplfranzhoepfinger:
> Another alternative is to use UART API directly in the application code. This is safe as the resource is being locked internally and probably okay.

yes, i try this now. looks great !

diplfranzhoepfinger:
so your suggestion works:

diplfranzhoepfinger:
https://github.com/diplfranzhoepfinger/esp-protocols/commit/0a33ce531c5543391e32cf2296423709ae660b32

PR #131, opened 2022-09-02
esp_modem: Expose set_error_cb method

tore-espressif:
Allow user to register his DTE error handler. Errors from underlying terminals are forwarded to this DTE error handler.

@david-cermak Currently both `data_term` and `command_term` use the same error handler, is it OK?

### DIFF
--- a/components/esp_modem/CMakeLists.txt
+++ b/components/esp_modem/CMakeLists.txt
@@ -47,5 +47,3 @@ if(${target} STREQUAL "linux")
     # This is needed for ESP_LOGx() macros, as integer formats differ on ESP32(..) and x64
     set_target_properties(${COMPONENT_LIB} PROPERTIES COMPILE_FLAGS -Wno-format)
 endif()
-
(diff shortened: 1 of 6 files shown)


tore-espressif:
cc @leeebo This change prepares esp_modem to handle USB modem disconnection

leeebo:
@tore-espressif Sorry for the late response, is there any way to pass USB Connect/Reconnect event to esp_modem or user code, due to that the dial-up process may need to be re-run for the BUS-powered device.

tore-espressif:
> @tore-espressif Sorry for the late response, is there any way to pass USB Connect/Reconnect event to esp_modem or user code, due to that the dial-up process may need to be re-run for the BUS-powered device.

Yes, there is a callback on device disconnection, where user would typically set his network interface to 'down' state.

For device connection/reconnection there is only blocking wait function (with timeout as parameter). An example here https://github.com/espressif/esp-protocols/pull/89

PR #135, opened 2022-09-07
mdns: fix IPV4 only build and also update CI configuration

mahavirj:
mdns: fix IPV4 only build and also update CI configuration
    
Earlier commit 48c157bc469d15c24d85eaaf0fd2bce004b87d45 had typo regarding variable name. This was not caught in CI because IPV6 was not really getting disabled due to an indirect dependency of `CONFIG_EXAMPLE_CONNECT_IPV6`.
    
This commit fixes both the problems.


mahavirj:
@kostaond Please take a look at SPI ethernet related changes. These changes were not noticed previously as CI was largely broken here (now fixed in this PR).

kostaond:
LGTM

PR #137, opened 2022-09-08
mdns: WIP make networking decoupled from _mdns_server

david-cermak:


david-cermak:
@antmak PTAL

antmak:
@david-cermak LGTM, thanks

PR #138, opened 2022-09-09
esp_modem: Fix IDF version resolution (IDFGH-8271)

tore-espressif:
Fix IDF version resolution in esp_modem


tore-espressif:
@david-cermak PTAL

david-cermak (approved):
Thanks for the fixes!

Issue #129, opened 2022-08-30
Command Timeouts (IDFGH-8185)


AshUK:
This is somewhat related to #66 

From testing and from the BG96/95 modem data sheet require longer timeouts.

- set_radio_state/AT+CFUN= upto 15s
- set_pdp_context/AT+CGACT= upto 150s

I have from testing been seeing both of these commands timing out when using the default 300ms.

It would also be useful if the `esp_modem_at` function had a custom timeout parameter.

See [BG95&BG77_AT_Commands_Manual.pdf ](https://www.quectel.com/download/quectel_sc200k-ce_%E7%A1%AC%E4%BB%B6%E8%AE%BE%E8%AE%A1%E6%89%8B%E5%86%8C_v1-0/quectel_bg95bg77bg600l_series_at_commands_manual_v2-0) pages 27 and 134.

david-cermak:
Timeouts updated in https://github.com/espressif/esp-protocols/pull/139, also added the timeout parameter to the `at()` command.
Thanks for reporting!

Issue #128, opened 2022-08-30
esp_modem_get_operator_name ACT value (IDFGH-8184)


AshUK:
esp-idf: 4.4.2
esp-modem: 0.1.19
module: BG95

When using the C API `esp_modem_get_operator_name` never updates the act value.

david-cermak:
Thanks for the report! It's a bug, fix available in https://github.com/espressif/esp-protocols/pull/139

PR #139, opened 2022-09-09
fix(esp_modem): Make get_operator_name() return also ACT value

david-cermak:
1) Updated the C++ API in a2866343, but forgot to fix the C API accordingly
2) Corrected timeouts for some commands (`set_pdp_context()`, `set_radio_state()`)
3) Added an explicit timeout parameter to the `esp_modem_at()`

* Closes https://github.com/espressif/esp-protocols/issues/128
* Closes https://github.com/espressif/esp-protocols/issues/129

### DIFF
--- a/components/esp_modem/src/esp_modem_command_library.cpp
+++ b/components/esp_modem/src/esp_modem_command_library.cpp
@@ -273,7 +273,7 @@ command_result set_pdp_context(CommandableIf *t, PdpContext &pdp)
     ESP_LOGV(TAG, "%s", __func__ );
     std::string pdp_command = "AT+CGDCONT=" + std::to_string(pdp.context_id) +
                               ",\"" + pdp.protocol_type + "\",\"" + pdp.apn + "\"\r";
-    return generic_command_common(t, pdp_command);
+    return generic_command_common(t, pdp_command, 150000);
 }
 
 command_result set_data_mode(CommandableIf *t)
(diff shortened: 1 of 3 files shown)


tore-espressif:
@david-cermak please don't forget to bump the idf_component.yml version!

@kumekay do we have any CI or pre-commit check for this usecase? It happens quite often that we forget to bump versions on relevant changes

kumekay:
> @kumekay do we have any CI or pre-commit check for this usecase? It happens quite often that we forget to bump versions on relevant changes

@tore-espressif As I understand components in this repo may be updated not after every PR. However, it's a nice idea to make a linter for component manifests and wrap it in a CI action or pre-commit hook

components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -273,7 +273,7 @@ command_result set_pdp_context(CommandableIf *t, PdpContext &pdp)
     ESP_LOGV(TAG, "%s", __func__ );
     std::string pdp_command = "AT+CGDCONT=" + std::to_string(pdp.context_id) +
                               ",\"" + pdp.protocol_type + "\",\"" + pdp.apn + "\"\r";
-    return generic_command_common(t, pdp_command);
+    return generic_command_common(t, pdp_command, 150000);
```
AshUK, line 276: if this is ms should it not be 15000 ?

euripedesrocha (approved):

Issue #37, opened 2022-05-15
Disable CMUX when it is no longer required (IDFGH-7372)


jonathandreyer:
Hi everyone,

Does it possible to disable the CMUX after the usage? My idea is to activate the CMUX when the use of PPP is necessary and disable it after the PPP mode is no longer required. I ask that because I have tried with the following code (at the end of the *simple_cmux_example*), but it is not working.
```
if (dce->set_mode(esp_modem::modem_mode::COMMAND_MODE)) {
    std::cout << "Modem has correctly entered command mode" << std::endl;
} else {
    ESP_LOGE(TAG, "Failed to configure desired mode... exiting");
    return;
}
```

Best regards,

Jonathan Dreyer

david-cermak:
Yes, it should be possible to exit CMUX mode (haven't really tried with real devices, though), but this is not implemented in the `esp-modem` component. The CMUX mode only creates two virtual terminals so that one could be used for commands and the other one for PPP mode.
When PPP mode is not longer required, you can still exit the PPP mode and have two virtual terminals in command mode.

jonathandreyer:
Hi, thanks for your message. I will study both approaches, but I have a preference for closing the CMUX mode. I think that is more elegant than leave unused resources open. Do you have any recommendations about that?
Best regards,
Jonathan

jonathandreyer:
Hi, thanks for your reply. I will study the behaviour of the CMUX and see if I can do something for this point (the CMUX protocol seems to me a bit complex). If I have something, I will leave a comment here.
Best regards,
Jonathan

david-cermak:
Hi Jonathan,

There's still a problem of the *meaning* of the modes. Currently, when we switch to `CMUX` we actually create two virtual terminals, both in command mode. then any further switching to `PPP`, changes the mode of one virtual terminal only.

So yes, you're correct.
1) set mode to `CMUX`
2) set mode to `PPP`
~ Do work
3) Set mode to `CMD` should exit both virtual terminals and leave only one (physical) in command mode.

I think it would probably make more sense to have the `CMUX` mode automatically create one command terminal and one data terminal (i.e. points `1)` and `2)` in a single step: set mode to `CMUX`).
---

> I have tested but that not work in my case. 

Could you please share the logs?

> implement it in the CMUX example (optionally activated with a define variable)

That would be helpful, thanks!

david-cermak:
[...]

PR #144, opened 2022-09-19
esp_modem: Exit CMUX after end of example

jonathandreyer:
Exit CMUX mode at the end of the *simple_cmux_client* example (as describe in this [comment](https://github.com/espressif/esp-protocols/issues/37#issuecomment-1155727790)).
Here is also two minor fixes (formatting & spelling).


david-cermak (approved):
Thanks for extending the CMUX example! LGTM!

PR #146, opened 2022-09-21
Prevent crash when hostname is null (IDFGH-8356)

MicroDev1:
Check whether `_mdns_server->hostname` is `_str_null_or_empty` before `strcasecmp`.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -2801,7 +2801,8 @@ static int _mdns_check_aaaa_collision(esp_ip6_addr_t * ip, mdns_if_t tcpip_if)
 
 static bool _hostname_is_ours(const char * hostname)
 {
-    if (strcasecmp(hostname, _mdns_server->hostname) == 0) {
+    if (!_str_null_or_empty(_mdns_server->hostname) &&
+        strcasecmp(hostname, _mdns_server->hostname) == 0) {
         return true;
     }
     mdns_host_item_t * host = _mdns_host_list;


gabsuren (approved):

gabsuren:
@MicroDev1 thanks for the contribution, LGTM!

gabsuren:
@david-cermak could you please merge it, if the changes look good to you

david-cermak:
> @david-cermak could you please merge it, if the changes look good to you

LGTM, as well.

@MicroDev1 Missing your agreement, could you please click trough the CLA assistant above

Could you also please elaborate on the conditions of the crash? Are you trying to advertise services without a hostname set?

MicroDev1:
> @MicroDev1 Missing your agreement, could you please click trough the CLA assistant above

It's signed now. :)

> Could you also please elaborate on the conditions of the crash? Are you trying to advertise services without a hostname set?

A crash could happen on immediately checking if hostname exists after setting it like in the following code from [CircuitPython](https://github.com/adafruit/circuitpython/blob/fcf7cfe8388da28f44c071cc1ba31850c2091dc5/ports/espressif/common-hal/mdns/Server.c#L91-L98).
```c
void common_hal_mdns_server_set_hostname(mdns_server_obj_t *self, const char *hostname) {
    mdns_hostname_set(hostname);
    // Wait for the mdns task to set the new hostname.
    while (!mdns_hostname_exists(hostname)) {
        RUN_BACKGROUND_TASKS;
    }
    self->hostname = hostname;
}
```

david-cermak:
This is weird, as the `mdns_hostname_set()` blocks and waits for the mdns task to accept the settings:

https://github.com/espressif/esp-protocols/blob/9d9d0db1ee1fc8c0fa02995aca9cb48b1201e471/components/mdns/mdns.c#L5325-L5330

so, should be enough to do:
```cpp
void common_hal_mdns_server_set_hostname(mdns_server_obj_t *self, const char *hostname) {
    mdns_hostname_set(hostname);
    self->hostname = hostname;
}
```

But the `mdns_hostname_exists()` is still a public API and shouldn't crash like that! Thank you for posting the fix and explaining.

PR #147, opened 2022-09-23
esp_modem: Allow C API extensions (IDFGH-8368)

tore-espressif:
**This PR focuses on C API of esp_modem**

## 1. Extensible C API
**The problem:** C++ API is easily expandable with different types of DTE (such as USB DTE). But C API offers only UART DTE in function `esp_modem_new_dev` with no way of extending this function in C.

**The solution:** Move definitions needed for extending C API to separate header file `include/esp_private/c_api_wrapper.h`. Developers of esp_modem extension can include this file with `#include "esp_private/c_api_wrapper.h"` and implement their own `esp_modem_new_dev_*` function.

## 2. Error callback in C
**The problem:** C API doesn't offer API to register error callback to Terminal errors.

Composition of esp_modem C wrapper `struct esp_modem_dce_wrap`:
struct esp_modem_dce_wrap -> DCE -> DTE -> Terminal -> error_callback_function

DTE is protected (inaccessible) member of DCE.

We cannot add the error_callback to DTE configuration structure, because its signature contains C++ enum class (`terminal_error`), thus it is not compatible with C.

**The solution:**
**TODO:**

- [ ] ~~Provide C functions to create DTE and DCE separately~~ This proved as suboptimal. We maintain lifetime of DCE and DTE through smart pointer. So we can't easily pass DTE to DCE.
- [x] Provide C function that will allow registering DTE error callback

Now we provide 
1. C enum type `esp_modem_terminal_error_t` that is a C version of `terminal_error` 
2. Conversion function between the two types `convert_terminal_error_enum` 
3. `esp_modem_set_error_cb(esp_modem_dce_t * dce, esp_modem_terminal_error_cbt err_cb)` for registering error callback from C

~~**This PR doesn't update the default UART implementation**.~~ UART can be extended with error handling too.
~~Disadvantage: User cannot change the error callback after initialization (we don't offer C alternative to C++ `DTE::set_error_cb`)~~

### DIFF
--- a/components/esp_modem/include/esp_private/c_api_wrapper.hpp
+++ b/components/esp_modem/include/esp_private/c_api_wrapper.hpp
@@ -12,6 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/**
+ * @file c_api_wrapper.hpp
+ * @brief Collection of C API wrappers
+ * 
+ * This file is located in include/esp_private because it is not intended for users,
+ * but rather for esp_modem C extension developers.
+ * 
+ * The C extension API must provide a 'factory function' that returns initialized pointer to esp_modem_dce_wrap.
+ * Helper functions provided below, can be used for conversion between C++ enum classes and standard C enums.
+ */
+
 #pragma once
 
 #include "cxx_include/esp_modem_dce_factory.hpp"
@@ -23,6 +34,8 @@ struct esp_modem_dce_wrap { // need to mimic the polymorphic dispatch as CPP use
     enum class modem_wrap_dte_type { UART, VFS, USB } dte_type;
     dce_factory::ModemType modem_type;
     DCE *dce;
+    std::shared_ptr<DTE> dte;
+    esp_modem_dce_wrap() : dce(nullptr), dte(nullptr) {}
 };
 
 inline dce_factory::ModemType convert_modem_enum(esp_modem_dce_device_t module)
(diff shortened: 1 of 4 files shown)


tore-espressif:
@david-cermak PTAL, Mostly the include path of `c_api_wrapper.h` and the proposed changed of C error_callbacks

tore-espressif:
@david-cermak I updated the PR according to our discussion:

1. The `c_api_wrapper.h` was moved to `esp_private` subfolder, so it can be included by `#include "esp_private/c_api_wrapper.h"`
2. New function `esp_err_t esp_modem_set_error_cb(esp_modem_dce_t * dce, esp_modem_terminal_error_cbt err_cb);` introduced. It allows registering DTE error callback anytime on runtime.

I didn't provide C alternative function that would create DTE and DCE separately; the main reason for this is that we maintain lifetime of DCE and DTE objects through smart-pointers. Once we return DTE to C application as plain pointer, it goes of of scope.

components/esp_modem/src/esp_modem_c_api.cpp
```diff
@@ -84,6 +47,7 @@ extern "C" esp_modem_dce_t *esp_modem_new_dev(esp_modem_dce_device_t module, con
         delete dce_wrap;
         return nullptr;
     }
+    dce_wrap->dte = dte.get(); // Save plain pointer to DTE
```
david-cermak, line 50: This is I think potentially dangerous, as we keep the plain pointer and two lines below we transfer its ownership to `dce`. Since we create a shared pointer from `create_uart_dte()` I think it better to make a copy.

components/esp_modem/include/esp_private/c_api_wrapper.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include "cxx_include/esp_modem_dce_factory.hpp"
+#include "esp_modem_c_api_types.h"
+
+using namespace esp_modem;
+
+struct esp_modem_dce_wrap { // need to mimic the polymorphic dispatch as CPP uses templated dispatch
+    enum class modem_wrap_dte_type { UART, VFS, USB } dte_type;
+    dce_factory::ModemType modem_type;
+    DCE *dce;
+    DTE *dte; // Plain pointer to DTE. Lifetime of the underlaying DTE object is managed by DCE.
```
david-cermak, line 37:
```suggestion
    std::make_shared<DTE> dte;
```

david-cermak: Just note, that I was trying to keep the headers compatible with C++11, but sources could use C++17 constructs. This is okay if this header is private only.

david-cermak:
> I didn't provide C alternative function that would create DTE and DCE separately; the main reason for this is that we maintain lifetime of DCE and DTE objects through smart-pointers. Once we return DTE to C application as plain pointer, it goes of of scope.

I see, we would have to provide another wrapper around DTE.

---

The changes LGTM, just a minor suggestion (again) about the lifetime.
Alternatively we can add a DTE getter to the `DCE` class. Do you think that a public member `get_dte()` would help here?

tore-espressif:
@david-cermak  thank you very much for the review.

I'm sorry about the lifetime issues, this the first time I'm working with smart pointers... I changed the plain pointer to shared_ptr; will the DTE be properly destroyed in `esp_modem_destroy(esp_modem_dce_t *dce_wrap)` now?

> Alternatively we can add a DTE getter to the DCE class. Do you think that a public member get_dte() would help here?

I was thinking about it too, but I'm not a big fan of getters that expose private members, so if we could make it work without `get_dte()`, it would be better IMHO.

david-cermak:
> will the DTE be properly destroyed in `esp_modem_destroy(esp_modem_dce_t *dce_wrap)` now?

Yes, it should be deleted when calling a destruction on the wrapper. (No worries, I'm also having trouble with these lifetime issues, just created [this](https://godbolt.org/z/YMMq3e5ns) to double check)

> I'm not a big fan of getters that expose private members, so if we could make it work without `get_dte()`, it would be better IMHO.

Yes, I agree.

david-cermak (approved):
LGTM!

PR #150, opened 2022-10-04
Take over MQTT Broker like in MQTT Example.

diplfranzhoepfinger:
Take over MQTT Broker like in MQTT Example.


components/esp_modem/examples/simple_cmux_client/main/Kconfig.projbuild
```diff
@@ -66,5 +66,11 @@ menu "Example Configuration"
         default n
         help
             Close the multiplexed mode at the end of the example and rollback to command mode.
+            
+    config BROKER_URI
+        string "Broker URL"
+        default "mqtts://mqtt.eclipseprojects.io:8883"
```
david-cermak, line 72: Just for the simplicity
```suggestion
        default "mqtt://mqtt.eclipseprojects.io"
```

(so we don't have to add a certificate and don't want to demonstrate an SSL connection with no verification)

david-cermak:
Accepted via https://github.com/espressif/esp-protocols/pull/165 (merged as 71a2388acb51e7820fb4d03c942610e6c23ba7ff)

PR #155, opened 2022-10-06
esp_modem: Added modem_console commands for deep-sleep and modem PSM.

espressif-abhikroy:
Added the following commands to esp_modem: model_console:
* set_deep_sleep
* wakeup_modem
* enable_psm
* disable_psm
* get_psm_cfg
* set_psm_cfg


components/esp_modem/examples/modem_console/main/modem_console_main.cpp
```diff
@@ -59,8 +64,37 @@ using namespace esp_modem;
 static SignalGroup exit_signal;
 
 
+extern "C" void config_gpio(void)
```
david-cermak, line 67: No need for the C linkage
```suggestion
void config_gpio(void)
```

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
```diff
@@ -59,8 +64,37 @@ using namespace esp_modem;
...
+    io_conf.intr_type = GPIO_INTR_DISABLE;          //disable interrupt
+    io_conf.mode = GPIO_MODE_OUTPUT;                //set as output mode
+    io_conf.pin_bit_mask = GPIO_OUTPUT_PIN_SEL;     //bit mask of the pins that you want to set,e.g.GPIO18/19
+    io_conf.pull_down_en = GPIO_PULLDOWN_DISABLE;   //disable pull-down mode
+    io_conf.pull_up_en = GPIO_PULLUP_DISABLE;       //disable pull-up mode
+
+    gpio_config(&io_conf);                          //configure GPIO with the given settings
+}
+
+extern "C" void wakeup_modem(void)
```
david-cermak, line 80: same here
```suggestion
void wakeup_modem(void)
```

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
```diff
@@ -145,6 +179,26 @@ extern "C" void app_main(void)
...
+
+                /* Set CMUX mode */
+                if (!dce->set_mode(mode_rtc)) {
+                    ESP_LOGE(TAG, "Failed to set the desired mode");
+                }
+            }
+            break;
+        case ESP_SLEEP_WAKEUP_UNDEFINED:
+        default:
+            ESP_LOGI(TAG, "Not a deep sleep reset\n");
```
david-cermak, line 199: Since I assume this example will be mostly used without deep sleep
```suggestion
            ESP_LOGD(TAG, "Not a deep sleep reset\n");
```

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
```diff
@@ -161,6 +215,7 @@ extern "C" void app_main(void)
                 dev_mode = esp_modem::modem_mode::DATA_MODE;
             } else if (mode == "CMUX") {
                 dev_mode = esp_modem::modem_mode::CMUX_MODE;
+                mode_rtc = dev_mode;
```
david-cermak, line 218: Please add a comment, that even if switching to `CMUX` fails, we make the DTE multiplex terminal. (this is potentially a bug and might be fixed eventually)

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
```diff
@@ -289,6 +347,80 @@ extern "C" void app_main(void)
...
+            ESP_LOGE(TAG, "Failed with ERROR: Invalid argument length");
+            return 1;
+        }
+
+        /* Get PSM config */
+        dce->at("AT+CPSMS?", out, 500);
+
+        /* Update PSM config */
+        char set_cmd[35];
+        sprintf(set_cmd, "AT+CPSMS=%c,,,\"%s\",\"%s\"", out.c_str()[8], periodic_tau.c_str(), active_time.c_str());
```
david-cermak, line 417: Suggest using `asprintf()` or better yet some C++ construct..
```suggestion
        std::string cmd =  "AT+CPSMS=..." +  periodic_tau + "," ...
```

Also, please explain `out.c_str()[8]` -- could we check if `out` really contains 8 bytes?

david-cermak (changes requested):
LGTM in general, just few minor comments

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
espressif-abhikroy: asprintf() needs to have a free in the end, but the next tile CHECK_ERR(... does a return. So there's no scope for a free. That's why I allocated a buffer and used sprint. 
Also, for Sim70xx series the output of AT+CPSMS? is of a fixed length of 35 bytes.

espressif-abhikroy: Done

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
espressif-abhikroy: Done

components/esp_modem/examples/modem_console/main/modem_console_main.cpp
david-cermak: Ok, Thanks for updates,

Could you also please check if `out.size() > 8`, just to be sure we could dereference `out.c_str()[8]`

espressif-abhikroy: Done

david-cermak (approved):

david-cermak:
@espressif-abhikroy Thanks for the updates!

components/esp_modem/examples/modem_console/sdkconfig.defaults
```diff
@@ -1,8 +1,12 @@
-# Override some defaults to enable PPP
+# This file was generated using idf.py save-defconfig. It can be edited manually.
+# Espressif IoT Development Framework (ESP-IDF) Project Minimal Configuration
+#
+CONFIG_IDF_TARGET="esp32c3"
+CONFIG_EXAMPLE_MODEM_DEVICE_SIM7600=y
+CONFIG_EXAMPLE_MODEM_UART_TX_PIN=4
+CONFIG_EXAMPLE_MODEM_UART_RX_PIN=5
+CONFIG_EXAMPLE_MODEM_UART_EVENT_TASK_STACK_SIZE=4096
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=7168
```
tore-espressif, line 9: Please delete all fields that are overriding defaults from Kconfig. And the IDF_TARGET too.

PR #157, opened 2022-10-10
fix(esp-modem): On read/command callback fixes

david-cermak:
* Made lambda to capture `std::function<>` by value (to control lifetime of the functor)
* Wait for a signal (*clear to update*) before modifying on-read callback


euripedesrocha (approved):

david-cermak:
thanks for the review! @euripedesrocha

PR #162, opened 2022-10-13
CI: Updated repo-checkout flag to consider submodules (IDFGH-8515)

gabsuren:


david-cermak (approved):

david-cermak:
@gabsuren I'm sorry I was too fast merging this and didn't realize that we should like to bump the component version

PR #122, opened 2022-08-23
extended pppos_client (IDFGH-8392)

diplfranzhoepfinger:
extended pppos_client

PR #119, opened 2022-08-23
get_gnss_information_sim70xx added ! (IDFGH-8118)

diplfranzhoepfinger:
3rd Try.

PR #165, opened 2022-10-17
Accept some long standing example/demo PRs

david-cermak:
This cherry-picks some PRs that we intend to accept, but keep struggling with minor problems, cosmetic changes and petty issues:
* https://github.com/espressif/esp-protocols/pull/150
* https://github.com/espressif/esp-protocols/pull/122
* https://github.com/espressif/esp-protocols/pull/119


components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/SIM7070_gnss.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "cxx_include/esp_modem_dce_factory.hpp"
+#include "cxx_include/esp_modem_dce_module.hpp"
+#include "nmea_parser.h"
+
+/**
+ * @brief Definition of a custom modem which inherits from the GenericModule, uses all its methods
+ * and could override any of them. Here, for demonstration purposes only, we redefine just `get_module_name()`
+ */
+class SIM7070_gnss: public esp_modem::SIM7070 {
```
euripedesrocha, line 22: Could these two classes be merged?

components/esp_modem/include/cxx_include/command_library.hpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * SPDX-FileCopyrightText: 2021-2022 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+namespace esp_modem::dce_commands {
+command_result generic_command(CommandableIf *t, const std::string &command,
+                               const std::list<std::string_view> &pass_phrase,
```
euripedesrocha, line 10: `std::string_view`  is only available after C++17.

components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/SIM7070_gnss.cpp
```diff
@@ -0,0 +1,340 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+//
+// Created on: 23.08.2022
+// Author: franz
+
+
+#include <cstdio>
+#include <cstring>
+#include <map>
```
euripedesrocha, line 13: Seems that `<map>` is unused.

components/esp_modem/examples/pppos_client/main/pppos_client_main.c
```diff
@@ -214,7 +240,7 @@ void app_main(void)
     int rssi, ber;
     esp_err_t err = esp_modem_get_signal_quality(dce, &rssi, &ber);
     if (err != ESP_OK) {
-        ESP_LOGE(TAG, "esp_modem_get_signal_quality failed with %d", err);
+        ESP_LOGE(TAG, "esp_modem_get_signal_quality failed with %d %s", err, esp_err_to_name(err));
```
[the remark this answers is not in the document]
david-cermak: I think it is, as it's possible to disable error-to-string in menuconfig (so save some memory) and in that case `esp_err_to_name()` returns a generic message `"ERROR"` or similar. For this config, it's useful to see the error number as well.

components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/SIM7070_gnss.hpp
david-cermak: They were one class originally, but I specifically asked to make two classes, since the `GNSS` is a specialization of the `7070`. Alternatively we could "compose" the `SIM7070_gnss` somehow statically to avoid the inheritance here (but I personally find the inheritance more straight forward)

david-cermak:
Thanks for the suggestions and review @euripedesrocha ! Could you please check the unresolved discussion?

components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/SIM7070_gnss.hpp
euripedesrocha: I suggest adding a description on both classes and why they are needed. Also, the description above `SIM7070_gnss` is correct?

components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/nmea_parser.h
```diff
@@ -0,0 +1,137 @@
...
+    float dop_v;                                                   /*!< Vertical dilution of precision  */
+    uint8_t sats_in_view;                                          /*!< Number of satellites in view */
+    gps_date_t date;                                               /*!< Fix date */
+    float speed;                                                   /*!< Ground speed, unit: m/s */
+    float cog;                                                     /*!< Course over ground */
+    float hpa;                                                     /*!< Horizontal Position Accuracy  */
+    float vpa;                                                     /*!< Vertical Position Accuracy  */
+};
+
+typedef struct gps_s gps_t;
```
euripedesrocha, line 120: Could we remove one of these? It can be confusing to have two different alias for the same type.

components/esp_modem/include/cxx_include/esp_modem_command_library_utils.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @brief Generic command that passes on "OK" and fails on "ERROR"
+ * @param t Anything that is "command-able"
+ * @param command Command to issue
+ * @param timeout
+ * @param timeout_ms Command timeout in ms
+ * @return Generic command return type (OK, FAIL, TIMEOUT)
+ */
+command_result generic_command_common(CommandableIf *t, const std::string &command, uint32_t timeout = 500);
```
euripedesrocha, line 43:
```suggestion
command_result generic_command_common(CommandableIf *t, const std::string &command, uint32_t timeout_ms = 500);
```

components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -10,15 +10,17 @@
...
 
 namespace esp_modem::dce_commands {
 
 static const char *TAG = "command_lib";
 
-command_result generic_command(CommandableIf *t, const std::string &command,
-                               const std::list<std::string_view> &pass_phrase,
-                               const std::list<std::string_view> &fail_phrase,
-                               uint32_t timeout_ms)
+static command_result generic_command(CommandableIf *t, const std::string &command,
```
euripedesrocha, line 20: Nitpick: `static` function could be in an unnamed namespace.

components/esp_modem/examples/simple_cmux_client/components/SIM7070_gnss/SIM7070_gnss.hpp
david-cermak: Thanks, that was a copy paste.

Updated the description of both classes.

components/esp_modem/src/esp_modem_command_library.cpp
david-cermak: Tried to enclose it in an anonymous namespace, but IMHO it makes the module less readable, as I'd have to split the API to `static`s and "others". I would keep it as is, if you agree, since they're already enclosed in their own namespaces.

david-cermak:
Thanks for your reviews, Euripedes, Franz!

PR #172, opened 2022-10-31
websocket: updated example to use local websocket echo server

gabsuren:
Updated websocket client example, to use local server


david-cermak (approved):

david-cermak:
Thanks for the quick fix Suren!

PR #173, opened 2022-11-07
mdns: fix some coverity reported issues

mahavirj:
- Array compared against 0, code with NO_EFFECT
- dereference null pointer, FORWARD_NULL check


david-cermak (approved):
Thanks for the fixes!

PR #156, opened 2022-10-07
feat(esp_modem): Add mqtt example in  AT-only mode

david-cermak:
Similar to the `pppos-client`, but without PPP mode.
This uses standard mqtt client and a loopback service that forwards the TCP traffic from localhost to the modem and vice-versa. (next step is to create a dedicated tcp-transport layer for modem)

```
I (525) sock_dce: Got IP +IPADDR: x.x.x.x
I (525) sock_dce: Socket created
I (525) sock_dce: Socket bound, port 1883
I (525) system_api: Base MAC address is not set
I (535) system_api: read default base MAC address from EFUSE
I (535) modem_client: MQTT other event id: 7
I (965) sock_dce: Connected!
I (1015) modem_client: MQTT_EVENT_CONNECTED
I (1025) modem_client: sent subscribe successful, msg_id=17361
I (1095) modem_client: MQTT_EVENT_SUBSCRIBED, msg_id=17361
I (1095) modem_client: sent publish successful, msg_id=0
I (1155) modem_client: MQTT_EVENT_DATA
TOPIC=/topic/esp-pppos
DATA=esp32-pppos
E (121345) mqtt_client: No PING_RESP, disconnected
I (121345) modem_client: MQTT_EVENT_DISCONNECTED
E (122345) sock_dce: Failed to get idle
E (122345) modem_client: Loop exit.. retrying
I (126885) sock_dce: Got IP +IPADDR: y.y.y.y
I (127115) sock_dce: Connected!
I (136345) modem_client: MQTT other event id: 7
I (136415) modem_client: MQTT_EVENT_CONNECTED
I (136415) modem_client: sent subscribe successful, msg_id=1196
I (136485) modem_client: MQTT_EVENT_SUBSCRIBED, msg_id=1196
I (136495) modem_client: sent publish successful, msg_id=0
I (136575) modem_client: MQTT_EVENT_DATA
TOPIC=/topic/esp-pppos
DATA=esp32-pppos
```

PR #179, opened 2022-11-24
URC Reception DRAFT (IDFGH-8813)

diplfranzhoepfinger:
URC Reception DRAFT


diplfranzhoepfinger:
hi
@david-cermak 


here a Example where we would need URC Reception. 

saw your https://github.com/espressif/esp-protocols/pull/156 but unsure how to use it. 

here: get_gnss_information_sim70xx_once_req() is always returning OK, 
while the +SGNSCMD: is sent as an URC. 

i have already prepared the Parser in the Function get_gnss_information_sim70xx_lib_once but the Question is how could it be called over a URC !

also other URCs are interesting, like drop of Network etc..

diplfranzhoepfinger:
@david-cermak other Question: 

```
void set_mode_and_report(std::unique_ptr<esp_modem::DCE> & dce, modem_mode m)
```

```
void set_mode_and_report(std::unique_ptr<DCE_gnss> & dce, modem_mode m)
```

is there a way to write such in a common way, so i do not need to change this, independent of the Modem Type ? 

Thanks, 

Franz

david-cermak:
Closing as https://github.com/espressif/esp-protocols/pull/156 has been merged

PR #199, opened 2022-12-22
Update esp_modem_command_library.cpp (IDFGH-9018)

franz-ms-muc:
AT+IFC=2,2 instead AT+IFC=2, 2


david-cermak (approved):
Thanks for the fix!

franz-ms-muc:
courios: on a SIM7070 it is running with the old version, but a **A7672E** (**FASE**) was failing, and only running without the space.

franz-ms-muc:
courios: on a SIM7070 it is running with the old version, but a **A7672E** (**FASE**) was failing, and only running without the space.

david-cermak:
> courios: on a SIM7070 it is running with the old version, but a **A7672E** (**FASE**) was failing, and only running without the space.

That's probably the reason why it slipped though (as I was apparently testing it with 7600 or similar which accepts the space). Thanks!

Issue #190, opened 2022-12-11
Misconfigured mDNS causes ESP32 hang with LoadProhibited error (IDFGH-8933)


wuyuanyi135:
Hi, I was using the code from the example but accidentally commented out the `mdns_hostname_set` line. 

The code run without error (I `ESP_ERROR_CHECK`ed everything). But once the mDNS request was initiated, the ESP responded once (saw from Wireshark PCAP), then stop responding. And `idf.py monitor` (esp32s3 built-in JTAG) hanged without printing panic messages. Also, the `monitor` is not able to exit by `CTRL+]` until I unplug the USB and then it sometimes printed an incomplete message with `LoadProhibited error`.

This is much like a stackoverflow behavior I have seen before.  I know this is error is due to my misconfiguration, but I think there should be some checking to prevent such a debug puzzle?

```
void start_mdns_service()
{
    //initialize mDNS service
    esp_err_t err = mdns_init();
    if (err) {
        printf("MDNS Init failed: %d\n", err);
        return;
    }

    //set hostname
    // mdns_hostname_set("my-esp32");   // <-------- Accidentally commented out
    //set default instance
    mdns_instance_name_set("Jhon's ESP32 Thing");
}
```

david-cermak:
@wuyuanyi135 Thanks for the report. We're adding error codes when initializing mDNS in wrong order, but it was probably missed in `mdns_instance_name_set()` API. will fix.

wuyuanyi135:
Thanks for your reply!

PR #206, opened 2023-01-18
fix(mdns): Make unit test executable with pytest

david-cermak:
* Rework unit tests to pytest
* Add build step for unit tests

---
Added two more commits:
1) fix(mdns): Allow setting instance name only after hostname set
    - Closes https://github.com/espressif/esp-protocols/issues/190

2) fix(mdns): Remove strict mode as it's invalid

    Strict mode was introduced to support "one-shot" queries (described in
    [RFC6762/sec5.1](https://www.rfc-editor.org/rfc/rfc6762.html#section-5.1)) that are sent by lwip or dig. It was incorrectly assumed
    that responding to such queries violates the spec, as we have to repeat
    queries in responces, which is forbidden in RFC6762/sec6. It is however
    required to repeat query fields according to the Section 6.7. [Legacy
    Unicast Responses](https://www.rfc-editor.org/rfc/rfc6762.html#section-6.7): "it MUST repeat the query ID and the question
    given in the query message."


.github/workflows/target-test.yml
```diff
@@ -256,28 +236,27 @@ jobs:
...
         env:
           PIP_EXTRA_INDEX_URL: "https://www.piwheels.org/simple"
         run: |
             sudo apt-get install -y dnsutils
-      - name: Download Example Test to target ${{ matrix.config }}
-        run: |
-            python -m esptool --chip ${{ matrix.idf_target }} write_flash 0x0 components/mdns/examples/build_${{ matrix.config }}/flash_image.bin
-      - name: Run Example Test on target ${{ matrix.config }}
-        working-directory: components/mdns/examples
+      - name: Run Example Test on target ${{ matrix.idf_target }}
```
gabsuren, line 246: Wouldn't it be better to somehow seperate unit test target tests and example tests? From CI perspective it will be more visible which one failed

.github/workflows/target-test.yml
```diff
@@ -62,48 +62,30 @@ jobs:
...
-          rm -rf sdkconfig sdkconfig.defaults build build_${{ matrix.config }}
-          cat sdkconfig.ci.${{ matrix.config }} >> sdkconfig.defaults
-          idf.py set-target ${{ matrix.idf_target }}
-          idf.py build
-          mv build build_${{ matrix.config }}
-      - name: Merge binaries with IDF-${{ matrix.idf_ver }} for ${{ matrix.config }}
-        working-directory: ${{ env.TEST_DIR }}
-        env:
-          IDF_TARGET: ${{ matrix.idf_target }}
+      - name: Build ${{ matrix.example }} with IDF-${{ matrix.idf_ver }} for ${{ matrix.idf_target }}
```
euripedesrocha, line 73: The  `matrix` don't have example defined, should it be `matrix.test.app`?

.github/workflows/target-test.yml
david-cermak: This is true, we should say what kind of test we're executing. Updated to use the matrix variable.

.github/workflows/target-test.yml
david-cermak: Thanks, fixed!

gabsuren (approved):

david-cermak:
Thanks for the review, Suren!

PR #209, opened 2023-01-19
Added the missing definition for the esp_modem_set_baud function (IDFGH-9181)

### LINKED ISSUE #208: Issues with `esp_modem_set_baud` in `esp_modem` component (IDFGH-9178)
Documentation [suggests](https://espressif.github.io/esp-protocols/esp_modem/api_docs.html#modem-commands) that there is a C function with the signature `command_result esp_modem_set_baud(int baud)`, but compilation fails with the error `too few arguments to function 'esp_modem_set_baud'` when I try to use it. 

From what I understand, the macro used to generate `esp_modem_set_baud`, generates the function with the signature `command_result esp_modem_set_baud(esp_modem_dce_t* dce, int baud)`. This seems like a recurring thing in the documentation, as all C-API functions seems to omit the first parameter `esp_modem_dce_t*`

Furthermore, although a declaration for the symbol `command_result esp_modem_set_baud(esp_modem_dce_t* dce, int baud)` is generated, it seems like there is no actual definition for it available, which fails the linkage with an undefined symbol error.

term-est:
This pull request should close #208

If I am not missing anything, it seems like the macro used to generate the C-API creates a signature for `esp_modem_set_baud ` function, but there is no definition for it available anywhere. I added a quick definition, which should solve the problem.

I've compiled some of the examples and used `esp_modem_set_baud` with success with this change.

Thank you for taking your time to review this pull request and have a great day!

### DIFF
--- a/components/esp_modem/src/esp_modem_c_api.cpp
+++ b/components/esp_modem/src/esp_modem_c_api.cpp
@@ -419,7 +419,7 @@ extern "C" esp_err_t esp_modem_command(esp_modem_dce_t *dce_wrap, const char *co
     }, timeout_ms));
 }
 
-extern "C" esp_err_t esp_modem_set_baud(esp_modem_dce_t* dce_wrap, int baud)
+extern "C" esp_err_t esp_modem_set_baud(esp_modem_dce_t *dce_wrap, int baud)
 {
-	return command_response_to_esp_err(dce_wrap->dce->set_baud(baud));
+    return command_response_to_esp_err(dce_wrap->dce->set_baud(baud));
 }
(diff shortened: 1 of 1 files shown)


david-cermak (approved):
Thanks for implementing the missing function!

components/esp_modem/src/esp_modem_c_api.cpp
```diff
@@ -418,3 +418,8 @@ extern "C" esp_err_t esp_modem_command(esp_modem_dce_t *dce_wrap, const char *co
         }
     }, timeout_ms));
 }
+
+extern "C" esp_err_t esp_modem_set_baud(esp_modem_dce_t* dce_wrap, int baud)
+{
+	return command_response_to_esp_err(dce_wrap->dce->set_baud(baud));
+}
```
david-cermak, line 425: you might need to install the pre-commit hook to help you with some formatting nitpicks, here I think, we miss a newline

term-est: Thank you for your feedback! I've installed the pre-commit and let that run. Sorry for not properly reading the Contribution guidelines

david-cermak: Thanks for the quick fix! Merging with squash function to keep the history clean.

PR #211, opened 2023-01-23
websocket: Added new API `esp_websocket_client_destroy_selected`

gabsuren:


components/esp_websocket_client/include/esp_websocket_client.h
```diff
@@ -166,6 +166,16 @@ esp_err_t esp_websocket_client_stop(esp_websocket_client_handle_t client);
...
 
+/**
+ * @brief      If `selected` set to True, WebSocket client will destroy and free all resources at the end of event loop.
+ *
+ * @param[in]  client      The client
+ * @param[in]  selected    The flag
+ *
+ * @return     esp_err_t
+ */
+esp_err_t esp_websocket_client_destroy_selected(esp_websocket_client_handle_t client, bool selected);
```
david-cermak, line 177: Does it make sense to add the `bool selected` parameter? Why would we call: `esp_websocket_client_destroy_selected(..., false)` ?

components/esp_websocket_client/include/esp_websocket_client.h
```diff
@@ -166,6 +166,16 @@ esp_err_t esp_websocket_client_stop(esp_websocket_client_handle_t client);
  */
 esp_err_t esp_websocket_client_destroy(esp_websocket_client_handle_t client);
 
+/**
+ * @brief      If `selected` set to True, WebSocket client will destroy and free all resources at the end of event loop.
+ *
```
david-cermak, line 171: We should warn users, that the client handle would be dangling and should never be used

components/esp_websocket_client/include/esp_websocket_client.h
david-cermak, line 177: Suggest renaming this to something more meaningful? `destroy_delayed()` ?

david-cermak (changes requested):
LGTM in general, but please create a unit test with `leaks==0` to double check that there's nothing forgotten to de-allocate.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -342,6 +343,29 @@ static esp_err_t esp_websocket_client_destroy_config(esp_websocket_client_handle
...
+    esp_websocket_client_destroy_config(client);
+    esp_transport_list_destroy(client->transport_list);
+    vQueueDelete(client->lock);
+    free(client->tx_buffer);
+    free(client->rx_buffer);
+    if (client->status_bits) {
+        vEventGroupDelete(client->status_bits);
+    }
+    free(client);
+    client = NULL;
```
david-cermak, line 366: Question: Could this part be reused in the standard `esp_websocket_client_destroy()` ?

components/esp_websocket_client/include/esp_websocket_client.h
gabsuren: @david-cermak not really sure about such scenario just wanted to let the users freedom to change the flag in case if it needed in some case. But without `bool selected` parameter it will be more convenient from API perspective and this part https://github.com/espressif/esp-protocols/pull/211#discussion_r1084946518 part can be reused in that case. So will modify to make it without parameter. Thanks!

Issue #101, opened 2022-07-29
Feature request: mDNS: Add reverse lookup (IDFGH-7941)


hurricanefrog:
Currently, the mDNS component / library doesn't seem to support reverse lookup.

i.e. I get no reply to `dig -p 5353 -x <myip> @224.0.0.251`

Could this be added?

david-cermak:
Yes, reverse lookup is currently not supported by the mDNS library. We are adding it to the road-map, but please understand that it doesn't have a very high priority, I expect the implementation to take place sometime in January.

PR #219, opened 2023-02-07
feat(mdns): Add support for IPv6 reverse query

david-cermak:
Implements reverse queries together with b87bef52e5e94d9142d09b129c68a93228aa64f2 (IPv4)

Closes https://github.com/espressif/esp-protocols/issues/101


david-cermak:
Reverse query in CI:
* [IPv4](https://github.com/espressif/esp-protocols/actions/runs/4127638367/jobs/7131203333#step:6:213)
* [IPv6](https://github.com/espressif/esp-protocols/actions/runs/4127638367/jobs/7131203333#step:6:217)

gabsuren (approved):

components/mdns/examples/main/mdns_example_main.c
```diff
@@ -276,16 +276,15 @@ void app_main(void)
      * This is typically performed in "GOT_IP" event handler, but we call it here directly
      * since the `EXAMPLE_INTERFACE` netif is connected already, to keep the example simple.
      */
-    ESP_ERROR_CHECK(mdns_netif_action(EXAMPLE_INTERFACE, MDNS_EVENT_ENABLE_IP4));
+    ESP_ERROR_CHECK(mdns_netif_action(EXAMPLE_INTERFACE, MDNS_EVENT_ENABLE_IP4 | MDNS_EVENT_ENABLE_IP6));
```
euripedesrocha, line 279: We enable both and announce only IPV4?

components/mdns/examples/pytest_mdns.py
```diff
@@ -165,11 +171,25 @@ def test_examples_protocol_mdns(dut):
...
+                dig_output = subprocess.check_output([
+                    'dig', '+short', '-p', '5353', '@224.0.0.251', '-x',
+                    '{}'.format(ip_address)
+                ])
+                print('Reverse lookup for {} using "dig" succeeded with:\n{}'.
+                      format(ip_address, dig_output))
+                if specific_host not in dig_output.decode():
+                    raise ValueError(
+                        'Test has failed: Incorrectly resolved DUT IP address using dig'
+                        "Output should've contained DUT's name:{}".format(
```
euripedesrocha, line 190: Is it possible to get also the return here and avoid the print ? Print is useful for debugging, so I think it's better to have it in the Exception message.

david-cermak: You mean removing the `dig`'s outputs from the previous print and adding it here to the exception message?

Updated, PTAL. I sill left the `print()` message with basic information (that the lookup succeeded) as I find it useful even if the test passes.

david-cermak: I've decided to revert this change, as I find it quite useful to see the actual output of the tool, even in the "happy flow" scenario.
Compare these:
* [prints just info](https://github.com/espressif/esp-protocols/actions/runs/4177245710/jobs/7242835054#step:6:215)
* [prints the result too](https://github.com/espressif/esp-protocols/actions/runs/4127638367/jobs/7131203333#step:6:213)

we're calling it with  `--short` argument, so it prints one more line, which is concise and useful IMO.

euripedesrocha: Yes exactly, I was concerned that the print may get lost in case of error and that it may not be that useful in successful runs.  But as you state it is in fact useful better to have both.

PR #222, opened 2023-02-14
components/mdns: use semaphore instead of task notification bits (IDFGH-9380) (IDFGH-9388)

boborjan2:
PR for https://github.com/espressif/esp-idf/issues/10754

Component should not alter the task notification bits of the calling task.

### DIFF
--- a/components/mdns/private_include/mdns_private.h
+++ b/components/mdns/private_include/mdns_private.h
@@ -380,6 +380,7 @@ typedef struct mdns_server_s {
     mdns_srv_item_t *services;
     SemaphoreHandle_t lock;
     QueueHandle_t action_queue;
+    SemaphoreHandle_t action_sema;
     mdns_tx_packet_t *tx_queue_head;
     mdns_search_once_t *search_once;
     esp_timer_handle_t timer_handle;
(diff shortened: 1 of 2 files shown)


david-cermak (approved):
Thanks for the quick fix!

components/mdns/private_include/mdns_private.h
```diff
@@ -390,7 +391,6 @@ typedef struct {
     union {
         struct {
             char *hostname;
-            TaskHandle_t calling_task;
```
david-cermak, line 393: Alternatively, we can add the semaphore here
```patch
    struct {
            char *hostname;
-            TaskHandle_t calling_task;
+            SemaphoreHandle_t hostname_ready;
        } hostname_set;
        char *instance;
```

And create it only when setting a new hostname, and delete afterwards  (advantage is that we save a few bytes of heap, when the system is up and running).

Just an optional suggestion to think about. The changes LGTM!

boborjan2: I would rather not allocate a semaphore in an API call like that. This semaphore might as well be used by other actions in the future(?). Btw, seemingly mdns_server->lock is not used at all.

david-cermak: > This semaphore might as well be used by other actions in the future(?).

Yes, that's quite possible.

> Btw, seemingly mdns_server->lock is not used at all.

Oh, good point, thanks! This mutex is indeed just allocated but never locked! (probably better idea to save those couple of bytes here!)

PR #230, opened 2023-02-23
esp_modem: WIP! return true from on_data callback in data mode

tore-espressif:
Very much WIP to see result of CI tests.

No need to review now


tore-espressif:
Divided into #231 and #243

PR #231, opened 2023-02-23
esp_modem: Return true from on_data callback in data mode

tore-espressif:
* The DTE's on_data callback should return false only if there is a problem with the received data.
* The updated UART Terminal implementation cannot longer clear its callback. It is DTE's responsibility to clear the callback.

As a result, the UART implementation ignores the returned value from on_data callback. However, this feature is important for USB DTE implementation that handles RX data differently.

PTAL @david-cermak

### DIFF
--- a/components/esp_modem/idf_component.yml
+++ b/components/esp_modem/idf_component.yml
@@ -1,4 +1,4 @@
-version: "0.1.26"
+version: "0.1.27"
 description: esp modem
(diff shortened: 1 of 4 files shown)


david-cermak (approved):

david-cermak:
Thanks for fixing this @tore-espressif !

The UART terminal was trying to reset the callback on completion (to avoid races when commands completed), but I agree that it must be DTE's responsibility to handle it, not an implementation detail of the supported terminals.

PR #237, opened 2023-02-28
fix: Add license file for `mdns` and `esp_mqtt_cxx` components

gabsuren:


david-cermak:
Thanks Suren, but closing as the license files were added by Sergei, including the url links to component manager files.

PR #239, opened 2023-03-01
Add homepage URL and License to all components (IDFGH-9521)

kumekay:
Add `url` fields to `idf_component.yml` manifests and `LICENSE` file with Apache 2 license to all components


components/asio/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
david-cermak, line 4: I think we cannot distribute the `asio` component under Apache License. The library's original license is BSL 1.0: https://github.com/chriskohlhoff/asio/blob/master/asio/LICENSE_1_0.txt

david-cermak (approved):

components/asio/LICENSE
kumekay: You are right. I've changed the license. Thank you for the review!

david-cermak:
Thanks Sergei for your help, here!

PR #242, opened 2023-03-07
fix(esp_modem): run CI build job for all targets

david-cermak:
Simplified the workflow, use `idf-build-apps` package to find apps for all supported targets.

Additional changes:
* removed IDFv4.1 from tests
* added common build_apps.py utility for finding/building apps
* restructured workflow files and rules:
    * `<component>__<test_type>.yml` -- component related test workflow (need the component label to run it in PR)
    * `common_check.yml` -- runs unconditionally

Future changes:
* all components to use local `build_apps.py` for all targets
* simplify `upload_artifact` job using `ci/clean_build_artifacts.sh` (as done in `esp_modem` target tests)
* build for all combination of supported IDF versions per component


.github/workflows/modem__build-host-tests.yml
```diff
@@ -0,0 +1,173 @@
+name: "esp-modem: build/host-tests"
```
gabsuren, line 1: As we are merging host test and coverage tests into one file, I would suggest to add/update title although to show that information

.github/workflows/modem__build-host-tests.yml
```diff
@@ -0,0 +1,173 @@
...
+          EXPECTED_WARNING: ${{ matrix.warning }}
+        shell: bash
+        run: |
+          . ${IDF_PATH}/export.sh
+          python -m pip install idf-build-apps
+          cd $GITHUB_WORKSPACE/protocols
+          python ./ci/build_apps.py components/esp_modem/examples/${{ matrix.example }}
+
+  host_test_esp_modem:
+    if: contains(github.event.pull_request.labels.*.name, 'modem')
```
gabsuren, line 50: nitpick: `modem` or `esp_modem`?

.github/workflows/mdns__host-tests.yml
```diff
@@ -0,0 +1,61 @@
+name: "mdns: host-tests"
```
gabsuren, line 1: here although we building host and AFL tests in one file, will be nice to show in title

gabsuren (changes requested):
Just a question regarding full test set running. Do we need such label also?

.github/workflows/modem__build-host-tests.yml
david-cermak: But these "coverage tests" are still host tests, correct? The title says that this file defines **all** workflows related to build and host tests (including the host tests with gcov enabled). The info about running the coverage analysis is still in the respective job name:
https://github.com/espressif/esp-protocols/pull/242/files#diff-2b92012e9e84fa80ea1bcff4a12777d87bcfef169dd05b9d9a39946e97d25261R88

.github/workflows/mdns__host-tests.yml
david-cermak: Same here. Naming mentioned below https://github.com/espressif/esp-protocols/pull/242/files#diff-76592f40f1f57d7e25dc47a9c8e919d9801e57f323dad7fbcc5dce5845331093R37
(since these are not AFL tests, just checkers that we're still able to compile when we want to run the AFL tests)

.github/workflows/modem__build-host-tests.yml
david-cermak: I'd prefer just `modem` (removed the `esp-` prefix for convenience, both in the workflow files, as well as in the labels.

gabsuren (approved):

gabsuren:
LGTM! 
2 questions
1. Wouldn't it be nice also have a label `run_all` or smt. to be able run all test sets?
2. When merging to the `master` all test sets should be run automatically, in this case will they be executed ?

david-cermak:
Thanks for the review, both points make sense -- will address, thanks!

> 1. Wouldn't it be nice also have a label `run_all` or smt. to be able run all test sets?

good point, will add support for this new label to run everything.

> 2. When merging to the `master` all test sets should be run automatically, in this case will they be executed in that way?

ATM it would check the modification paths to run only relevant tests. But I agree that we should check all components after merge to keep `master` branch clean,. will fix,

euripedesrocha (approved):

david-cermak:
> 1. Wouldn't it be nice also have a label `run_all` or smt. to be able run all test sets?

I've decided not to implement one common label, since it would unnecessarily complicate the conditions (and would need to apply it everywhere). we can always add it in future if needed. for now, I don't think we'd have too many PRs affecting multiple components and if so, we can add more labels.

```patch
-    if: contains(github.event.pull_request.labels.*.name, 'mdns')
+    if: $${ contains(github.event.pull_request.labels.*.name, 'mdns') && contains(github.event.pull_request.labels.*.name, 'run_all') }}
```

> 2. When merging to the `master` all test sets should be run automatically, in this case will they be executed in that way?

Done.

PR #243, opened 2023-03-07
esp_modem: Dual DTE support

tore-espressif:
Modems can expose 2 terminals, which can be used simultaneously. One for AT commands, the other one for data.

* Added new DTE constructors that accept two instances of `Terminal` class
* Added new DTE state 'DUAL'

Tested with USB modem SimCom A7670E

@david-cermak PTAL

### DIFF
--- a/components/esp_modem/src/esp_modem_dce.cpp
+++ b/components/esp_modem/src/esp_modem_dce.cpp
@@ -88,6 +88,7 @@ bool DCE_Mode::set_unsafe(DTE *dte, ModuleIf *device, Netif &netif, modem_mode m
 {
     switch (m) {
     case modem_mode::UNDEF:
+    case modem_mode::DUAL_MODE: // Only DTE can be in Dual mode
         break;
     case modem_mode::COMMAND_MODE:
         if (mode == modem_mode::COMMAND_MODE || mode >= modem_mode::CMUX_MANUAL_MODE) {
--- a/components/esp_modem/idf_component.yml
+++ b/components/esp_modem/idf_component.yml
@@ -1,4 +1,4 @@
-version: "0.1.27"
+version: "0.1.28"
 description: esp modem
 url: https://github.com/espressif/esp-protocols/tree/master/components/esp_modem
 dependencies:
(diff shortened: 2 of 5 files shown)


components/esp_modem/src/esp_modem_dce.cpp
```diff
@@ -88,6 +88,7 @@ bool DCE_Mode::set_unsafe(DTE *dte, ModuleIf *device, Netif &netif, modem_mode m
 {
     switch (m) {
     case modem_mode::UNDEF:
+    case modem_mode::DUAL_MODE: // Only DTE can be in Dual mode
```
david-cermak, line 91: Question: When we construct the DTE in `DUAL_MODE`, which transitions are supported for `DCE` ? Could we get out of sync? How do we switch the second terminal to the `DATA_MODE`?

david-cermak (approved):
Changes LGTM, just left one question about mode transitions from user perspective (DCE).

components/esp_modem/src/esp_modem_dce.cpp
tore-espressif: The 2 terminals in USB modems are equal -> it does not matter which is used for data or commands.

So `set_mode(DATA)` connects the modem to cellular network: primary terminal is switched to data mode and swapped with secondary terminal.
And `set_mode(CMD)` disconnects the modem from cellular network.
CMUX is not supported (but not explicitly forbidden).

> Could we get out of sync?

In case ESP resets and connects to modem that is already in data mode, we will be out of sync. (I don't know how this can be solved in `esp_modem` :/)

Does it answer your questions?

tore-espressif:
> Changes LGTM, just left one question about mode transitions from user perspective (DCE).

Thank you for the quick review @david-cermak !
Generally speaking, this is a minimum viable feature for dual-terminal modems, I expect that the feature will be extended as we test it with more devices. Is it OK? Or should we do more testing?

components/esp_modem/src/esp_modem_dce.cpp
david-cermak: I see, thanks for explaining!

> (I don't know how this can be solved in esp_modem :/)

This is a general problem in `esp_modem` and we assume the initial mode is `UNDEF`, I've planned to implement the mode detection mechanism on startup, which would address this issue. But it's unrelated to your changes.

david-cermak:
> Generally speaking, this is a minimum viable feature for dual-terminal modems, I expect that the feature will be extended as we test it with more devices. Is it OK? Or should we do more testing?

Yes, we can merge it as is. Thanks for adding support for two *real* terminals!

PR #246, opened 2023-03-09
make GPS Data Structures more Module specific (IDFGH-9962)

diplfranzhoepfinger:
make GPS Data Structures more Module specific


david-cermak:
Thanks for this PR, we'll accept it via https://github.com/espressif/esp-protocols/pull/531

PR #254, opened 2023-03-15
MQTT: complete client authentication configuration (IDFGH-9656)

lhauswald:
Configuration for client authentication methods certificate/key and secure element added.


euripedesrocha (approved):

euripedesrocha:
Hi @lhauswald thanks for the addition. 

Could you just adjust the commit message to the correct scope? It should be `mqtt_cxx` instead of `mqtt`.

PR #261, opened 2023-03-28
Improve and cleanup SIM7070 (IDFGH-9965)

diplfranzhoepfinger:


david-cermak:
Part of the changes are going to be accepted via https://github.com/espressif/esp-protocols/pull/531

Other changes only reflect overriding an URC handler of a custom device, this would be really confusing in the examples IMO. the ultimate solution would be the official URC support, which is coming soon.
I'm sorry, we cannot accept all the changes here.

PR #262, opened 2023-03-28
Feat/shiny as component (IDFGH-9967)

diplfranzhoepfinger:


diplfranzhoepfinger:
@david-cermak this is Modem as well

david-cermak:
Thanks for proposing this change, but I think this would be confusing to support another sub-component while we're just trying to showcase implementing a custom device here.

PR #263, opened 2023-03-30
mdns: Add negative replies (A/AAAA)

david-cermak:
According to RFC6762-6.1, mdns must provide NSEC record in the addtional fields, if the host exclusively owns the queried name, but has no answer for the queried type.


gabsuren (approved):

david-cermak:
Thanks for the review @gabsuren 

@xieqinan @antmak Could you PTAL, too?

(note that the CI failure is going to be fixed soon, after the latest IDF master goes public)

david-cermak:
Closing as the change is no longer needed

PR #266, opened 2023-04-04
[WIP] docs: improved script to deploy new version of documents component yml file changed

gabsuren:


.github/workflows/publish-docs-component.yml
```diff
@@ -25,24 +27,60 @@ jobs:
...
         shell: bash
         run: |
-            source $GITHUB_WORKSPACE/docs/utils.sh
-            add_doc_server_ssh_keys $DOCS_DEPLOY_KEY $DOCS_DEPLOY_SERVER $DOCS_DEPLOY_SERVER_USER
-            export GIT_VER=$(git describe --always)
-            export GITHUB_REF_NAME=latest
-            export DOCS_BUILD_DIR=$GITHUB_WORKSPACE/docs
-            deploy-docs
+            MASTER_BRANCH="master"
+            component_str=${{ env.COMPONENTS }}
```
david-cermak, line 40: could we get list of components from the `component` dir?

.github/workflows/publish-docs-component.yml
```diff
@@ -25,24 +27,60 @@ jobs:
...
+            IFS=',' read -a components <<< $component_str
+
+            for component in "${components[@]}"; do
+              YAML_FILE=components/$component/idf_component.yml
```
david-cermak, line 44: finding this quite complex, would this work instead?
```bash
for comp in `ls components`; do 
    ...
done
```

.github/workflows/publish-docs-component.yml
```diff
@@ -25,24 +27,60 @@ jobs:
...
+              # Get the current version of the file
+              CURRENT_VERSION=$(grep "^version:" ${YAML_FILE} | awk '{print $2}')
+
+              # Get the version of the file on the master branch
+              MASTER_VERSION=$(git show ${MASTER_BRANCH}:${YAML_FILE} | grep "^version:" | awk '{print $2}')
+
+              # Check if the major or minor version has changed
+              IFS='.' read -ra CURRENT_VERSION_PARTS <<< "${CURRENT_VERSION}"
+              IFS='.' read -ra MASTER_VERSION_PARTS <<< "${MASTER_VERSION}"
```
david-cermak, line 53: Need to check the entire version string, not only major and minor

david-cermak (changes requested):
I'm afraid this needs to be reworked. We should probably also *tag* version on master tree (merge commits)

PR #268, opened 2023-04-04
mdns: Add APIs to lookup delegated services (IDFGH-9790)

wqx6:
We need APIs which can search the delegated services.


wqx6:
> Also, I'd suggest renaming the API from `search` to `list`, as in this case, we're not searching for services, correct? we're just going over the internal structs to return already registered services.

Considering that we are look up the service list to find some services with specific type and proto. So I use `lookup` for these APIs. Is it OK?

wqx6:
Also I think the lookup can be synchronous since all of the actions are done locally. @david-cermak Any ideas?

david-cermak (approved):
Thanks for the update and making this a synchronous call!

PR #270, opened 2023-04-05
Fix deadly mdns crash (IDFGH-9800)

0xFEEDC0DE64:
A customer had hundrets of thousands of recorded crashes in the mdns code because answers was a nullptr.

We are inserting all esp crashes in our sentry which has the stack already decoded and ordered by fimrware version, esp-idf version and customer serial number.

I know if I recorded a malicious packet and sent it to espressif I could get a bounty price but I do this for charity now :)
I was too lazy to build a remote packet sniffer for the esp as driving to the customers site was not possible.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1614,18 +1614,20 @@ static void _mdns_remove_scheduled_answer(mdns_if_t tcpip_if, mdns_ip_protocol_t
     while (q) {
         if (q->tcpip_if == tcpip_if && q->ip_protocol == ip_protocol && q->distributed) {
             mdns_out_answer_t *a = q->answers;
-            if (a->type == type && a->service == service->service) {
-                q->answers = q->answers->next;
-                free(a);
-            } else {
-                while (a->next) {
-                    if (a->next->type == type && a->next->service == service->service) {
-                        mdns_out_answer_t *b = a->next;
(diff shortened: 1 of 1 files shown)


david-cermak (approved):
Thanks for your charitable contribution!

david-cermak:
@0xFEEDC0DE64 Thanks, this is a clear oversight from our side. One potential reason for *empty* answer field that comes to mind is if the device looses both IPv4 and IPv6 addresses, which is quite a fair scenario, e.g. after startup, before getting a DHCP lease.

> I was too lazy to build a remote packet sniffer 

Note that the `CONFIG_MDNS_ENABLE_DEBUG_PRINTS` is now part of menuconfig, so would give at least some pointer about the packet content at the time of crash

0xFEEDC0DE64:
I do not see the log output at the customers device as there is no serial or usb port on our products, and we never ship firmwares with log output enabled as this could help reverse engeneer parts of the firmware (especially with my modified logging that shows filenames and method names).

But thanks for the hint, if I experience a mdns crash on my development devices, this would be easier.

In the past I have fixed another mdns crash and added code so that when the crash would have happened, the mdns packet will be forwarded via UDP to my own server and be recorded so I can give the malicious packet to espressif :)

Issue #258, opened 2023-03-22
mDNS: the host does not support NSEC record. (IDFGH-9699)


xieqinan:
### Is your feature request related to a problem?

The host with no associated IPv6 addresses receives a AAAA record question query, the mDNS responds nothing. Accord to the [RFC6762](https://www.rfc-editor.org/rfc/rfc6762#section-11) 6.1 section, the NSEC record is required to add the addtional response. 

### Describe the solution you'd like.

I hope the mDNS's responder can repond the DNS NSEC record to the question querier, when the question's resource record does not exist.

### Additional context.

The feature is preferable to the querier having to retransmit its query multiple times, and eventually give up with a timeout.

PR #271, opened 2023-04-06
mDNS: Add an NSEC record for a host that has no address. (IDFGH-9966)

xieqinan:
### Problem Description
- When a querier is querying for a AAAA record, and the host name in question has no associated IPv6 addresses.  In this case, the responding host knows it currently has exclusive ownership of that name, and it knows that it currently does not have any IPv6 addresses, so an explicit negative response is preferable to the querier having to retransmit its query multiple times.

###  Modification
- Allow to add the host which has no address.
- Add address detection to the host.
- Add NSEC record type. 

### Relate Issuse
[issue](https://github.com/espressif/esp-protocols/issues/258)

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -97,6 +97,16 @@ typedef struct mdns_result_s {
     mdns_ip_addr_t *addr;                   /*!< linked list of IP addresses found */
 } mdns_result_t;
 
+/**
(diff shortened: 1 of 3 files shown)


gabsuren:
@xieqinan I noticed you closed the [issue](https://github.com/espressif/esp-protocols/issues/258) associated with this PR. Should I close this PR as well ?

PR #278, opened 2023-04-19
fix(websocket): esp_websocket_client client allow sending 0 byte packets (IDFGH-9316)

gabsuren:


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1099,7 +1099,7 @@ int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client,
     int ret = ESP_FAIL;
 
     if (client == NULL || len < 0 ||
-            (opcode != WS_TRANSPORT_OPCODES_CLOSE && (data == NULL || len <= 0))) {
+            (opcode != WS_TRANSPORT_OPCODES_CLOSE && len < 0)) {
```
david-cermak, line 1102: Could you please also add a note to the header file, that for sending a zero payload message we have to set x and y to zero?

david-cermak: I wonder if we can remove the entire sub condition? Since `len < 0` is taken care of before, and checking opcode with the same length condition seems redundant to me?

gabsuren:
@david-cermak PTAL

david-cermak (approved):

PR #282, opened 2023-04-26
docs(common): deploy docs if component version get changed

gabsuren:


docs/asio/generate_docs
```diff
@@ -1,26 +1,64 @@
...
+    mySpan.style.float = 'left';
     mySpan.setAttribute('type', 'text');
     mySpan.setAttribute('maxLength', '10');
     mySpan.value = 'all targets';
     mySpan.setAttribute('disabled', true);
     myAnchor.parentNode.replaceChild(mySpan, myAnchor);
 
 })();
-</script>" >> html/index.html
+</script>" | tee -a docs/generic/html/index.html > /dev/null
```
david-cermak, line 64: Question: what's the reason of using `tee` and discarding the output to `/dev/null` later ?

gabsuren: Good point, I initially used it for another case and forget to modify. Thanks!

david-cermak (approved):

PR #295, opened 2023-05-15
mdns: Add APIs for looking up self hosted services and getting the self hostname (IDFGH-10130)

wqx6:
We need the APIs to lookup self hosted services and get the self hostname.

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -129,6 +129,19 @@ void mdns_free(void);
 esp_err_t mdns_hostname_set(const char *hostname);
 
 /**
+ * @brief Get the hostname for mDNS server
+ *
+ * @param hostname      pointer to the hostname, it should be allocated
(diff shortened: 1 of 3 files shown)


components/mdns/include/mdns.h
```diff
@@ -128,6 +128,18 @@ void mdns_free(void);
  */
 esp_err_t mdns_hostname_set(const char *hostname);
 
+/**
+ * @brief Get the hostname for mDNS server
+ *
+ * @param hostname      pointer to the hostname
```
david-cermak, line 134: Could probably say that this `pointer to the hostname` needs to be allocated and hold at least `MDNS_NAME_BUF_LEN` chars.

david-cermak (approved):
LGTM

david-cermak:
@wqx6 Please install the pre-commit hook and fix the commit message.

components/mdns/include/mdns.h
wqx6: Added.

PR #296, opened 2023-05-15
fix(ci): enable code coverage publishing to github pages

gabsuren:


.github/workflows/modem__build-host-tests.yml
```diff
@@ -155,23 +155,22 @@ jobs:
...
+    name: Publish-Results
+    if: github.ref == 'refs/heads/master' || github.repository != 'espressif/esp-protocols'
+    runs-on: ubuntu-22.04
+    needs: host_test_gcov_esp_modem
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Download Artifacts
```
david-cermak, line 168: Do you think it would be possible to deploy the pages directly in the job above? (that way we could remove the upload/download artifact section and simplify the workflow significantly) What was the reason to run it in a separate job?

.github/workflows/modem__build-host-tests.yml
```diff
@@ -124,12 +127,13 @@ jobs:
         run: |
           python -m pip install gcovr
           cd $GITHUB_WORKSPACE/${{ env.COMP_DIR }}
+          git --version
```
david-cermak, line 130: Forgotten?

.github/workflows/modem__build-host-tests.yml
```diff
@@ -124,12 +127,13 @@ jobs:
...
           gcov-8 `find . -name "esp_modem*gcda" -printf '%h\n' | head -n 1`/*
           gcovr --gcov-ignore-parse-errors -g -k -r . --html index.html  -x esp_modem_coverage.xml
           mkdir docs_gcovr
           cp $GITHUB_WORKSPACE/${{ env.COMP_DIR }}/index.html docs_gcovr
-          touch docs_gcovr/.nojekyll
-
+          cp -rf docs_gcovr $GITHUB_WORKSPACE
+          echo "GITHUB_WORKSPACE=${GITHUB_WORKSPACE}" >> $GITHUB_ENV
```
david-cermak, line 136: Do we need this? I thought GITHUB_WORKSPACE is already an automatic variable?

.github/workflows/modem__build-host-tests.yml
```diff
@@ -154,24 +157,9 @@ jobs:
...
-#        uses: actions/download-artifact@v1
-#        with:
-#          name: docs_gcovr
-#
-#      - name: Deploy generated docs
-#        uses: JamesIves/github-pages-deploy-action@v4
-#        with:
-#          branch: gh-pages
-#          folder: 'docs_gcovr'
+      - name: Deploy generated docs
```
david-cermak, line 160: We should probably rename to "deploy code coverage results" or something similar

david-cermak (approved):

PR #302, opened 2023-05-26
mdns: allow for adding a delegated host with no address (IDFGH-10246)

gytxxsy:
Allow mdns to add a delegated host with no address.


david-cermak (approved):
LGTM, but please fix the commit message!

Note, that the commit messages are now used to populate release notes, so I'd suggest rephrasing it to:
```
feat(mdns): Allow for adding a delegated host with no address
```
(or something similar)
```

PR #303, opened 2023-05-31
Reduce heap allocations by getting rid of temporary std::list and std::string (IDFGH-10298)

0xFEEDC0DE64:
Hi,

I was unhappy with unnecessary std::list and std::string constructions, so I fixed the api.

Do we really need linked lists for pass and error arguments or are sequential memory containers also ok? for stack usage you can now use c style arrays and for dynamic length arguments you can use std::vector.

### DIFF
--- a/components/esp_modem/CMakeLists.txt
+++ b/components/esp_modem/CMakeLists.txt
@@ -37,7 +37,7 @@ idf_component_register(SRCS "${srcs}"
 
 
 set_target_properties(${COMPONENT_LIB} PROPERTIES
-    CXX_STANDARD 17
+    CXX_STANDARD 23
     CXX_STANDARD_REQUIRED ON
     CXX_EXTENSIONS ON
 )
--- a/components/esp_modem/include/cxx_include/esp_modem_command_library.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_command_library.hpp
@@ -31,9 +31,9 @@ namespace dce_commands {
  * @param fail_phrase String to be present in the reply to fail this command
  * @param timeout_ms Timeout in ms
  */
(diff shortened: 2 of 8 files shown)


david-cermak:
@0xFEEDC0DE64 Thanks for your contribution!
It makes perfect sense to use `std::string_view` anywhere possible, but we'd like to keep the API compatible with C++11. That's why we use `std::string` (const refs if possible) in public API, but the string_view in private API.
I think we can possibly add a new include folder for users with c++17 and later?

0xFEEDC0DE64:
Could we please increase the minimum required cpp version for esp_modem?

david-cermak:
> Could we please increase the minimum required cpp version for esp_modem?

Any reason you need some C++23 features in `esp_modem` ?

I think we could use C++17 for IDF < `v5.0` or default IDF standard otherwise (which is `gnu++23` in current master).

david-cermak:
We still need to support `esp_modem` with IDF < `v5.0` and we cannot use c++23 with IDF `v4.x`

I agree that using `std::span` and `std:string_view` would be great, but we cannot switch to it at the moment.

@0xFEEDC0DE64 I've put up a draft PR which is trying to address some of the issues here, since we need to
* maintain backward compatibility (still need to use `std::string` based API)
* support IDF `v4.x` (created a simple struct `lw_span` that serves a temp replacement of `std::span`)
* keep c++17 API headers separately, so they could be included only on demand.

PTAL at https://github.com/espressif/esp-protocols/pull/419 ; not very clear solution, but that's probably as far as we can get, I'm open to suggestions.

david-cermak:
As mentioned above, cannot accept the changes at this state. Have created https://github.com/espressif/esp-protocols/pull/419 to followup with your ideas (picked one your commits)

PR #312, opened 2023-06-23
fix(modem): Refactor of DTE callbacks (and other fixes)

david-cermak:
* Simplifies network switching
* Implements inflatable buffer in CMUX and CMD mode
* Sets the ground for URC handling in the library
* Fixes potential data races (adding an extra mutex)

## TODO

- [x] Cleanup, naming
- [x] Make inflatable buffer configurable
- [x] Simplify Term and Netif -- no need for event flags anymore

### DIFF
--- a/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
@@ -79,6 +79,13 @@ public:
      */
     void set_read_cb(std::function<bool(uint8_t *data, size_t len)> f);
 
+    /**
+     * @brief Sets read callback for manual command processing
+     * Note that this API also locks the command API, which can only be used
+     * after you remove the callback by dte->on_read(nullptr)
+     *
+     * @param on_data Function to be called when a command response is available
+     */
     void on_read(got_line_cb on_data) override;
 
     /**
@@ -122,7 +129,6 @@ protected:
     }
     friend class Scoped<DTE>;                               /*!< Declaring "Scoped<DTE> lock(dte)" locks this instance */
 private:
-    static const size_t GOT_LINE = SignalGroup::bit0;       /*!< Bit indicating response available */
 
     [[nodiscard]] bool setup_cmux();                        /*!< Internal setup of CMUX mode */
     [[nodiscard]] bool exit_cmux();                         /*!< Exit of CMUX mode and cleanup  */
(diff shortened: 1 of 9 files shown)


components/esp_modem/src/esp_modem_dce.cpp
```diff
@@ -21,10 +21,17 @@ static bool exit_data(DTE &dte, ModuleIf &device, Netif &netif)
     netif.stop();
     auto signal = std::make_shared<SignalGroup>();
     std::weak_ptr<SignalGroup> weak_signal = signal;
-    dte.set_read_cb([weak_signal](uint8_t *data, size_t len) -> bool {
+    dte.set_read_cb([&netif, weak_signal](uint8_t *data, size_t len) -> bool {
```
euripedesrocha, line 24: I think we already had this discussion, but, is it safe to capture `netif` by reference here?

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -17,53 +17,118 @@ static const size_t dte_default_buffer_size = 1000;
...
 
-command_result DTE::command(const std::string &command, got_line_cb got_line, uint32_t time_ms, const char separator)
+void DTE::set_command_callbacks()
 {
-    Scoped<Lock> l(internal_lock);
-    result = command_result::TIMEOUT;
-    signal.clear(GOT_LINE);
-    primary_term->set_read_cb([this, got_line, separator](uint8_t *data, size_t len) {
-        if (!data) {
+    primary_term->set_read_cb([this](uint8_t *data, size_t len) {
```
euripedesrocha, line 51: In the current state of the code, I *think* we are safe capturing `this` and setting it to the callback, but I would consider start using a `shared_ptr` with `shared_from_this`.

components/esp_modem/src/esp_modem_uart.cpp
```diff
@@ -91,7 +90,6 @@ class UartTerminal : public Terminal {
     static const size_t TASK_INIT = BIT0;
     static const size_t TASK_START = BIT1;
     static const size_t TASK_STOP = BIT2;
-    static const size_t TASK_PARAMS = BIT3;
```
euripedesrocha, line 94: Could you elaborate on why it's no longer needed?

components/esp_modem/include/cxx_include/esp_modem_dte.hpp
```diff
@@ -133,9 +139,68 @@ class DTE : public CommandableIf {
...
+     * @brief Implements an extra buffer that is used to capture partial reads from underlying terminals
+     * when we run out of the standard buffer
+     */
+    struct extra_buffer {
+        extra_buffer(): buffer(nullptr) {}
+        ~extra_buffer()
+        {
+            delete buffer;
+        }
+        std::vector<uint8_t> *buffer;
```
euripedesrocha, line 155: Why use the `vector<>*` instead of `vector<>`?

components/esp_modem/src/esp_modem_dce.cpp
david-cermak: Yes, it is safe (in the current design), as DCE has exclusive ownership of the netif class.
Passing this to the callback should be safe too, as the callback is valid in the current scope and we get the reference from the function params.

components/esp_modem/src/esp_modem_uart.cpp
david-cermak: Yes, I've introduced one new lock to DTE. Starting from 44bae24c we shifted the responsibility of thread safety from Terminal to UART. This fix actually finishes and secures the idea.

With this change we could actually get rid of the second lock (API-lock), employing some status variables, but I would leave this to a future PR...

components/esp_modem/include/cxx_include/esp_modem_dte.hpp
david-cermak: The reason is that I wanted to deallocate the members, which is not under my control if I used just `vector<>`.

The idea is that we're able to grow the extra buffer in exceptional cases and after we do that and process all the data, we should free these blocks.

david-cermak:
@euripedesrocha @suren-gabrielyan-espressif PTAL

euripedesrocha (approved):

david-cermak:
Thanks for the review @euripedesrocha !

tore-espressif:
@david-cermak This change is breaking DUAL_MODE for USB modems. Is there any reason why the DTE constructor cannot initialize DUAL_MODE anyomore?  https://github.com/espressif/esp-protocols/pull/312/commits/cb6e03ac628eb937befc9d31883c3e73a6a35de6#diff-d13893478ad7bd229158ba84355f1d361f2afd680a28605a304de8945280ce0d

Related issue https://github.com/espressif/idf-extra-components/issues/249

david-cermak:
> @david-cermak This change is breaking DUAL_MODE for USB modems. Is there any reason why the DTE constructor cannot initialize DUAL_MODE anyomore? [cb6e03a#diff-d13893478ad7bd229158ba84355f1d361f2afd680a28605a304de8945280ce0d](https://github.com/espressif/esp-protocols/commit/cb6e03ac628eb937befc9d31883c3e73a6a35de6#diff-d13893478ad7bd229158ba84355f1d361f2afd680a28605a304de8945280ce0d)
> 
> Related issue [espressif/idf-extra-components#249](https://github.com/espressif/idf-extra-components/issues/249)

Oh, I missed that. This really is just a copy-paste error. I'll fix it.

It seems we really need at least one CI runner with a USB device.

PR #314, opened 2023-06-26
Update issue template with correct docs links

david-cermak:
Just updating docs links. Decided to give a common link to the README, rather that listing all the components and docs.


tom-borcin (approved):
Looks good to me!

PR #324, opened 2023-07-14
feat(mdns): Allow setting length of mDNS action queue in menuconfig

gytxxsy:
Allow users to configure the length of mDNS action queue via menuconfig.


david-cermak (approved):
Thanks for the contribution!

PR #334, opened 2023-08-03
fix(mdns): added guard check for null pointer (IDF-7449)

gabsuren:


david-cermak (changes requested):
Could you please give some context? like a description on this change. what issue does it fix?

gabsuren:
> Could you please give some context? like a description on this change. what issue does it fix?

@david-cermak thank you for your note. Updated the PR description

PR #336, opened 2023-08-08
 fix(mdns): fix build issue if CONFIG_ESP_WIFI_ENABLED disabled (IDFGH-9699)

gabsuren:


components/mdns/mdns.c
```diff
@@ -4180,9 +4181,10 @@ void mdns_preset_if_handle_system_event(void *arg, esp_event_base_t event_base,
         default:
             break;
         }
-    }
+    } else
```
david-cermak, line 4184: I think this won't compile when `CONFIG_ESP_WIFI_ENABLED==1 && CONFIG_ETH_ENABLED==0`

david-cermak (approved):

Issue #339, opened 2023-08-09
Null pointer exception (LoadProhibited) on incoming UART data after failed CMUX initialization  (IDFGH-10845)


victorrar:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### Issue:
After unsuccessful CMUX initialization, incoming UART data will cause panic.

<details><summary>Details</summary>
<p>

E (00:02:45.702) uart: uart_read_bytes(1283): uart data null
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x4000c250  PS      : 0x00060c30  A0      : 0x800fcb83  A1      : 0x3ffd0210  
A2      : 0x00000000  A3      : 0x0000000a  A4      : 0xffffffff  A5      : 0xffffffff  
A6      : 0x3f414474  A7      : 0x00000000  A8      : 0x800fb528  A9      : 0x3ffd01a0  
A10     : 0xffffffff  A11     : 0x00000000  A12     : 0x00000001  A13     : 0xffffffff  
A14     : 0x3ffd01e0  A15     : 0x0000000c  SAR     : 0x00000004  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000000  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xfffffffb  

Backtrace: 0x4000c24d:0x3ffd0210 0x400fcb80:0x3ffd0230 0x400fb6e5:0x3ffd0260 0x400fb7f1:0x3ffd02a0

</p>
</details> 

I'm testing it on SIM7080G modem over UART, here is an example snippet.

```c++
auto dte = esp_modem::create_uart_dte(&storedConfig.dte_config);
auto dce = esp_modem::create_SIM7070_dce(&storedConfig.dce_config, dte, netifNB);
...
dce->set_mode(esp_modem::modem_mode::CMUX_MODE);

//<now I send some data or touch a floating RX pin with a finger>
```

This is caused in esp_modem_dte.cpp, because when CMUX is created, the terminal and buffer are moved to it and no longer valid.
But if `cmux_term->init()` fails, it does not restore the validity of the terminal/buffer.

[...]

PR #343, opened 2023-08-18
Fix LoadProhibited after failed CMUX initialization

victorrar:
Fixes https://github.com/espressif/esp-protocols/issues/339 (After unsuccessful CMUX initialization, incoming UART data will cause panic.)

Also, I was forced to bump the check-copyright version because I could not run precomit_hooks. The core issue was https://github.com/espressif/check-copyright/pull/10


david-cermak (approved):
Changes LGTM and the tests pass; I'd still like to spend some time testing this manually.

Thanks for the fix, will merge soon.

PR #357, opened 2023-09-18
Add override_path in example manifests

igrr:
This PR attempts to fix the following issues:
- Currently some of the examples in esp-protocols cannot be used from Component Registry using `idf.py create-project-from-example` command. This is because they are adding the containing component only by adding it to EXTRA_COMPONENT_DIRS, which doesn't work when the example alone is created by the component manager.
  - This is fixed by removing the component from EXTRA_COMPONENT_DIRS and adding it via idf_component.yml, instead.
- Some of the examples do add a dependency on the component via idf_component.yml file, but still use EXTRA_COMPONENT_DIR to pick up the containing component when working with a Git copy of the repository. This is replaced by `override_path` in idf_component.yml.
- Many examples add `protocol_examples_common` component using EXTRA_COMPONENT_DIRS. This makes the examples not work when they are created using `idf.py create-project-from-example`. After an offline discussion with @david-cermak, we have agreed to remove protocol-examples-common from esp-protocols and for the time being reference the copy we have in IDF. Later we will add protocol-examples-common (or a newer and fancier version of it) back to esp-protocols and publish it to the component registry. Then the examples can be made dependent on it.
- `mdns` component example was placed one directory level lower than it should be. Have moved it into its own directory.

Some TODOs after this PR:

- Examples from the Component Registry will still not work until we make bugfix releases of all the components with the above changes.
- Linux-specific examples still make use of various stub components via EXTRA_COMPONENT_DIRS. This has to be cleaned up, as well.


components/asio/examples/asio_chat/main/idf_component.yml
```diff
@@ -1,4 +1,6 @@
 dependencies:
   ## Required IDF version
   idf: ">=5.0"
-  espressif/asio: ">=1.0.1"
+  espressif/asio:
+    version: "^1.14.1"
```
igrr, line 5: I have changed >= to ^ just in case there is a new major version with an incompatible API in the future.

>= would still pick up that new version, and projects created from such examples will start failing.

david-cermak: makes sense, thanks for changing this!

david-cermak (approved):
Thanks for updating this and all related minor fixes!

david-cermak:
This PR is approved and ready for merging, but didn't pass CI. Closing as the changes have been taken to https://github.com/espressif/esp-protocols/pull/420 to complete the process.

PR #360, opened 2023-09-20
fix(modem): TLS example: Add restore session support in mbedtls wrapper

david-cermak:
Reusable component in modem_tcp_client example implements a simple mbedtls wrapper. This update add support for mbedtls deinit() and for saving and resotoring TLS session.

### DIFF
--- a/components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
+++ b/components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <utility>
+#include <memory>
 #include <span>
 #include "mbedtls/ssl.h"
 #include "mbedtls/entropy.h"
@@ -22,6 +23,7 @@ public:
     Tls();
     virtual ~Tls();
     bool init(is_server server, do_verify verify);
+    bool deinit();
     int handshake();
     int write(const unsigned char *buf, size_t len);
     int read(unsigned char *buf, size_t len);
@@ -41,6 +43,11 @@ protected:
     mbedtls_entropy_context entropy_{};
     virtual void delay() {}
 
+    bool set_session();
+    bool get_session();
+    void reset_session();
+    bool is_session_loaded();
+
 private:
     static void print_error(const char *function, int error_code);
     static int bio_write(void *ctx, const unsigned char *buf, size_t len);
@@ -48,5 +55,21 @@ private:
     int mbedtls_pk_parse_key( mbedtls_pk_context *ctx,
                               const unsigned char *key, size_t keylen,
                               const unsigned char *pwd, size_t pwdlen);
+    struct unique_session {
+        unique_session()
+        {
+            ::mbedtls_ssl_session_init(&s);
+        }
+        ~unique_session()
+        {
+            ::mbedtls_ssl_session_free(&s);
+        }
+        mbedtls_ssl_session *ptr()
+        {
+            return &s;
+        }
+        mbedtls_ssl_session s;
+    };
+    std::unique_ptr<unique_session> session_;
 
 };
--- a/components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/mbedtls_wrap.cpp
+++ b/components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/mbedtls_wrap.cpp
@@ -35,6 +35,16 @@ bool Tls::init(is_server server, do_verify verify)
     return true;
 }
 
+bool Tls::deinit()
+{
+    ::mbedtls_ssl_config_free(&conf_);
+    ::mbedtls_ssl_free(&ssl_);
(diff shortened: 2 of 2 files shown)


components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
```diff
@@ -18,10 +19,14 @@ class Tls {
...
     enum class is_server : bool {};
     enum class do_verify : bool {};
+    enum class session_op {
+        GET, SET, FREE, CHECK
+    };
 
     Tls();
     virtual ~Tls();
     bool init(is_server server, do_verify verify);
+    bool deinit();
```
euripedesrocha, line 29: Function has no error path.
```suggestion
    void deinit();
```

components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
```diff
@@ -39,7 +44,9 @@ class Tls {
     mbedtls_ssl_config conf_{};
     mbedtls_ctr_drbg_context ctr_drbg_{};
     mbedtls_entropy_context entropy_{};
+    std::unique_ptr<mbedtls_ssl_session> session_{nullptr};
     virtual void delay() {}
+    bool session(session_op op);
```
euripedesrocha, line 49: Have you considered adding different functions instead of the usage of the enum? 
```suggestion
    bool set_session();
    bool get_session();
    void reset_session();
    bool session_loaded();
```

Also, from the documentation I think store and load are better names than get and set, but I understand the usage of the latter from mbedtls names.

david-cermak: Good point, thanks for suggesting. This looks more "programatically" correct. Just changed `session_loaded()` to `is_session_loaded()`

components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
david-cermak: It currently doesn't, but I'd like to keep this option open. You probably remember our discussion about `init()` function of this wrapper, I wanted to keep it separate from the constructor, so we could use the object or have better control. The same applies for this `deinit()`.

euripedesrocha: Ok, but cleaning up code usually doesn't have errors to return.

components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
```diff
@@ -41,12 +43,33 @@ class Tls {
...
+        {
+            ::mbedtls_ssl_session_free(&s);
+        }
+        mbedtls_ssl_session *ptr()
+        {
+            return &s;
+        }
+        mbedtls_ssl_session s;
+    };
+    std::unique_ptr<unique_session> session_;
```
euripedesrocha, line 73: Reading the code in the current state it seems that we only have the `std::unique_ptr` because the session isn't necessarily loaded, it seems that we could use a `std::optional` instead.

euripedesrocha (approved):
LGTM.

components/esp_modem/examples/modem_tcp_client/components/extra_tcp_transports/include/mbedtls_wrap.hpp
david-cermak: Yes, the `session_` could be used but doesn't necessarily have to be used, so for the entire lifetime could remain `nullptr`.
And yes, I use `unique_ptr` only destroy the potentially allocated object, if it's used.

I think we could also use `std::optional`, but it seems to me that it would be easier to use a raw pointer and add a `delete session_` to the Tls destructor?

euripedesrocha: With your current design, with the `unique_session` type, the std::optional adds the same effect since the destructor would be called on TLS destructor. The raw pointer cleanup would work also, but I prefer to keep a type dedicated to lifetime control in such cases, just a matter of taste though. The solution with `unique_ptr` is ok, just that optional reflects better the meaning.

david-cermak: Have reworked to use `std::optional`, could you please take a quick look?

david-cermak: Reverted to the `unique_session` per discussion with @euripedesrocha since this is being added to the base class and using session tickets doesn't seem to be a very common usecase for everyone.

PR #361, opened 2023-09-20
fix(modem): Update vfs terminal; Removed designated init

david-cermak:
## Fix netif to set PPP config in C++ way
    
    If the config struct contains more items (or more items will be added in
    future) this init will not compile cleanly

## Fix netif to set PPP config in C++ way
    
    If the config struct contains more items (or more items will be added in
    future) this init will not compile cleanly


euripedesrocha (approved):

david-cermak:
Thanks for the review Euripedes, I'd have to add one commit by @suren-gabrielyan-espressif to pass the CI.

PR #364, opened 2023-09-21
Fix MDNS free host list (IDFGH-11116)

tannewt:
Without this, any use of `_mdns_host_list` risks crashing due to use of freed memory.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -2884,6 +2884,7 @@ static void free_delegated_hostnames(void)
         host = host->next;
         free(item);
     }
+    _mdns_host_list = NULL;
 }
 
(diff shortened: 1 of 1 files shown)


gabsuren:
@tannewt Thank you for your contribution. The modification seems good to me.  
Please update the commit message to align it with our formatting rules, so it can pass the CI.  As commit message must be scoped with `fix(mdns):`
It can be for example:
 `fix(mdns): free host list on destroy `

gabsuren (changes requested):
Thank you for your contribution. The modification seems good to me.
Please update the commit message to align it with our formatting rules, so it can pass the CI. As commit message must be scoped with fix(mdns):
It can be for example:
fix(mdns): free host list on destroy

gabsuren (approved):

PR #368, opened 2023-09-26
fead(websocket): Added new API s to allow fragmented text to be send (IDFGH-10198)

gabsuren:
Added new APIs `esp_websocket_client_send_text_partial`,
                    `esp_websocket_client_send_bin_partial`
                    `esp_websocket_client_send_with_exact_opcode`


components/esp_websocket_client/examples/target/main/websocket_example.c
```diff
@@ -161,6 +161,16 @@ static void websocket_app_start(void)
         vTaskDelay(1000 / portTICK_PERIOD_MS);
     }
 
+    ESP_LOGI(TAG, "Sending fragmented message");
+    memset(data, 'a', 32);
```
david-cermak, line 165: Nitpick
```suggestion
    memset(data, 'a', sizeof(data));
```

david-cermak (approved):
LGTM, nice work!

david-cermak:
@gabsuren please fix the commit message, so it's readable (has the title and the body). Note that this commit message would be used to compose the release notes!

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1126,41 +1178,13 @@ int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client,
...
-                esp_websocket_client_error(client, "esp_transport_write() returned %d, errno=%d", ret, errno);
-            }
-            esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
-            goto unlock_and_return;
-        }
-        current_opcode = 0;
-        widx += wlen;
-        need_write = len - widx;
-
+    if (esp_websocket_client_send_with_exact_opcode(client, opcode | WS_TRANSPORT_OPCODES_FIN, data, len, timeout) != true) {
```
david-cermak, line 1181: So this is the original problem causing https://github.com/espressif/esp-protocols/issues/460

One nitpick here (sorry, missed that) is that this API shouldn't be called `with_exact_opcode()` if we're changing the opcode based on the fragmentation.
So after the prefix, we should at least mention it in the comments.
I'd suggest though adding one extra boolean parameter `append_fin` to make it cleaner.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1092,17 +1128,33 @@ int esp_websocket_client_send_text(esp_websocket_client_handle_t client, const c
     return esp_websocket_client_send_with_opcode(client, WS_TRANSPORT_OPCODES_TEXT, (const uint8_t *)data, len, timeout);
 }
 
+int esp_websocket_client_send_text_partial(esp_websocket_client_handle_t client, const char *data, int len, TickType_t timeout)
+{
+    return esp_websocket_client_send_with_exact_opcode(client, WS_TRANSPORT_OPCODES_TEXT, (const uint8_t *)data, len, timeout);
```
david-cermak, line 1133: Now, this is another potential problem, we're sending something with the "exact" opcode, but the last frame should contain `FIN` (which is apparently missing)

and I think it's actually missing even after https://github.com/espressif/esp-protocols/pull/461 merged

david-cermak: sorry, the problem is actually here https://github.com/espressif/esp-protocols/pull/368/files#diff-53ead0f3896f79090b1ed1e79e1cca84ac4e505faa1bb76727ab856dabebc297L1092

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1092,17 +1128,33 @@ int esp_websocket_client_send_text(esp_websocket_client_handle_t client, const c
...
+}
+
+int esp_websocket_client_send_cont_msg(esp_websocket_client_handle_t client, const char *data, int len, TickType_t timeout)
+{
+    return esp_websocket_client_send_with_exact_opcode(client, WS_TRANSPORT_OPCODES_CONT, (const uint8_t *)data, len, timeout);
+}
+
 int esp_websocket_client_send_bin(esp_websocket_client_handle_t client, const char *data, int len, TickType_t timeout)
 {
     return esp_websocket_client_send_with_opcode(client, WS_TRANSPORT_OPCODES_BINARY, (const uint8_t *)data, len, timeout);
```
david-cermak, line 1143: Same issue as here https://github.com/espressif/esp-protocols/pull/368/files#r1459119170

components/esp_websocket_client/esp_websocket_client.c
gabsuren: In order to mark the last frame I added `esp_websocket_client_send_fin` API. 
And documented for `esp_websocket_client_send_text_partial` 
`- To mark the end of fragmented data, you should use the 'esp_websocket_client_send_fin(...)' API. This sends a FIN frame.`
https://github.com/espressif/esp-protocols/blob/f62db5cfa2e64a27828a90c42ce92c2d2cb4ae89/components/esp_websocket_client/include/esp_websocket_client.h#L284C1-L289C128

david-cermak: Plus we're not locking the client and perform the usually API parameters check?

PR #374, opened 2023-10-11
fix(console): Fixed publish job failure by adding .cz.yaml

espressif-abhikroy:
Added .cz.yaml file to fix publish job failure.

### DIFF
--- a/components/console_simple_init/.cz.yaml
+++ b/components/console_simple_init/.cz.yaml
@@ -1,0 +1,8 @@
+---
+commitizen:
+  bump_message: 'bump(console_simple_init): $current_version -> $new_version'
+  pre_bump_hooks: python ../../ci/changelog.py console_simple_init
+  tag_format: console_simple_init-v$version
+  version: 1.0.0
+  version_files:
+  - idf_component.yml
--- a/components/console_simple_init/CHANGELOG.md
+++ b/components/console_simple_init/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+
+## [1.0.0](https://github.com/espressif/esp-protocols/commits/console_simple_init-v1.0.0)
+
+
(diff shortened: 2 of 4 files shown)


gabsuren (approved):

components/console_simple_init/idf_component.yml
```diff
@@ -1,6 +1,6 @@
-version: 1.0.0
-url: https://github.com/espressif/esp-protocols/tree/master/components/console_simple_init
+version: 0.0.1
```
david-cermak, line 1: Should we still use the `1.0.0` ?

components/console_simple_init/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## [0.0.1]()
```
david-cermak, line 3: Was this file generated by the script? I don't see the `bump` commit

espressif-abhikroy: I manually created it.

david-cermak: also, this should point to the commit id: https://github.com/espressif/esp-protocols/commits/123456)

components/console_simple_init/.cz.yaml
```diff
@@ -0,0 +1,8 @@
+---
+commitizen:
+  bump_message: 'bump(console_simple_init): $current_version -> $new_version'
+  pre_bump_hooks: python ../../ci/changelog.py console_simple_init
+  tag_format: console_simple_init-v$version
+  version: 1.0.0
```
david-cermak, line 6: would it trigger the master jobs now? I know that it's a bit tricky, since version `1.0.0` is already on master, but it didn't succeed so we probably need to update it?

PR #376, opened 2023-10-12
fix(modem): Fix DUAL_MODE regression from cb6e03ac

david-cermak:
In the recent DTE refactoring (cb6e03ac) we install terminal callbacks in the constructor, but the change missed initializing modem_state in DTE constructors which take two terminals as arguments to work in DUAL mode.


tore-espressif (approved):
LGTM! Thank you for the quick update

PR #378, opened 2023-10-12
Fix console component deployment

gabsuren:


components/console_simple_init/.cz.yaml
```diff
@@ -0,0 +1,8 @@
+---
+commitizen:
+  bump_message: 'bump(console): $current_version -> $new_version'
+  pre_bump_hooks: python ../../ci/changelog.py console_simple_init
+  tag_format: console-v$version
```
david-cermak, line 5:
```suggestion
  tag_format: console_simple_init-v$version
```
cc @espressif-abhikroy

PR #386, opened 2023-10-18
Enable ESP_WIFI_CONFIG when ESP-IDF <= 5.1 (IDFGH-11221)

gabsuren:


components/mdns/mdns.c
```diff
@@ -25,6 +25,9 @@
 #include "esp_wifi.h"
 #endif
 
+#if ESP_IDF_VERSION <= ESP_IDF_VERSION_VAL(5, 1, 0)
+#define CONFIG_ESP_WIFI_ENABLED
```
david-cermak, line 29: I guess that's why the CI keeps failing
```suggestion
#define CONFIG_ESP_WIFI_ENABLED        1
```

components/mdns/mdns.c
```diff
@@ -25,6 +25,11 @@
 #include "esp_wifi.h"
 #endif
 
+#if ESP_IDF_VERSION <= ESP_IDF_VERSION_VAL(5, 1, 0)
+#ifndef CONFIG_ESP_WIFI_ENABLED
+#define CONFIG_ESP_WIFI_ENABLED 1
+#endif
+#endif
```
david-cermak, line 33: Suggest that we create a local config called something like `MDNS_ESP_WIFI_ENABLED` and do something like:
```suggestion
#if ESP_IDF_VERSION <= ESP_IDF_VERSION_VAL(5, 1, 0)
  // IDF <= v5.1 does not support enabling/disabling esp-wifi
  #define MDNS_ESP_WIFI_ENABLED 1
#else
  #define MDNS_ESP_WIFI_ENABLED CONFIG_ESP_WIFI_ENABLED
#endif
```

david-cermak: I don't think it's wise to update skdconfig option, even if it's not supposed to be defined

david-cermak (approved):

PR #387, opened 2023-10-19
Components/console cmd ifconfig 2

espressif-abhikroy:
Components/console cmd ifconfig 2


components/console_cmd_ifconfig/idf_component.yml
```diff
@@ -0,0 +1,10 @@
+version: 0.0.9
+url: https://github.com/espressif/esp-protocols/tree/master/components/console_cmd_ifconfig
+description: The component offers a console that enables runtime network interface configuration and monitoring.
+dependencies:
+  idf:
+    version: '>=5.0'
+  espressif/console_simple_init:
+    version: '>=1.0.2'
```
igrr, line 8: over here you need `override_path: ../console_simple_init`, otherwise a version from the registry will be used during development.

espressif-abhikroy: This branch is not in use. Sorry, I forgot to delete it.
This is the correct branch https://github.com/espressif-abhikroy/esp-protocols/tree/components/console_cmd_ifconfig

PR #391, opened 2023-10-20
[esp-modem]: Added USB runner with ESP32S2 with A7670

david-cermak:
* Added USB runner with ESP32S2 with A7670
* Fix DTE to post fragments to parsers for USB term


tore-espressif (approved):
Changes LGTM

Issue #392, opened 2023-10-23
esp_websocket_client_send_with_opcode  returns ESP_FAIL even if send is successful


sjames:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

esp_websocket_client_send_with_opcode in https://github.com/espressif/esp-protocols/blob/3e8de3af3a8bda2c1b0b1eb3e20c277fe123c40a/components/esp_websocket_client/esp_websocket_client.c#L1201C5-L1201C42 always returns `ESP_FAIL`

This is a bug in the code.  After all the operations are successful, the execution flows into the `unlock_and_return` label which incorrectly returns ESP_FAIL.

The fix is simple, here it is
```C
int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client, ws_transport_opcodes_t opcode, const uint8_t *data, int len, TickType_t timeout)
{
    int error = ESP_OK;
    if (client == NULL || len < 0 || (data == NULL && len > 0)) {
        ESP_LOGE(TAG, "Invalid arguments");
        return ESP_FAIL;
    }

    if (xSemaphoreTakeRecursive(client->lock, timeout) != pdPASS) {
        ESP_LOGE(TAG, "Could not lock ws-client within %" PRIu32 " timeout", timeout);
        return ESP_FAIL;
    }

    if (!esp_websocket_client_is_connected(client)) {
        ESP_LOGE(TAG, "Websocket client is not connected");
        error = ESP_FAIL;
        goto unlock_and_return;
    }

    if (client->transport == NULL) {
        ESP_LOGE(TAG, "Invalid transport");
        error = ESP_FAIL;
        goto unlock_and_return;
    }
    if (esp_websocket_new_buf(client, true) != ESP_OK) {
        ESP_LOGE(TAG, "Failed to setup tx buffer");
        error = ESP_FAIL;
        goto unlock_and_return;
    }
    if (esp_websocket_client_send_with_exact_opcode(client, opcode | WS_TRANSPORT_OPCODES_FIN, data, len, timeout) != true) {
        ESP_LOGE(TAG, "Failed to send the buffer");
        error = ESP_FAIL;
        goto unlock_and_return;
    }
unlock_and_return:
[...]

PR #393, opened 2023-10-23
fix : Return correct error code for esp_websocket_client_send_with_opcode (IDFGH-11291)

sjames:
This function esp_websocket_client_send_with_opcode was incorrectly sending ESP_FAIL even on success.  Fixing this.

Fixes issue: #392

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1200,7 +1200,7 @@ int esp_websocket_client_send_fin(esp_websocket_client_handle_t client, TickType
 
 int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client, ws_transport_opcodes_t opcode, const uint8_t *data, int len, TickType_t timeout)
 {
-    int error = ESP_OK;
+    int ret = ESP_OK;
     if (client == NULL || len < 0 || (data == NULL && len > 0)) {
         ESP_LOGE(TAG, "Invalid arguments");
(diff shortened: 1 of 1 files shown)


david-cermak (approved):
Thanks for the fix!

gabsuren:
@sjames thank you for the contribution. 
Nitpick, I would like to suggest rename of `int error = ESP_OK` to `int ret = ESP_OK`, as in a good case it is ESP_OK which is not error.
And also need to modify commit message to `fix(websocket): Return error code correctly` so it can pass pre-commit hook check

Thanks again

sjames:
> @sjames thank you for the update! The changes look good to me. But as you see now there are two commits now Screen Shot 2023-10-24 at 11 27 42 AM
> 
> The CI will still fail if we merge it like that. You can merge those to commit to one, by
> 
>     1. `git rebase -i HEAD~2`
> 
>     2. replace pick with squash for the second commit
> 
>     3. Edit the Commit Message:
> 
>     4. Finalize the Rebase and push changes by --force  (`git push --f`)
> 
> 
> If you face any issues, I can cherry-pick your changes(keeping the history/metadata) and push it in separate PR. Please let me know in case any help needed.
> 
> Thanks Suren

I think I followed the instructions and pushed the changes. If this still doesn't work, I would really appreciate your cherry-picking offer!

Thanks,
Sojan

gabsuren:
Created https://github.com/espressif/esp-protocols/pull/401 based on this PR

PR #397, opened 2023-10-23
ci(common): Created reusable actions for host and coverage tests

gabsuren:


.github/workflows/run-host-tests.yml
```diff
@@ -56,7 +73,12 @@ jobs:
             apt-get update && apt-get install -y python3-pip rsync
             python -m pip install gcovr
             cd $GITHUB_WORKSPACE/${{inputs.component_path}}
-            gcov `find . -name "*gcda"`
+            component_name=$(basename ${{ inputs.component_path }})
+            if [[ ${{ inputs.idf_version }} == "release-v4.3" ]]; then
+                gcov-8 `find . -name "$component_name*gcda" -printf '%h\n' | head -n 1`/*
+            else
+                gcov `find . -name "$component_name*gcda"`
+            fi
```
david-cermak, line 81: can we create a simple function in bash, e.g. `get_gdca()` and put it into the `env.sh`?

then we can simply call `gcov get_gdca()` with no ifdef here

components/esp_modem/test/host_test/env.sh
```diff
@@ -1,10 +1,19 @@
...
+
+if [ "$idf_version" == "release-v4.3" ]; then
+  lwip=lwip-2.1.2
+  lwip_uri=http://download.savannah.nongnu.org/releases/lwip
+  lwip_contrib=contrib-2.1.0
+
+  wget --no-verbose ${lwip_uri}/${lwip}.zip
+  unzip -oq ${lwip}.zip
+  wget --no-verbose ${lwip_uri}/${lwip_contrib}.zip
+  unzip -oq ${lwip_contrib}.zip
```
david-cermak, line 13: This will probably need to be added only if we need `lwip`, so for modem tests.
Suggest that we take two arguments, not only the IDF version, but also a component and decide based on it

david-cermak (approved):
LGTM with few nitpicks

components/esp_modem/test/host_test/env.sh
gabsuren: Ok, so you suggest to create common `env.sh` for all the tests, right?

.github/workflows/run-host-tests.yml
david-cermak: Okay, so could we find a better way of choosing which method we use for finding the files, other than based on `idf_version`? Maybe `gcov --version`?
Sorry, I just don't find this condition descriptive enough, maybe after adding some comments it might be...?

components/esp_modem/test/host_test/env.sh
```diff
@@ -1,10 +1,20 @@
...
-lwip_contrib=$3
-
-wget --no-verbose ${lwip_uri}/${lwip}.zip
-unzip -oq ${lwip}.zip
-wget --no-verbose ${lwip_uri}/${lwip_contrib}.zip
-unzip -oq ${lwip_contrib}.zip
+idf_version=$1
+component=$2
+
+if [ "$idf_version" == "release-v4.3" ] && [ "$component" == "esp_modem" ]; then
```
david-cermak, line 6: probably need something like... ?
```suggestion
if [[ "$idf_version" == "release-v4.3" ]] && [[ "$component" == "esp_modem" ]]; then
```

.github/workflows/run-host-tests.yml
```diff
@@ -56,7 +74,12 @@ jobs:
             apt-get update && apt-get install -y python3-pip rsync
             python -m pip install gcovr
             cd $GITHUB_WORKSPACE/${{inputs.component_path}}
-            gcov `find . -name "*gcda"`
+            component=$(basename ${{ inputs.component_path }})
+            if [[ ${{ inputs.idf_version }} == "release-v4.3" ]]; then
+                gcov-8 `find . -name "$component*gcda" -printf '%h\n' | head -n 1`/*
+            else
+                gcov `find . -name "$component*gcda"`
+            fi
```
david-cermak, line 82: TBH, I don't very much like this condition. could be decide based on `gcov` version, rather than IDF version?

components/esp_modem/test/host_test/env.sh
```diff
@@ -1,10 +1,20 @@
...
+  lwip_uri=http://download.savannah.nongnu.org/releases/lwip
+  lwip_contrib=contrib-2.1.0
+
+  wget --no-verbose ${lwip_uri}/${lwip}.zip
+  unzip -oq ${lwip}.zip
+  wget --no-verbose ${lwip_uri}/${lwip_contrib}.zip
+  unzip -oq ${lwip_contrib}.zip
+
+  apt-get update && apt-get install -y gcc-8 g++-8
+  update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8
```
david-cermak, line 17: What is the reason for installing `gcc-8` here? Why we can't use the default one from the image?

.github/workflows/run-host-tests.yml
gabsuren: Thank you for the suggestion. The problem is, for the `idf-version-4.3` gcov will not work, as it is linked to gcov-9, which failed, in order to build for idf-version-4.3 `gcov-8` is needed

components/esp_modem/test/host_test/env.sh
gabsuren: https://github.com/espressif/esp-protocols/pull/397#discussion_r1380291725

.github/workflows/modem__build-host-tests.yml
```diff
@@ -54,105 +54,12 @@ jobs:
...
-        uses: actions/upload-artifact@v3
-        if: always()
-        with:
-          name: modem_coverage_report
-          path: |
-            ${{ env.COMP_DIR }}/modem_coverage_report
-          if-no-files-found: error
+    uses: "./.github/workflows/run-host-tests.yml"
+    with:
+        idf_version: "release-v4.3"
```
euripedesrocha, line 59: Shall we introduce host tests with other idf releases?

euripedesrocha (approved):

.github/workflows/modem__build-host-tests.yml
gabsuren: Yes, there is a story regarding it IDF-7828

components/esp_modem/examples/linux_modem/sdkconfig.ci.linux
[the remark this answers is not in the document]
gabsuren: As it is reusable action, it considers that `sdkconfig.ci.linux` and `sdkconfig.ci.coverage` should be available in all host  tests/examples ( as in websocket host example). 
I could add a check if `sdkconfig.defaults` is available skip it. But preferred to leave it like that for now. Other suggestions?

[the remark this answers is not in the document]
david-cermak: Okay, let's leave it as is. but I'd suggest mentioning this in the comments of the reusable action, saying what it does and what it expects.

PR #402, opened 2023-10-25
feat(console): Added runtime component registration support in console_simple_init

espressif-abhikroy:
Added support for runtime component registration in console_simple_init.

### DIFF
--- a/components/console_simple_init/README.md
+++ b/components/console_simple_init/README.md
@@ -32,6 +32,28 @@ It also provides an api to register an user provided command.
     // This allows you to execute the do_user_cmd function when the "user" command is invoked
     ESP_ERROR_CHECK(console_cmd_user_register("user", do_user_cmd));
 
+    // Register any other plugin command added to your project
+    ESP_ERROR_CHECK(console_cmd_all_register());
 
     ESP_ERROR_CHECK(console_cmd_start());    // Start console
     ```
+
+### Automatic registration of console commands
+The `console_simple_init` component includes a utility function named `console_cmd_all_register()`. This function automates the registration of all commands that are linked into the application. To use this functionality, the application can call `console_cmd_all_register()` as demonstrated above.
+
+When creating a new component, you can ensure that its commands are registered automatically by placing the registration function into the `.console_cmd_desc` section within the output binary.
+
+To achieve this, follow these steps:
+1. Add the following lines to the main file of the component
+```
+static const console_cmd_plugin_desc_t __attribute__((section(".console_cmd_desc"), used)) PLUGIN = {
+    .name = "cmd_name_string",
+    .plugin_regd_fn = &cmd_registration_function
+};̌
+```
+2. Add the `WHOLE_ARCHIVE` flag to CMakeLists.txt of the component.
+
+
+For more details refer:
+* [IDF Component Manager](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-component-manager.html)
+* [Linker Script Generation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/linker-script-generation.html)
--- a/components/console_simple_init/console_simple_init.c
+++ b/components/console_simple_init/console_simple_init.c
@@ -8,13 +8,13 @@
 #include "esp_log.h"
 #include "console_simple_init.h"
 
+
 static esp_console_repl_t *repl = NULL;
 static const char *TAG = "console_simple_init";
 
 /**
  * @brief Initializes the esp console
- * @return
- *          - esp_err_t
+ * @return ESP_OK on success
  */
 esp_err_t console_cmd_init(void)
 {
@@ -40,13 +40,12 @@ esp_err_t console_cmd_init(void)
 }
 
 /**
- * @brief Initialize Ethernet driver based on Espressif IoT Development Framework Configuration
+ * @brief Register a user supplied command
  *
  * @param[in] cmd string that is the user defined command
  * @param[in] do_user_cmd Function pointer for a user-defined command callback function
  *
- * @return
- *          - esp_err_t
+ * @return ESP_OK on success
  */
 esp_err_t console_cmd_user_register(char *cmd, esp_console_cmd_func_t do_user_cmd)
 {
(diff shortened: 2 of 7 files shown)


components/console_simple_init/linker.lf
```diff
@@ -0,0 +1,13 @@
...
+    .plugins_desc
+
+[scheme:plugins_desc_default]
+entries:
+    plugins_desc -> flash_rodata
+
+[mapping:plugins_desc]
+archive: *
+entries:
+    * (plugins_desc_default);
```
igrr, line 12: If you use the section name which an IDF example uses, i think you are inviting potential conflicts with users' code which is also copied from the same example :)

Could you change the section names and other identifiers to be specific to your component? (console_simple_init, console_simple_init_desc, console_simple_init_default, etc)

components/console_simple_init/README.md
```diff
@@ -32,6 +32,13 @@ It also provides an api to register an user provided command.
     // This allows you to execute the do_user_cmd function when the "user" command is invoked
     ESP_ERROR_CHECK(console_cmd_user_register("user", do_user_cmd));
 
+    // Register any other plugin command added to your project
+    ESP_ERROR_CHECK(console_cmd_all_register());
 
     ESP_ERROR_CHECK(console_cmd_start());    // Start console
     ```
+
+### Adding a plugin command:
```
igrr, line 41:
```suggestion
### Automatic registration of console commands
```

components/console_simple_init/README.md
```diff
@@ -32,6 +32,13 @@ It also provides an api to register an user provided command.
...
+To add a plugin command or any component from IDF component manager into your project, simply include an entry within the `idf_component.yml` file.
```
igrr, line 42:
```suggestion
`console_simple_init` component provides a function `console_cmd_all_register` which performs automatic registration of all commands linked into the application.

In order for the command to be registered automatically, it has to place the registration function into console_simple_init_desc section.

<Here, either add a code snippet showing how to do this, or link to a piece of code in this repository>
```

components/console_simple_init/console_simple_init.c
```diff
@@ -67,6 +68,34 @@ esp_err_t console_cmd_user_register(char *cmd, esp_console_cmd_func_t do_user_cm
     return ret;
 }
 
+
+/**
+ * @brief Register all the console commands in .plugins_desc section
+ *
+ * @return
+ *          - esp_err_t
```
igrr, line 76: Description of error codes is missing.

components/console_simple_init/console_simple_init.c
```diff
@@ -67,6 +68,34 @@ esp_err_t console_cmd_user_register(char *cmd, esp_console_cmd_func_t do_user_cm
...
+ */
+esp_err_t console_cmd_all_register(void)
+{
+    esp_err_t ret = ESP_FAIL;
+    extern const console_cmd_plugin_desc_t _plugins_array_start;
+    extern const console_cmd_plugin_desc_t _plugins_array_end;
+
+    printf("List of plugin commands:\n");
```
igrr, line 84: It looks like you have copy-pasted IDF example code here. Is logging really necessary in this case? If it is, could we use the standard ESP_LOG mechanism instead of writing directly to stdout?

components/console_simple_init/console_simple_init.c
```diff
@@ -67,6 +68,34 @@ esp_err_t console_cmd_user_register(char *cmd, esp_console_cmd_func_t do_user_cm
...
+    for (const console_cmd_plugin_desc_t *it = &_plugins_array_start; it != &_plugins_array_end; ++it) {
+        printf("- Plugin '%s', function plugin_regd_fn=%p\n", it->name, it->plugin_regd_fn);
```
igrr, line 86: Let's not use the term "plugin" but rather "console command".

components/console_simple_init/console_simple_init.h
```diff
@@ -13,15 +13,27 @@
...
 
+
+/* This stucture describes the plugin to the rest of the application */
+typedef struct {
+    /* A pointer to the plugin name */
+    const char *name;
+
+    /* A function which the plugin provides to the application. */
+    esp_err_t (*plugin_regd_fn)(void);
+} console_cmd_plugin_desc_t;
```
igrr, line 24: The descriptions here seem to be copy-pasted from IDF example. Could you try to make them more specific to this particular use case (auto-registration of console commands)?

For example, what is this "function which the plugin provides to the application" — is it just about any function or some specific function? I think it is the specific function which performs registration of console commands...

igrr (changes requested):

igrr:
Just one note, overall this component seems pretty generic and not specifically tied to esp-protocols. Perhaps we can move it to https://github.com/espressif/idf-extra-components in the future? Implementations of commands specific to protocols should stay here, of course.

components/console_simple_init/README.md
igrr: The `<Here, either add a code snippet showing how to do this, or link to a piece of code in this repository>` was supposed to be a placeholder for you to fill :)

components/console_simple_init/linker.lf
espressif-abhikroy: Changed the section name to .console_cmd_desc.

components/console_simple_init/README.md
espressif-abhikroy: I committed the suggestion just for convenience.
I have made the necessary changes.

components/console_simple_init/console_simple_init.c
espressif-abhikroy: It's reassuring to see the list of all registered plugin components at the starting of the application.
Change the `printf` to `ESP_LOGI`.

components/console_simple_init/README.md
```diff
@@ -32,6 +32,26 @@ It also provides an api to register an user provided command.
...
 
     ESP_ERROR_CHECK(console_cmd_start());    // Start console
     ```
+
+### Automatic registration of console commands
+`console_simple_init` component provides a function `console_cmd_all_register()` which performs automatic registration of all commands linked into the application.
+
+In order for the commands to be registered automatically, it has to place the registration function into `console_cmd_desc` section.
+
+To do that follow these steps:
```
david-cermak, line 46: It should be clear that this part is only for implementing a new component with plug-able commands. This chapter may still give an impression that this is something users need to do to work with commands

euripedesrocha (approved):

david-cermak (approved):

igrr:
Just one remark, LGTM otherwise!

igrr (approved):

PR #414, opened 2023-11-03
fix(mdns): fix compiling issue when disabling IPv4 (IDFGH-11453)

wqx6:
IDF v5.1 provides a menuconfig option to disable IPv4 in LwIP. And mdns component compiling would fail when disabling IPv4


igrr:
It would be nice to also add an `sdkconfig.ci.*` configuration with IPV4 disabled, so that at least the builds for this use case are validated in CI.

components/mdns/mdns.c
```diff
@@ -17,6 +17,7 @@
 #include "mdns_networking.h"
 #include "esp_log.h"
 #include "esp_random.h"
+#include "lwip/opt.h"
```
david-cermak, line 20: Please don't use lwip headers from this file, it's supposed to be network stack independent.

You can use `CONIG_LWIP_IPV4` instead or define a `MNDS_IPV4` based on the config option in `mdns_config.h`

wqx6: `CONFIG_LWIP_IPV4` was introduced in IDF v5.1 but this managed component is for `IDF version>=5.0`, so I didn't use this config option here. May I create a menuconfig option in this component instead of using `CONIG_LWIP_IPV4`?

wqx6:
> It would be nice to also add an `sdkconfig.ci.*` configuration with IPV4 disabled, so that at least the builds for this use case are validated in CI.

Added an configuration file that disables IPv4.

components/mdns/Kconfig
```diff
@@ -77,6 +77,20 @@ menu "mDNS"
...
+        help
+            Enable IPv4 for mDNS
+
+    config MDNS_IPV6
+        bool "Enable IPv6 for mDNS"
+        default y
+        select LWIP_IPV6
+        help
+            Enable IPv6 for mDNS
+
```
david-cermak, line 93: It would be great if we found another solution. If we start doing it like that, then soon most of network protocols start to define their own `XXX_IPV4/6`

components/mdns/mdns.c
```diff
@@ -185,7 +185,7 @@ static inline void _mdns_clean_netif_ptr(mdns_if_t tcpip_if)
 static mdns_if_t _mdns_get_if_from_esp_netif(esp_netif_t *esp_netif)
 {
     for (int i = 0; i < MDNS_MAX_INTERFACES; ++i) {
-        if (esp_netif == s_esp_netifs[i].netif) {
+        if (esp_netif == s_esp_netifs[i].netif || (s_esp_netifs[i].predefined && esp_netif == esp_netif_from_preset_if(s_esp_netifs[i].predef_if))) {
```
david-cermak, line 188: Could you please explain the reason behind this change? I thought the static array already contains the predef netif, correct?

components/mdns/mdns.c
```diff
@@ -3510,22 +3520,28 @@ void mdns_parse_packet(mdns_rx_packet_t *packet)
...
-        if (esp_netif_get_ip6_linklocal(_mdns_get_esp_netif(packet->tcpip_if), &if_ip6) == ESP_OK &&
-                memcmp(&if_ip6, &packet->src.u_addr.ip6, sizeof(esp_ip6_addr_t)) == 0) {
-            return;
-        }
-#endif
     }
+#endif /* CONFIG_MDNS_IPV4 */
+#if defined(CONFIG_MDNS_IPV4) && defined(CONFIG_MDNS_IPV6)
+    else
 #endif
```
david-cermak, line 3534: Nitpick: I would remove this part for readability, since the compiler would address this if we aim for optimizations

components/mdns/mdns_networking_lwip.c
```diff
@@ -21,6 +21,18 @@
...
+#error "Please enable IPv4 in LwIP to use IPv4 in mDNS component"
+#endif
+
+#if defined(CONFIG_MDNS_IPV6) && !LWIP_IPV6
+#error "Please enable IPv6 in LwIP to use IPv6 in mDNS component"
+#endif
+
+#if !defined(CONFIG_MDNS_IPV4) && !defined(CONFIG_MDNS_IPV6)
+#error "Neither IPv4 nor IPv6 is used in mDNS component"
+#endif
```
david-cermak, line 34: We could get rid of these asserts if we find a better way of propagating which IP stack is enabled

components/mdns/mdns_networking_lwip.c
```diff
@@ -152,45 +166,52 @@ static void _udp_recv(void *arg, struct udp_pcb *upcb, struct pbuf *pb, const ip
...
-        else {
-            packet->ip_protocol = MDNS_IP_PROTOCOL_V6;
-            struct ip6_hdr *ip6hdr = (struct ip6_hdr *)(((uint8_t *)(packet->pb->payload)) - UDP_HLEN - IP6_HLEN);
-            memcpy(&packet->dest.u_addr.ip6.addr, (uint8_t *)ip6hdr->dest.addr, 16);
-            packet->multicast = ip6_addr_ismulticast(&(packet->dest.u_addr.ip6));
-        }
+#endif // LWIP_IPV4
+#if LWIP_IPV4 && LWIP_IPV6
+        else
 #endif
```
david-cermak, line 191: Same here, would skip for readability

components/mdns/mdns_networking_socket.c
```diff
@@ -27,6 +27,13 @@
 #include <net/if.h>
 #endif
 
+#if defined(CONFIG_IDF_TARGET_LINUX)
+#define SOCKET_IPV4 1
+#define SOCKET_IPV6 CONFIG_LWIP_IPV6
+#else
+#define SOCKET_IPV4 LWIP_IPV4
+#define SOCKET_IPV6 LWIP_IPV6
+#endif
```
david-cermak, line 36: Again, need to find a better way of IPv4/6 config

david-cermak (commented):
LGTM in general.

The only concern is the LWIP dependency that you're bringing to everything or introduction of our own config options.

I cannot think of a decently elegant way to address this issue, but probably creating definitions of `MDNS_IPV4/6` in `mdns_config.h` based on global config options and IDF release versions.

components/mdns/Kconfig
wqx6: I added these to the mdns_private.h so that all the sources file can use LWIP_IPV4/LWIP_IPV6.
```
#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 0)
/* CONFIG_LWIP_IPV4 was introduced in IDF v5.1 */
#ifndef LWIP_IPV4
#ifdef CONFIG_LWIP_IPV4
#define LWIP_IPV4 1
#else
#define LWIP_IPV4 0
#endif // CONFIG_LWIP_IPV4
#endif // LWIP_IPV4
#else
/* For IDF v5.0, set LWIP_IPV4 to 1 by default */
#ifndef LWIP_IPV4
#define LWIP_IPV4 1
#endif // LWIP_IPV4
#endif // ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 0)

#ifndef LWIP_IPV6
#ifdef CONFIG_LWIP_IPV6
#define LWIP_IPV6 1
#else
#define LWIP_IPV6 0
#endif // CONFIG_LWIP_IPV6
#endif // LWIP_IPV6
```

components/mdns/mdns.c
wqx6: The predefined netifs in the static array are NULL when firstly calling this function if IPv4 is disabled, and the `mld6_joingroup_netif` will not be called rightly then.

wqx6: Added comment here/

components/mdns/mdns.c
wqx6: Removed

components/mdns/mdns_networking_lwip.c
wqx6: Removed

components/mdns/mdns_networking_lwip.c
wqx6: Removed.

components/mdns/mdns_networking_socket.c
wqx6: Removed

wqx6:
The CI fails for socket example, but I can resolve the hostnames with the socket sdkconfig locally. @david-cermak

david-cermak:
> The CI fails for socket example, but I can resolve the hostnames with the socket sdkconfig locally. @david-cermak

@gabsuren Could you please help with the CI failure?

components/mdns/mdns.c
david-cermak: > The predefined netifs in the static array are NULL 

@gabsuren Could you please check?

@wqx6 Do you use the predefined interfaces (default STA, AP or Ethernet) or do you initialize netifs manually?

wqx6: Yes, I used predefined STA interface

gabsuren:
> The CI fails for socket example, but I can resolve the hostnames with the socket sdkconfig locally. @david-cermak

I tried to run it locally with `sdkconfig.ci.eth_socket` and it still fails after a few attempts. Attached you can see the pass and failed logs with `CONFIG_MDNS_ENABLE_DEBUG_PRINTS`. 

[fail.txt](https://github.com/espressif/esp-protocols/files/13523457/fail.txt)

[pass.txt](https://github.com/espressif/esp-protocols/files/13523458/pass.txt)

As you can see in passing  cases the query 
```
TX[0][0]: To: 224.0.0.251:5353, Packet[5240]: 
Q: tinytester.local. A IN 
```
Responses the answer 
```
RX[0][0]: From: 192.168.0.150:5353, To: 0.0.0.0, Created query for esp host: DNS(qd=[Q(name='esp32-mdns-E79568.local')]) 
Created query for esp host: DNS(qd=[Q(name='esp32-mdns-E79568-delegated.local')])
Packet[5280]: AUTHORITATIVE
A: tinytester.local. A IN 0[4] 127.0.0.1
```

However in the failed cases the packet is not there. So there is an issue with `sdkconfig.ci.eth_socket` configuration.
In order to reproduce it please try to run it a few times.

wqx6:
@gabsuren I have also tested locally. In my test, I cannot capture the mdns response for the host-query `tinytester.local` but the response of the `tinytester-lwip.local` is here. But I don't know the reason for it. Could it be the issue of test environment?
image

gabsuren:
@wqx6 I able to see the capture packet in wireshark. 
Screen Shot 2023-12-05 at 5 25 07 PM

And from the log I see that python test sends the answer.

Screen Shot 2023-12-05 at 5 27 24 PM

But some how `mdns_parse_packet` is not capturing it. I can't see the received answer inside `mdns_parse_packet` function. So the issue is in other layers I suppose.

Please check below the both logs

[fail_new.txt](https://github.com/espressif/esp-protocols/files/13560570/fail_new.txt)

[pass_new.txt](https://github.com/espressif/esp-protocols/files/13560573/pass_new.txt)

wqx6:
> Could you please try rerunning it a few times to see if you can reproduce the issue?

I have run the pytest script for over 10 times, but none of them failed.

chshu:
@david-cermak @gabsuren The previous CI issue has been fixed. Could you review and merge it if no other comments? We also want to make a new mDNS release to publish this change, it's required by the Matter Sleep example for C6.

components/mdns/examples/query_advertise/pytest_mdns.py
```diff
@@ -64,6 +64,7 @@ def mdns_server(esp_host, events):
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+    sock.setsockopt(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, 'eth0\0'.encode('utf-8'))
```
gabsuren, line 67: Why we are binding it to `eth0` only?

wqx6: I think the pytest script is using `eth0` to communicate with the test board on CI runner.

wqx6: If I didn't bind the interface, CI will fail https://github.com/espressif/esp-protocols/actions/runs/7550392768/job/20556119271

components/mdns/mdns_networking_socket.c
```diff
@@ -346,17 +352,9 @@ static bool create_pcb(mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol)
...
     int sock = s_interfaces[tcpip_if].sock;
     esp_netif_t *netif = _mdns_get_esp_netif(tcpip_if);
-    if (sock >= 0) {
-        mdns_ip_protocol_t other_ip_proto = ip_protocol == MDNS_IP_PROTOCOL_V4 ? MDNS_IP_PROTOCOL_V6 : MDNS_IP_PROTOCOL_V4;
-        int err = join_mdns_multicast_group(sock, netif, other_ip_proto);
-        if (err < 0) {
-            ESP_LOGE(TAG, "Failed to add ipv6 multicast group for protocol %d", ip_protocol);
-            return false;
-        }
-        s_interfaces[tcpip_if].proto |= (other_ip_proto == MDNS_IP_PROTOCOL_V4 ? PROTO_IPV4 : PROTO_IPV6);
```
david-cermak, line 356: I wonder why this was removed? The idea is that we already have a socket supporting one IP protocol (say IPv4) and someone calls `create_pcb(same_netif, IPv6)`. Then we'll never join the IPv6 multicast group, correct?

wqx6: The `create_pcb(some_netif, IPv6)` will still call `join_mdns_multicast_group(sock, netif, IPv6);` in the following logic.

But if we keep the original logic, When we already have a socket supporting IPv4, and then someone calls `create_pcb(some_netif, IPv4)`, the netif will never join IPv4 multicast group.

david-cermak: Oh, I see. so you actually always join the multicast group (again).
This would (probably?) work as we guard this in `mdns.c` using `mdns_is_netif_ready()` API.

david-cermak (approved):
changes LGTM, provided that the socket networking part has been tested with IPv4 only and dual stack.

components/mdns/examples/query_advertise/pytest_mdns.py
gabsuren: There are two things concerning with this change.
1. The tests may fail if we switch to a different interface or operating system, such as macOS or Windows. I can address this issue later by updating pytest myself.
2. More worryingly, it's unclear why the tests were passing previously without being bound to the specific interface. This raises questions about potential breaking changes that could arise from this discrepancy.

gabsuren: I have executed the changes without binding, in separate [PR503](https://github.com/espressif/esp-protocols/pull/503) and it passed
https://github.com/espressif/esp-protocols/actions/runs/7783718151

So I suppose this change is not necessary . Could you please, revert this and rebase with master so we can merge it?

wqx6: Reverted

gabsuren (approved):

components/mdns/examples/query_advertise/pytest_mdns.py
gabsuren: Thank you

Issue #320, opened 2023-07-03
need a Forced Exit of CMUX Mode


diplfranzhoepfinger:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

need a Forced Exit of CMUX Mode

diplfranzhoepfinger:
so we hacked something together, 
but i fear this is not 100% the right way: 

I would more like to have a Mode like: `CMUX_MANUAL_EXIT_FORCED`,
so we call it only when we know that we want to "force" esp-modem to leave cmux without checking the circumstances. 

Log: https://gist.github.com/diplfranzhoepfinger/c7ab11ed17bde1920a7094c6288f1cc4

Code: https://github.com/diplfranzhoepfinger/esp-protocols-tree-master-components-esp_modem-examples-simple_cmux_client/tree/feature/CMUX_FORCE_Vasil

diplfranzhoepfinger:
we hacked further: 

Log: [https://gist.github.com/diplfranzhoepfinger/a36fd06a8b3ed9398deb52cc9fb774b0](https://gist.github.com/diplfranzhoepfinger/a36fd06a8b3ed9398deb52cc9fb774b0)

Code: [https://github.com/diplfranzhoepfinger/esp-protocols-tree-master-components-esp_modem-examples-simple_cmux_client/tree/feature/CMUX_FORCE_Vasil2](https://github.com/diplfranzhoepfinger/esp-protocols-tree-master-components-esp_modem-examples-simple_cmux_client/tree/feature/CMUX_FORCE_Vasil2)

as expected, the Code is crashing, BUT, at least, we can start Cmux again and again get a IP Address. 

so @david-cermak i think the Problem is solvable, but needs expertise of the Internals.

diplfranzhoepfinger:
@david-cermak any response ?

david-cermak:
Noted, I think it would be really useful to kind of reset to a default state on ESP32.

Could we use this "Unknown state" for that? For example if we know that we got somehow out of sync
1) we switch the modem to the "unknown" (just need to allow this transition)
2) from there, everything should be the same as after init

diplfranzhoepfinger:
cool idea. go forward with it.

[...]

PR #416, opened 2023-11-03
fix(modem): Fixed mode transitions between any state and UNDEF

david-cermak:
Closes https://github.com/espressif/esp-protocols/issues/320

### DIFF
--- a/components/esp_modem/src/esp_modem_dte.cpp
+++ b/components/esp_modem/src/esp_modem_dte.cpp
@@ -151,10 +151,14 @@ command_result DTE::command(const std::string &cmd, got_line_cb got_line, uint32
 
 bool DTE::exit_cmux()
 {
+    if (!cmux_term) {
+        return false;
+    }
     if (!cmux_term->deinit()) {
         return false;
     }
     exit_cmux_internal();
+    cmux_term.reset();
     return true;
 }
 
@@ -174,6 +178,10 @@ void DTE::exit_cmux_internal()
 
 bool DTE::setup_cmux()
 {
+    if (cmux_term) {
+        ESP_LOGE("esp_modem_dte", "Cannot setup_cmux(), cmux_term already exists");
+        return false;
+    }
     cmux_term = std::make_shared<CMux>(primary_term, std::move(buffer));
     if (cmux_term == nullptr) {
         return false;
@@ -198,6 +206,11 @@ bool DTE::setup_cmux()
 
 bool DTE::set_mode(modem_mode m)
 {
+    // transitions (any) -> UNDEF
+    if (m == modem_mode::UNDEF) {
+        mode = m;
+        return true;
+    }
     // transitions (COMMAND|UNDEF) -> CMUX
     if (m == modem_mode::CMUX_MODE) {
         if (mode == modem_mode::UNDEF || mode == modem_mode::COMMAND_MODE) {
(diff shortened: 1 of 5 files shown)


gabsuren (approved):

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -151,10 +151,14 @@ command_result DTE::command(const std::string &cmd, got_line_cb got_line, uint32
...
 bool DTE::exit_cmux()
 {
+    if (!cmux_term) {
+        return false;
+    }
     if (!cmux_term->deinit()) {
         return false;
     }
     exit_cmux_internal();
+    cmux_term = nullptr;
```
euripedesrocha, line 161:
```suggestion
    cmux_term.reset();
```

euripedesrocha: Using reset makes the intention clear.

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -174,6 +178,9 @@ void DTE::exit_cmux_internal()
 
 bool DTE::setup_cmux()
 {
+    if (cmux_term) {
```
euripedesrocha, line 181: Shall we provide some debug logs here?

euripedesrocha (approved):
LGTM. Left a few minor comments.

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: Thanks! changed

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: Makes sense, added `ESP_LOGE(...)`

fdkz:
Doesn't seem to fully work as intended with SIM7600.

Test case:

dce->set_mode(esp_modem::modem_mode::CMUX_MODE); // OK
physically reset modem
dce->set_mode(esp_modem::modem_mode::UNDEF); // OK
dce->set_mode(esp_modem::modem_mode::CMUX_MODE); // FAIL

esp_modem starts to send +++ and AT wrapped:
`\xF9\t\xEF\x07+++5\xF9`
`\xF9\t\xEF\x07AT\r5\xF9`

and doesn't recognize the unwrapped response by the modem:
`AT\r\r\nOK\r\n`

david-cermak:
> Doesn't seem to fully work as intended with SIM7600.
> 
> Test case:
> 
> dce->set_mode(esp_modem::modem_mode::CMUX_MODE); // OK physically reset modem dce->set_mode(esp_modem::modem_mode::UNDEF); // OK dce->set_mode(esp_modem::modem_mode::CMUX_MODE); // FAIL
> 
> esp_modem starts to send +++ and AT wrapped: `\xF9\t\xEF\x07+++5\xF9` `\xF9\t\xEF\x07AT\r5\xF9`
> 
> and doesn't recognize the unwrapped response by the modem: `AT\r\r\nOK\r\n`

That's not how the `UNDEF` mode is supposed to work. It just a meta-state indicating that all transitions are allowed, not that they'd necessarily succeed (nor it is a synonym to INIT or RESET state).

What you describe in your test case is more like a situation  suitable for mode detection (another planned feature), in other words: We don't know what state the modem is in, so we check how it responds to all protocols and set the mode accordingly.

PR #417, opened 2023-11-06
change(mdns): Update mdns task affinity config type (IDFGH-11387)

Dazza0:
ESP-IDF updates the task affinity config option from hex to int, and provides constant values for each affinity value. This commit syncs MDNS_TASK_AFFINITY with those changes.

See internal MR !26701 for more details.

### DIFF
--- a/components/mdns/Kconfig
+++ b/components/mdns/Kconfig
@@ -56,10 +56,10 @@ menu "mDNS"
     endchoice
 
     config MDNS_TASK_AFFINITY
-        hex
+        int
+        default FREERTOS_CORE0_AFFINITY if MDNS_TASK_AFFINITY_CPU0
+        default FREERTOS_CORE1_AFFINITY if MDNS_TASK_AFFINITY_CPU1
         default FREERTOS_NO_AFFINITY if MDNS_TASK_AFFINITY_NO_AFFINITY
-        default 0x0 if MDNS_TASK_AFFINITY_CPU0
-        default 0x1 if MDNS_TASK_AFFINITY_CPU1
 
     config MDNS_SERVICE_ADD_TIMEOUT_MS
         int "mDNS adding service timeout (ms)"


Dazza0:
@gabsuren @david-cermak Is the mdns component's bugfix version bumped for every change?

david-cermak:
> Is the mdns component's bugfix version bumped for every change?

No, we update version manually. You can check the CONTRIBUTION guidelines:

https://github.com/espressif/esp-protocols/blob/7c5a832821f0fac53cb9ae54e11a16a5a90b4297/CONTRIBUTING.md?plain=1#L63-L66

components/mdns/Kconfig
```diff
@@ -56,10 +56,10 @@ menu "mDNS"
     endchoice
 
     config MDNS_TASK_AFFINITY
-        hex
+        int
+        default FREERTOS_CORE0_AFFINITY if MDNS_TASK_AFFINITY_CPU0
```
david-cermak, line 60: Are you going backport these macros to IDF stable releases? Currently we use the same `mdns` branch for all IDF releases `>= v5.0`

Dazza0: @david-cermak There changes are going to be applied starting from ESP-IDF v5.2, so I think v5.2 will need to increase the minimum required version for mdns. Will this be an issue?

david-cermak: It would be great if we can avoid it, as the `mdns` is in maintenance mode (we don't plan new features), so bugfixing would result in double work.

We usually add `#ifdefs` to the code, but I don't think it's too easy with `Kconfig` (something like?)
```
    if !FREERTOS_CORE0_AFFINITY
        config FREERTOS_CORE0_AFFINITY
            int
            default 0
    endif
```

Dazza0:
@david-cermak  Sorry, looks like there's not a good way to introduce this change to the managed components in a non-breaking manner. I'll closing this PR (and associated MR) and will delay this change until ESP-IDF v6.0.

PR #419, opened 2023-11-08
[esp-modem]: Use std::string_view and (lw)span in (semi) private headers

david-cermak:
* Based on https://github.com/espressif/esp-protocols/pull/303
* Keeps backward compatibility in headers that still use std::string
* Move the header that needs c++17 to a separate folder
* Use c++17 only for IDF < `v5.0` (the default standard otherwise)


david-cermak:
Closing as this doesn't seem as a viable option. @euripedesrocha @0xFEEDC0DE64 please comment if you think otherwise.

PR #427, opened 2023-11-14
mdns: enable CONFIG_ESP_WIFI_ENABLED if CONFIG_SOC_WIFI_SUPPORTED is also enabled (for ESP-IDF <= 5.1)

gabsuren:


components/mdns/mdns.c
```diff
@@ -26,8 +26,7 @@
 #endif
 
 #if ESP_IDF_VERSION <= ESP_IDF_VERSION_VAL(5, 1, 0)
-// IDF <= v5.1 does not support enabling/disabling esp-wifi
-#define MDNS_ESP_WIFI_ENABLED 1
+#define MDNS_ESP_WIFI_ENABLED CONFIG_SOC_WIFI_SUPPORTED
```
david-cermak, line 29: Is this `CONFIG_SOC_WIFI_SUPPORTED` defined in say `v4.3` ?

gabsuren: `mdns` as a component should be used only if `idf: ">=5.0"`, why we need to check `v4.3` ?

me-no-dev: ah so true... mdns is part of IDF before 5.0...

david-cermak: I see. no more concerns if we support mdns only from `v5.0`

david-cermak (approved):

PR #443, opened 2023-11-28
bump(websocket): 1.2.0 -> 1.2.1

gabsuren:


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1228,7 +1228,7 @@ int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client,
         goto unlock_and_return;
     }
     ret = esp_websocket_client_send_with_exact_opcode(client, opcode | WS_TRANSPORT_OPCODES_FIN, data, len, timeout);
-    if (ret <= 0) {
+    if (ret < 0) {
```
david-cermak, line 1231: Was this forgotten in ba335880088feb665d1beab29008203fbd91ca0e ?

gabsuren: Yes, I rechecked it. From the return value perspective it is the same, in both case `less than 0` or `equal 0`, it will send the same return value. But to print `error message` should be only if it is less than 0, IMO.

gabsuren: Also need to improve example tests, so it can check this changes. I will do it after.

david-cermak (approved):

Issue #426, opened 2023-11-13
Esp Modem panic_abort on unsolicited data from SIM card (IDFGH-11441)


antfarming:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I am using a roaming SIM card that will perform an OTA occasionally. During this process you can expect unsolicited data from the module, due to the sim tool kit. I cannot stop the sim from doing this, or anticipate when it will.

The modem library does not seem to like this, and I am not sure how to dig further into the issue. I have tried to provide as much info as possible below. As you can see, I request the CGREG status, and before it returns a valid response, sim tool kit data is coming through first.

```
[V][esp_modem_command_library.cpp:522] get_network_attachment_cgreg()::(79409):command_lib: get_network_attachment_cgreg
[V][esp_modem_command_library.cpp:56] generic_get_string()::(79409):command_lib: generic_get_string
[V][esp_modem_command_library.cpp:66] operator()()::(79429):command_lib: Token: {
[V][esp_modem_command_library.cpp:66] operator()()::(79429):command_lib: Token: {+MSTK: 11, "D02A810301400382028182050035010339020200470A036F74610573757065723C030100353E052108080808"}
[V][esp_modem_command_library.cpp:66] operator()()::(79449):command_lib: Token: {
[V][esp_modem_command_library.cpp:66] operator()()::(79459):command_lib: Token: {+CGEV: ME PDN DEACT 9}
[V][esp_modem_command_library.cpp:66] operator()()::(79469):command_lib: Token: {
[V][esp_modem_command_library.cpp:66] operator()()::(79469):command_lib: Token: {+MSTK: 11, "D02A810301400382028182050035010339020200470A036F74610573757065723C030100353E052108080808"}
[V][esp_modem_command_library.cpp:66] operator()()::(79489):command_lib: Token: {
[V][esp_modem_command_library.cpp:66] operator()()::(79499):command_lib: Token: {+CGEV: ME PDN DEACT 9}
[V][esp_modem_command_library.cpp:66] operator()()::(79509):command_lib: Token: {
[...]

PR #446, opened 2023-11-29
fix(modem): Fixed inconsistent state on data after OK

david-cermak:
If we received response in two chunks and the first one completed the command (result=OK or FAIL), and the second chunk preempts command processing, then we performed processing again rewriting the result back to TIMEOUT. This would invalidate the command, but also causes an exception: `ESP_MODEM_THROW_IF_ERROR(ESP_ERR_INVALID_STATE)`

Fixed by checking if the processing already finished in process_line().

Closes https://github.com/espressif/esp-protocols/issues/426


euripedesrocha (approved):

david-cermak:
Thanks for the review @dset0x @euripedesrocha !

PR #447, opened 2023-12-01
fix(modem): Added test injecting unexpected replies

david-cermak:
This test exhibits an issue of getting unexpected data just after completing a command. The DTE rewrites the state to TIMEOUT, but the signal has been already triggered, thus results in an unexpected state.

Related to Closes https://github.com/espressif/esp-protocols/issues/426


euripedesrocha (approved):

david-cermak:
Thanks for the review @euripedesrocha 

The test (previously failing on https://github.com/espressif/esp-protocols/actions/runs/7058401985/job/19213908309) now passed after including bf99f287bc5a84b12d56f1e1cb3a63c6034212f0

Issue #455, opened 2023-12-08
Timeout for setting PDP (AT+CGDCONT) in esp-modem is probably wrong? (IDFGH-11638)


huming2207:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### What component are you using? If you choose Other, provide details in More Information.

esp_modem

### component version

82c2cf8936d4d363617825c3ed4028d1b0d20386

### IDF version.

v5.1.1-678-g8c6114d0d2

### More Information.

Hi there,

It seems like the `set_pdp_context()` call's timeout delay is way too long, see: https://github.com/espressif/esp-protocols/blob/master/components/esp_modem/src/esp_modem_command_library.cpp#L269

It supposed to be 300ms according to Quectel's AT manual, or 1 second for other modems:

<img width="970" alt="image" src="https://github.com/espressif/esp-protocols/assets/4463497/74e65445-9bc4-4f8d-adcc-6eec6e1f9442">

This may cause a hang for up to 150s. We have reproduced that for multiple times in the recent a few days. I guess if this times out in 300ms, the modem should be reset and start over the network registration again.

Regards,
Jackson

david-cermak:
Thanks for reporting this and sorry for the inconvenience.
This is definitely wrong, I'll push a fix.

huming2207:
@david-cermak great, thanks!

Meanwhile I've seen your pull request here: https://github.com/espressif/esp-protocols/pull/456/commits/aa5003c73bc0bf8da363117af672139b7fd95b33#diff-cb4f0037ddffd05f90dbe90c2a6b6469e5f33673fe100376fd0fc2d5258c5349R58

I guess the C++ API can be overloaded, instead of another function names `set_pdp_context_with_timeout()`?

Jackson

huming2207:
@david-cermak thanks! The latest commit looks good!

david-cermak:
Thanks for your suggestion on function overloads! I wasn't originally thinking of that as this is a part of the shared/generated API (as mentioned here https://github.com/espressif/esp-protocols/pull/456#discussion_r1431601188), but makes perfect sense and the fix look cleaner!

Thanks for helping with the review @huming2207 !

PR #456, opened 2023-12-08
fix(modem): Fixed set_pdp_context() command timeout

david-cermak:
Closes https://github.com/espressif/esp-protocols/issues/455

### DIFF
--- a/components/esp_modem/include/cxx_include/esp_modem_command_library.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_command_library.hpp
@@ -55,6 +55,7 @@ command_result power_down_sim70xx(CommandableIf *t);
 command_result set_network_bands_sim76xx(CommandableIf *t, const std::string &mode, const int *bands, int size);
 command_result power_down_sim8xx(CommandableIf *t);
 command_result set_data_mode_sim8xx(CommandableIf *t);
+command_result set_pdp_context(CommandableIf *t, PdpContext &pdp, uint32_t timeout_ms);
 
 /**
  * @}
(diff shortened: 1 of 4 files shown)


components/esp_modem/include/cxx_include/esp_modem_command_library.hpp
```diff
@@ -55,6 +55,7 @@ command_result power_down_sim70xx(CommandableIf *t);
 command_result set_network_bands_sim76xx(CommandableIf *t, const std::string &mode, const int *bands, int size);
 command_result power_down_sim8xx(CommandableIf *t);
 command_result set_data_mode_sim8xx(CommandableIf *t);
+command_result set_pdp_context_with_timeout(CommandableIf *t, PdpContext &pdp, uint32_t timeout_ms);
```
david-cermak, line 58:
```suggestion
command_result set_pdp_context(CommandableIf *t, PdpContext &pdp, uint32_t timeout_ms);
```

components/esp_modem/src/esp_modem_command_library.cpp
```diff
@@ -261,12 +261,17 @@ command_result set_echo(CommandableIf *t, bool on)
     return generic_command_common(t, "ATE0\r");
 }
 
-command_result set_pdp_context(CommandableIf *t, PdpContext &pdp)
```
euripedesrocha, line 264: Consider adding the timeout as a default parameter.

euripedesrocha (approved):

components/esp_modem/src/esp_modem_command_library.cpp
david-cermak: I'd prefer to keep the two prototypes separately, since this API is one of the generic one, declared here:

https://github.com/espressif/esp-protocols/blob/82c2cf8936d4d363617825c3ed4028d1b0d20386/components/esp_modem/include/generate/esp_modem_command_declare.inc#L98-L102

from which we generate C-API and C++ API (plus others, like docs and DCE's methods)

david-cermak:
Thanks for the review @euripedesrocha

Issue #460, opened 2023-12-19
websocket client decided to send first a FIN then a continuation, which is not allowed?! (IDFGH-11736)


0xFEEDC0DE64:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### What component are you using? If you choose Other, provide details in More Information.

esp_websocket_client

### component version

modem-v1.0.5

### IDF version.

v5.3-dev-1043-g8c9e29898f

### More Information.

websocket client unexpectatly sends first a FIN text message, then a continuation, which causese my websocket server to kick out esp32 connections.
I have added a few log statements to esp-protocols to follow this:
```
I (107646) CMDHANDLER: connected! socketIdentifier=cloud                             
I (107646) websocket_client: esp_websocket_client_send_with_exact_opcode() opcode=129 len=223                                                                              
I (107646) transport_ws: esp_transport_ws_send_raw() opcode=129 len=223              
I (107656) transport_ws: _ws_write() opcode=129 len=223                              
I (107666) websocket_client: esp_websocket_client_send_with_exact_opcode() esp_transport_ws_send_raw() returned 223                                                        
I (107686) CMDHANDLER: executing command type=hello socketIdentifier=cloud is_secured=true                                                                                 
I (107686) CMDHANDLER: skip=0                                                        
W (107706) CMDHANDLER: scan overflowed                                               
I (107706) websocket_client: esp_websocket_client_send_with_exact_opcode() opcode=129 len=2435                                                                             
I (107706) transport_ws: esp_transport_ws_send_raw() opcode=129 len=1024             
I (107716) transport_ws: _ws_write() opcode=129 len=1024                             
I (107726) websocket_client: esp_websocket_client_send_with_exact_opcode() esp_transport_ws_send_raw() returned 1024                                                       
I (107736) transport_ws: esp_transport_ws_send_raw() opcode=0 len=1024               
I (107736) transport_ws: _ws_write() opcode=0 len=1024                               
I (107746) websocket_client: esp_websocket_client_send_with_exact_opcode() esp_transport_ws_send_raw() returned 1024                                                       
[...]

PR #466, opened 2023-12-22
bump(websocket): 1.2.1 -> 1.2.2

gabsuren:
1.2.2
Bug Fixes
- continuation after FIN in websocket client (#460) (774d1c75e6)
- Re-applied refs to common comps idf_component.yml (9fe44a4504)

### DIFF
--- a/components/esp_websocket_client/.cz.yaml
+++ b/components/esp_websocket_client/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(websocket): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py esp_websocket_client
   tag_format: websocket-v$version
-  version: 1.2.1
+  version: 1.2.2
   version_files:
   - idf_component.yml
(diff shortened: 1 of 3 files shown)


david-cermak (changes requested):
I've checked the original PR and added few comments: https://github.com/espressif/esp-protocols/pull/368

I'd suggest looking at and addressing these ~~three~~ two issues:
1) ~~`esp_websocket_client_send_text()` API might not include FIN flag~~
2) ~~`esp_websocket_client_send_bin()` API might send message without the mandatory FIN flag.~~
3) Update`esp_websocket_client_send_with_exact_opcode()` to include one additional parameter indicating whether or not to add the FIN flag in the last fragment.
4) These APIs do not lock the client not check parameters:
 * `esp_websocket_client_send_text_partial()`
 * `esp_websocket_client_send_cont_msg()`
 * `esp_websocket_client_send_bin_partial()`
 * `esp_websocket_client_send_fin()`
5) Check and fix the behaviour under `CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER=y`

david-cermak (approved):
Let's go ahead and create a new release, but please fix the issue later

PR #469, opened 2024-01-03
fix(mdns): fix copy delegated host addr (IDFGH-11829)

DejinChen:
fix loop problem in function '_copy_delegated_host_address_list'


david-cermak (approved):
Thanks for the fix!

Issue #459, opened 2023-12-14
Build fails if ETH is enabled, but removed from predefined netifs


marekmaskarinec:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Setting `CONFIG_MDNS_PREDEF_NETIF_ETH` to `n` while `CONFIG_ETH_ENABLED` is set to `y` will result in compilation errors. This is because the `esp_eth.h` header will not be included, but it's being used.

One of the errors:

```
mdns.c:4194:27: error: 'ETH_EVENT' undeclared (first use in this function); did you mean 'IP_EVENT'?
 4194 |         if (event_base == ETH_EVENT) {
      |                           ^~~~~~~~~
      |                           IP_EVENT
```

david-cermak:
Fixed in 6d2c475c2039e57c83c7fe2b679b18501405c9de and fc59f87c4e829bc92b71d938e8c114a618a51959

PR #479, opened 2024-01-10
fix(mdns): Fix compile error when MDNS_PREDEF_NETIF_ETH is defined, but ETH_ENABLED is not (IDFGH-11868)

marekmaskarinec:
This is a fix for the issue #459.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -132,7 +132,7 @@ static inline esp_netif_t *esp_netif_from_preset_if(mdns_predef_if_t predef_if)
         return esp_netif_get_handle_from_ifkey("WIFI_STA_DEF");
     case MDNS_IF_AP:
         return esp_netif_get_handle_from_ifkey("WIFI_AP_DEF");
-#if CONFIG_ETH_ENABLED
+#if CONFIG_ETH_ENABLED && CONFIG_MDNS_NETIF_ETH
     case MDNS_IF_ETH:
         return esp_netif_get_handle_from_ifkey("ETH_DEF");
 #endif
@@ -4189,7 +4189,7 @@ void mdns_preset_if_handle_system_event(void *arg, esp_event_base_t event_base,
         }
     } else
 #endif
-#if CONFIG_ETH_ENABLED
+#if CONFIG_ETH_ENABLED && CONFIG_MDNS_NETIF_ETH
         if (event_base == ETH_EVENT) {
             switch (event_id) {
             case ETHERNET_EVENT_CONNECTED:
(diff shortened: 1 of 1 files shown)


david-cermak (commented):
Thanks for posting the fix!

I think that we shouldn't try to handle Ethernet events if `CONFIG_MDNS_PREDEF_NETIF_ETH=n`. 
Would this change fix the issue?
```patch
diff --git a/components/mdns/mdns.c b/components/mdns/mdns.c
index 1149ff7..2c95bdc 100644
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -3925,7 +3925,7 @@ void mdns_preset_if_handle_system_event(void *arg, esp_event_base_t event_base,
             break;
         }
     }
-#if CONFIG_ETH_ENABLED
+#if CONFIG_ETH_ENABLED && CONFIG_MDNS_PREDEF_NETIF_ETH
     else if (event_base == ETH_EVENT) {
         switch (event_id) {
         case ETHERNET_EVENT_CONNECTED:
```

marekmaskarinec:
I didn't do this because I thought the user could add their own
Ethernet netif, even if the predefined one is disabled. Should I add
this?

david-cermak:
> I thought the user could add their own Ethernet netif, even if the predefined one is disabled.

Yes, this is the idea, but if users add their own interfaces, they have to take care of event handling in the application code (as the library couldn't possibly handle status changes of various netifs which are under user control), there's a convenient API provided just for this purpose:

https://github.com/espressif/esp-protocols/blob/5ab699d6f4906a3ce6c97377f798944533d7152e/components/mdns/include/mdns.h#L812-L831

The role of predefined interfaces is a legacy handling of the default netifs that were present (hardcoded) in older IDF versions.

> Should I add this?

If you're willing to address this issue, I'd suggest you update your commit accordingly. (but only if you feel like you'd like to contribute, which is much appreciated :-)

marekmaskarinec:
> I think that we shouldn't try to handle Ethernet events if CONFIG_MDNS_PREDEF_NETIF_ETH=n.

Done.

components/mdns/mdns.c
```diff
@@ -18,7 +18,7 @@
 #include "esp_log.h"
 #include "esp_random.h"
 
-#if CONFIG_ETH_ENABLED && CONFIG_MDNS_PREDEF_NETIF_ETH
+#if CONFIG_ETH_ENABLED
```
david-cermak, line 21: suggest reverting this change

```suggestion
#if CONFIG_ETH_ENABLED && CONFIG_MDNS_PREDEF_NETIF_ETH
```

The reason is that even if we're using a custom Ethernet interface there's no need for `esp_eth` headers (as the netif itself is well abstracted), which helps reduce component coupling.

david-cermak (commented):
Thanks for the effort! One minor nitpick, LGTM otherwise.

david-cermak:
Saw that you had pushed from dce4339 to e529bee, but there's no diff in the code.

Also please note that we've got some trouble with our CI, most of the tests are failing (should be fixed soon).
(This [one](https://github.com/espressif/esp-protocols/actions/runs/7503602020/job/20428699989?pr=479) however relies on the Ethernet module dependency -- should be fixed once you address the last comment)

marekmaskarinec:
Sorry, I seem to have committed incorrectly.

david-cermak (approved):
Thanks for the contribution!

PR #483, opened 2024-01-15
fix(common): Fixed scripts for build test failure

espressif-abhikroy:
Added changes to fix build test failures.


david-cermak (commented):
I think we still need to fix the mdns tests.
LGTM otherwise

gabsuren (approved):

david-cermak (approved):
Thanks for the fix!

Issue #485, opened 2024-01-15
set APN after esp_modem_new_dev() is called? (IDFGH-11898)


JB-DX:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Is there a way to define an APN after `esp_modem_new_dev()` is called?

In our case we read the APN off the SIM-card. This requires a valid dce handle, which we only get from `esp_modem_new_dev()`. So we end up in a catch-22 situation.
Currently we create a dce just to read the SIM-card, then deinit() and init() the dce again.
Not that optimal, but doable.
Is there a better approach?

```
 esp_modem_dce_config_t dce_config = ESP_MODEM_DCE_DEFAULT_CONFIG(myAPN);  
 ESP_ERROR_CHECK(esp_event_handler_register(IP_EVENT, ESP_EVENT_ANY_ID, &on_ip_event, NULL));
 ESP_ERROR_CHECK(esp_event_handler_register(NETIF_PPP_STATUS, ESP_EVENT_ANY_ID, &onPPP_Event, NULL));
....
dce= esp_modem_new_dev(ESP_MODEM_DCE_SIM800, &dte_config, &dce_config, gsmNetif);
```

david-cermak:
You can set APN via the pdp context as shown in

https://github.com/espressif/esp-protocols/blob/585e4b30b28e21ad3203519318f8252cef56be9f/components/esp_modem/examples/modem_console/main/modem_console_main.cpp#L393-L394

(I'll add corresponding C API if needed)

JB-DX:
Yes please add the C API,  since I only use C

PR #491, opened 2024-01-17
fix(modem): Added C-API to configure APN

david-cermak:
Closes https://github.com/espressif/esp-protocols/issues/485

### DIFF
--- a/components/esp_modem/test/README.md
+++ b/components/esp_modem/test/README.md
@@ -1,0 +1,47 @@
+# ESP-Modem Testing
+
+This folder contains automated and manual tests for esp-modem component. Beside these tests, some jobs are executed in CI to exercise standard examples (please refer to the CI definition and CI related sdkconfigs in examples).
+
+List of test projects:
+
+* `host_test` -- esp_modem is build on host (linux), modem's terminal in mocked using Loobpack class which creates simple responders to AT and CMUX mode. This test is executed in CI.
+* `target`  -- test executed on target with no modem device, just a pppd running on the test runner. This test is executed in CI.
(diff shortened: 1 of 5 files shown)


components/esp_modem/test/README.md
```diff
@@ -0,0 +1,46 @@
...
+List of test projects:
+
+* `host_test` -- esp_modem is build on host (linux), modem's terminal in mocked using Loobpack class which creates simple responders to AT and CMUX mode. This test is executed in CI.
+* `target`  -- test executed on target with no modem device, just a pppd running on the test runner. This test is executed in CI.
+* `target_ota` -- Manual test which perform OTA over PPP.
+* `target_iperf` -- Manual test to measure data throughput via PPP.
+
+## Manual testing
+
+Prior to every `esp_modem` release, these manual tests must be execute and pass:
```
euripedesrocha, line 14: Is it possible to have some auto reminder on PR?

euripedesrocha (approved):

components/esp_modem/test/README.md
david-cermak: Good idea, I'll think about it and check if we can generate some TODO list on PRs with bump commits

david-cermak: I've checked that it's possible to read PR description and check for certain text. This way we could enforce a task list with marked `manual test` field for all PRs that contain a bump commit. Additionally we could provide a PR template including this task list. The drawback is that we're in mono repo and the only component that needs manual testing is esp_modem. Also we would need to document this process.
Considering the advantages and effort, I've decided to create a tracker IDF-9074 to migrate the manual tests to CI.

Please resolve if you agree.

PR #495, opened 2024-01-23
fix(websocket): Continue waiting for TCP connection to be closed (IDFGH-11965)

eht-datarespons:
An issue was observed where the websocket client did not send WEBSOCKET_EVENT_CLOSED after server closed the websocket.

An alternative solution to fix the issue would be  to remove the if instead. It could be more robust, but the existing comment seem to suggest that it was not the original intent.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1039,10 +1039,9 @@ static void esp_websocket_client_task(void *pv)
             ESP_LOGD(TAG, " Waiting for TCP connection to be closed by the server");
             int ret = esp_transport_ws_poll_connection_closed(client->transport, 1000);
             if (ret == 0) {
-                // still waiting
-                continue;
-            }
-            if (ret < 0) {
+                ESP_LOGW(TAG, "Did not get TCP close within expected delay");
+
+            } else if (ret < 0) {
                 ESP_LOGW(TAG, "Connection terminated while waiting for clean TCP close");
             }
             client->run = false;
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1040,7 +1040,7 @@ static void esp_websocket_client_task(void *pv)
             int ret = esp_transport_ws_poll_connection_closed(client->transport, 1000);
             if (ret == 0) {
                 // still waiting
-                break;
+                continue;
             }
             if (ret < 0) {
                 ESP_LOGW(TAG, "Connection terminated while waiting for clean TCP close");


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1040,7 +1040,7 @@ static void esp_websocket_client_task(void *pv)
             int ret = esp_transport_ws_poll_connection_closed(client->transport, 1000);
             if (ret == 0) {
                 // still waiting
-                break;
+                continue;
```
david-cermak, line 1043: But that would loop forever if the connection closure wasn't "confirmed" by the server, correct?

We should let the client to close abruptly with "some" timeout.

But we could make the timeout configurable, e.g. in [L#1040](https://github.com/espressif/esp-protocols/pull/495/files#diff-53ead0f3896f79090b1ed1e79e1cca84ac4e505faa1bb76727ab856dabebc297L1040)

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1040,7 +1040,7 @@ static void esp_websocket_client_task(void *pv)
             int ret = esp_transport_ws_poll_connection_closed(client->transport, 1000);
```
david-cermak, line 1040: For example:
```suggestion
            int ret = esp_transport_ws_poll_connection_closed(client->transport, CONFIG_ESP_WS_CLIENT_CLOSE_TOUT);
```

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1040,7 +1040,7 @@ static void esp_websocket_client_task(void *pv)
...
             if (ret == 0) {
                 // still waiting
-                break;
+                continue;
             }
             if (ret < 0) {
```
eht-datarespons, line 1045: My impression from the testing I have done is that esp_websocket_client treats the connection as closed during the esp_transport_ws_poll_connection_closed you refer to. Which I think is valid, but it means that having a longer timeout could cause other problems where the application attempt to send more messages after the server initiates the close and before the client sends the WEBSOCKET_EVENT_CLOSED. I think the current timeout should be sufficient. 

how about:
```suggestion
            if (ret == 0) {
                ESP_LOGW(TAG, "Did not get TCP close within expected delay");

            }else if (ret < 0) {
```

eht-datarespons: This will treat ret == 0 similarly to ret < 0 and won't break out of the loop before  dispatching the WEBSOCKET_EVENT_CLOSED.

eht-datarespons: Here is a log for what could happen with a long continue instead of break.
```
W (13:46:33.779) OCPP_LISTENER: Recieved connection closed
D (13:46:33.780) websocket_client: Received close frame
D (13:46:33.785) websocket_client: Closing initiated by the server, sending close frame
D (13:46:33.799) websocket_client:  Waiting for TCP connection to be closed by the server
I (13:46:33.803) OCPP_TASK: Sending next call (StatusNotification) [470e6d50-46d0-4bcb-b697-e7633c43d4e1]
E (13:46:33.812) websocket_client: Websocket client is not connected
E (13:46:33.819) OCPP_TASK: Got websocket error when sending ocpp message
E (13:46:33.827) OCPP_TASK: Failing call
E (13:46:33.831) OCPP CALLBACK  : [status notification|470e6d50-46d0-4bcb-b697-e7633c43d4e1]: (InternalError) 'CP unable to send' ''
I (13:46:33.844) OCPP_TASK: Registration status: accepted
I (13:46:33.853) OCPP_TASK: Preparing status notification
I (13:46:33.857) OCPP_TASK: Sending status notification to queue
I (13:46:33.863) OCPP_TASK: starting heartbeat with interval 15 sec
I (13:46:33.869) OCPP SMART     : Initializing smart charging
I (13:46:33.876) OCPP SMART     : Awaiting renewal (4294967295 sec) or notification
I (13:46:33.884) OCPP: Restarting clock aligned meter value timer
W (13:46:33.891) OCPP: Clock aligned meter values are disabled
I (13:46:33.897) OCPP: Setting firmware update state
I (13:46:33.916) OCPP: Handling websocket event
I (13:46:33.917) OCPP_TASK: Sending next call (StatusNotification) [80b242df-f1ef-4f98-9b70-0ce3cd6fa5c4]
E (13:46:33.921) websocket_client: Websocket client is not connected
E (13:46:33.928) OCPP_TASK: Got websocket error when sending ocpp message
E (13:46:33.935) OCPP_TASK: Failing call
E (13:46:33.940) OCPP CALLBACK  : [status notification|80b242df-f1ef-4f98-9b70-0ce3cd6fa5c4]: (InternalError) 'CP unable to send' ''
E (13:46:33.953) OCPP: Unable to handle ocpp call
W (13:46:33.958) OCPP: Transition offline at: 1706276793
W (13:46:33.964) OCPP: Blocking generic and transaction messages. Storing current mask: 0
I (13:46:33.977) CLOUD LISTENER : MQTT_EVENT_PUBLISHED, msg_id=10961
I (13:46:34.692) CLOUD LISTENER : MQTT_EVENT_PUBLISHED, msg_id=40227
D (13:46:34.806) websocket_client:  Waiting for TCP connection to be closed by the server
W (13:46:35.408) OCPP_LISTENER: WEBSOCKET_EVENT_CLOSED
```
Note that it logs "Websocket client is not connected" from the websocket client before the last "Waiting for TCP connection to be closed" message.

If the timeout would be increased then this would be more prominent. This is while attempting to stress the system

david-cermak: I see, you're right, We probably have to let the entire loop running to process remaining messages..
I like your suggestion [above](https://github.com/espressif/esp-protocols/pull/495/files#r1467661942)

eht-datarespons:
@david-cermak Are there any other changes i should make or is the current state of the pull request acceptable?

david-cermak (approved):
sorry for the delay. The change LGTM, thanks for posting this PR!

PR #497, opened 2024-01-26
feat(mdns): Make including mdns_console KConfig

rretanubun:
The CLI that mdns_console offers is very useful for debugging and bootstrapping mDNS onto new projects, but some projects may not want console CLI in place to reduce binary size.

This gives a KConfig option to do so.

### DIFF
--- a/components/mdns/CMakeLists.txt
+++ b/components/mdns/CMakeLists.txt
@@ -4,6 +4,12 @@ else()
     set(MDNS_NETWORKING "mdns_networking_lwip.c")
 endif()
 
+if(CONFIG_MDNS_ENABLE_CONSOLE_CLI)
+    set(MDNS_CONSOLE "mdns_console.c")
+else()
+    set(MDNS_CONSOLE "")
+endif()
+
 idf_build_get_property(target IDF_TARGET)
 if(${target} STREQUAL "linux")
     set(dependencies esp_netif_linux esp_timer esp_system)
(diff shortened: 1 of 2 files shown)


igrr:
@rretanubun do you see mdns_console contributing to binary size even when not calling mdns_console_register? (You can check with `idf.py size*` commands.)
I think this should normally not happen, as the linker will eliminate all the unused object files and functions from the resulting binary.

rretanubun:
@igrr : No I do not, the linker does eliminate it. I was thinking adding an upstream supported KConfig option to allow projects to build with
```
#ifdef CONFIG_MDNS_ENABLE_CONSOLE_CLI
    mdns_console_register();
#endif
```
then each sdkconfig variant controls if the project builds gets `mdns console on the CLI` or not and the community going forward all just use the same KConfig name.

david-cermak:
I agree with Ivan that this change won't affect the binary size, but may help reduce build time and since the console API is rarely used within mdns component I would tend to accept this PR, WYT @suren-gabrielyan-espressif ?

The ultimate goal should migration of the mdns_console to other console components, cc @espressif-abhikroy

david-cermak (approved):

gabsuren (approved):

david-cermak:
@rretanubun Thanks for your contribution!
@suren-gabrielyan-espressif Thanks for the quick review!

PR #499, opened 2024-01-29
feat(console): Added component for iperf command

espressif-abhikroy:
The iperf component, previously located in esp-idf **examples/common_components/iperf**, has been relocated to esp-protocols as a managed component. It encompasses an iperf network performance tester implementation and includes a console for executing the 'iperf' command in any example project.


david-cermak:
This could be closed now, correct @espressif-abhikroy ?

PR #500, opened 2024-01-30
fix(websocket): added dependency checks, sdkconfig.defaults and refined README.md

gabsuren:


components/esp_websocket_client/examples/linux/sdkconfig.defaults
```diff
@@ -0,0 +1,8 @@
+CONFIG_IDF_TARGET="linux"
+CONFIG_IDF_TARGET_LINUX=y
+CONFIG_ESP_EVENT_POST_FROM_ISR=n
+CONFIG_ESP_EVENT_POST_FROM_IRAM_ISR=n
+CONFIG_WEBSOCKET_URI="ws://echo.websocket.events"
+CONFIG_WEBSOCKET_URI_FROM_STRING=y
+CONFIG_WEBSOCKET_URI_FROM_STDIN=n
+CONFIG_EXAMPLE_CONNECT_WIFI=n
```
david-cermak, line 8: Are these config option used in the example's project configuration? If so I'd suggest we remove it, as I think they won't work.

david-cermak (approved):
LGTM otherwise

components/esp_websocket_client/examples/linux/sdkconfig.defaults
gabsuren: @david-cermak Thank you for noticing. Removed

PR #513, opened 2024-02-26
feat(console): wifi commands updates

espressif-abhikroy:
It adds the option to disable auto-registration of wifi command from menuconfig.
And adds the option for the wifi command to connect to a pre-configured network.


components/console_cmd_wifi/console_wifi.c
```diff
@@ -184,7 +186,13 @@ static esp_err_t wifi_sta_join_op(wifi_op_t *self, int argc, char *argv[])
         return ESP_FAIL;
     }
 
-    strlcpy((char *) wifi_config.sta.ssid, argv[self->start_index + 1], sizeof(wifi_config.sta.ssid));
+    if (self->arg_cnt == 3) {
+        strcpy((char *) wifi_config.sta.ssid, CONFIG_WIFI_CMD_NETWORK_SSID);
+        strcpy((char *) wifi_config.sta.password, CONFIG_WIFI_CMD_NETWORK_PASSWORD);
```
david-cermak, line 191: Nitpick: Can we also use `strlcpy()` here, as the Kconfig options could be very long, too, I think?

david-cermak (approved):
LGTM!

PR #515, opened 2024-02-28
[ws-client]: Fix usage of the dynamic buffer

david-cermak:
Fixes crash when using ws Tx functions if `CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER=y`

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1208,26 +1208,24 @@ int esp_websocket_client_send_fin(esp_websocket_client_handle_t client, TickType
 
 int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client, ws_transport_opcodes_t opcode, const uint8_t *data, int len, TickType_t timeout)
 {
-    int ret = ESP_OK;
+    int ret = -1;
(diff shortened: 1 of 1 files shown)


gabsuren (approved):

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1204,33 +1208,27 @@ int esp_websocket_client_send_fin(esp_websocket_client_handle_t client, TickType
 
 int esp_websocket_client_send_with_opcode(esp_websocket_client_handle_t client, ws_transport_opcodes_t opcode, const uint8_t *data, int len, TickType_t timeout)
 {
-    int ret = ESP_OK;
+    int ret = -1;
```
euripedesrocha, line 1211: Isn't this changing the original logic?

david-cermak: The logic is unchanged, just the default return value is `-1`, so each time we fail we don't have to re-assign this to `-1`.
(the original logic didn't actually use this value anywhere, as if all the subsequent check passed we set `ret = esp_websocket_client_s...()` )

euripedesrocha (approved):

PR #520, opened 2024-03-06
feat(websocket): Support DS peripheral for mutual TLS (IDFGH-12285)

johanstokking:
Fantastic project guys. Here's something I use that you might want to take into consideration: supporting the DS peripheral for mutual TLS.

Example usage with https://github.com/espressif/esp_secure_cert_mgr/:

```c
    char *client_cert = NULL;
    uint32_t client_cert_len = 0;
    esp_err_t ret = esp_secure_cert_get_device_cert(&client_cert, &client_cert_len);
    assert(esp_ret == ESP_OK);

    esp_ds_data_ctx_t *ds_data = esp_secure_cert_get_ds_ctx();
    assert(ds_data != NULL);

    esp_websocket_client_config_t config = {
        .uri = "wss://echo.websocket.org",
        .client_cert = client_cert, /* in this example it's PEM so client_cert_len is not set */
        .client_ds_data = ds_data,
        .crt_bundle_attach = esp_crt_bundle_attach,
    };
```

I considered adding this to the example project, but I feel like it gets bloated quickly with dependencies and ifdefs.


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -519,6 +522,10 @@ static esp_err_t esp_websocket_client_create_transport(esp_websocket_client_hand
             } else {
                 esp_transport_ssl_set_client_key_data_der(ssl, client->config->client_key, client->config->client_key_len);
             }
+#if CONFIG_ESP_TLS_USE_DS_PERIPHERAL
+        } else if (client->config->client_ds_data) {
+            esp_transport_ssl_set_ds_data(ssl, client->config->client_ds_data);
```
johanstokking, line 527: The signature here is:

```c
void esp_transport_ssl_set_ds_data(esp_transport_handle_t t, void *ds_data)
```

Hence I'm using `void *` and not `esp_ds_data_ctx_t *` to avoid a dependency on Mbed TLS.

johanstokking:
@gabsuren Can you shed some light on this? Thanks!

gabsuren:
> @gabsuren Can you shed some light on this? Thanks!

Sorry for the late response, it's on the line with https://github.com/espressif/esp-protocols/pull/621

david-cermak (approved):

components/esp_websocket_client/esp_websocket_client.c
david-cermak: Agree with using `void *` here

david-cermak (approved):
Nice, thanks for the updates (and those little fixes to the docs)!

david-cermak:
> This is what the TLS section looks like:

sorry, hit edit instead of reply. wanted to say that we don't generate the docs for each target, only one static page.

PR #522, opened 2024-03-08
mdns add browse support (IDFGH-12512)

zwx1995esp:
This MR introduces a new query mode: browse(like the `avahi-browse -r _test._tcp`).

`mdns_browse_new` is used for starting a browse, it will add a browse in the link list of `_mdns_server`, if there is some service start or leave, it will be notified by the callback passed in the API. When the service is leaved, the ttl will be `0`, the related result will be freed after syncing to the notifier callback.

`mdns_browse_delete` is used for stop the browse query and free the related memory.

All apis are followed the mdns action mechanism.

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -26,6 +26,11 @@ extern "C" {
  */
 typedef struct mdns_search_once_s mdns_search_once_t;
 
+/**
+ * @brief   Daemon query handle
+ */
+typedef struct mdns_browse_s mdns_browse_t;
+
 typedef enum {
     MDNS_EVENT_ENABLE_IP4                   = 1 << 1,
     MDNS_EVENT_ENABLE_IP6                   = 1 << 2,
@@ -97,6 +102,7 @@ typedef struct mdns_result_s {
 } mdns_result_t;
 
 typedef void (*mdns_query_notify_t)(mdns_search_once_t *search);
+typedef void (*mdns_browse_notify_t)(mdns_result_t *result);
 
 /**
  * @brief  Initialize mDNS on given interface
@@ -829,6 +835,28 @@ esp_err_t mdns_unregister_netif(esp_netif_t *esp_netif);
  *     - ESP_ERR_NO_MEM         memory error
  */
 esp_err_t mdns_netif_action(esp_netif_t *esp_netif, mdns_event_actions_t event_action);
+
+/**
+ * @brief   Browse mDNS for a service `_service._proto`.
+ *
+ * @param service  Pointer to the `_service` which will be browsed.
+ * @param proto    Pointer to the `_proto` which will be browsed.
+ * @param notifier The callback which will be called when the browsing service changed.
+ * @return mdns_browse_t pointer to new browse object if initiated successfully.
+ *         NULL otherwise.
+ */
+mdns_browse_t *mdns_browse_new(const char *service, const char *proto, mdns_browse_notify_t notifier);
+
+/**
+ * @brief   Stop the `_service._proto` browse.
+ * @param service  Pointer to the `_service` which will be browsed.
+ * @param proto    Pointer to the `_proto` which will be browsed.
+ * @return
+ *     - ESP_OK                 success.
+ *     - ESP_ERR_FAIL           mDNS is not running or the browsing of `_service._proto` is never started.
+ *     - ESP_ERR_NO_MEM         memory error.
+ */
+esp_err_t mdns_browse_delete(const char *service, const char *proto);
 
 #ifdef __cplusplus
 }
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -18,6 +18,14 @@
 #include "esp_log.h"
 #include "esp_random.h"
 
+static void _mdns_browse_item_free(mdns_browse_t *browse);
+static esp_err_t _mdns_send_browse_action(mdns_action_type_t type, mdns_browse_t *browse);
+static esp_err_t _mdns_sync_browse_action(mdns_action_type_t type, mdns_browse_sync_t *browse_sync);
+static void _mdns_browse_sync(mdns_browse_sync_t *browse_sync);
+static void _mdns_browse_finish(mdns_browse_t *browse);
+static void _mdns_browse_add(mdns_browse_t *browse);
+static void _mdns_browse_send(mdns_browse_t *browse);
+
 #if CONFIG_ETH_ENABLED && CONFIG_MDNS_PREDEF_NETIF_ETH
 #include "esp_eth.h"
 #endif
@@ -57,6 +65,18 @@ static SemaphoreHandle_t _mdns_service_semaphore = NULL;
 
 static void _mdns_search_finish_done(void);
 static mdns_search_once_t *_mdns_search_find_from(mdns_search_once_t *search, mdns_name_t *name, uint16_t type, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol);
+static mdns_browse_t *_mdns_browse_find_from(mdns_browse_t *b, mdns_name_t *name, uint16_t type, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol);
+static void _mdns_browse_result_add_srv(mdns_browse_t *browse, const char *hostname, const char *instance, const char *service, const char *proto,
+                                        uint16_t port, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+static void _mdns_browse_result_add_ip(mdns_browse_t *browse, const char *hostname, esp_ip_addr_t *ip,
+                                       mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+static void _mdns_browse_result_add_txt(mdns_browse_t *browse,  const char *instance, const char *service, const char *proto,
+                                        mdns_txt_item_t *txt, uint8_t *txt_value_len, size_t txt_count, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol,
+                                        uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+#ifdef MDNS_ENABLE_DEBUG
+static void debug_printf_browse_result(mdns_result_t *r_t, mdns_browse_t *b_t);
+static void debug_printf_browse_result_all(mdns_result_t *r_t);
+#endif // MDNS_ENABLE_DEBUG
 static void _mdns_search_result_add_ip(mdns_search_once_t *search, const char *hostname, esp_ip_addr_t *ip,
                                        mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl);
 static void _mdns_search_result_add_srv(mdns_search_once_t *search, const char *hostname, uint16_t port,
@@ -1494,11 +1514,16 @@ static void _mdns_dispatch_tx_packet(mdns_tx_packet_t *p)
 
 #ifdef MDNS_ENABLE_DEBUG
     _mdns_dbg_printf("\nTX[%lu][%lu]: ", (unsigned long)p->tcpip_if, (unsigned long)p->ip_protocol);
+#ifdef CONFIG_LWIP_IPV4
     if (p->dst.type == ESP_IPADDR_TYPE_V4) {
         _mdns_dbg_printf("To: " IPSTR ":%u, ", IP2STR(&p->dst.u_addr.ip4), p->port);
-    } else {
+    }
+#endif
+#ifdef CONFIG_LWIP_IPV6
+    if (p->dst.type == ESP_IPADDR_TYPE_V6) {
         _mdns_dbg_printf("To: " IPV6STR ":%u, ", IPV62STR(p->dst.u_addr.ip6), p->port);
     }
+#endif
     mdns_debug_packet(packet, index);
 #endif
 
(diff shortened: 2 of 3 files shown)


gabsuren (changes requested):

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+    if (!_mdns_server || _str_null_or_empty(service) != _str_null_or_empty(proto)) {
+        return NULL;
+    }
+
+    browse = _mdns_browse_init(service, proto, notifier);
+    if (!browse) {
+        return NULL;
+    }
+
+    if (_mdns_send_browse_action(ACTION_BROWSE_ADD, browse)) {
```
gabsuren, line 7036: There is no forwarding of error code types in case of `_mdns_send_browse_action` fails

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+ * @brief  Free browse structure (except the results)
+ */
+static void _mdns_browse_free(mdns_browse_t *browse)
+{
+    free(browse->service);
+    free(browse->proto);
+    if (browse->result) {
+        mdns_query_results_free(browse->result);
+    }
+    free(browse);
```
gabsuren, line 6986: Potential memory leak, as `browse` is a list

components/mdns/mdns.c
```diff
@@ -4889,6 +4992,14 @@ static void _mdns_free_action(mdns_action_t *action)
     case ACTION_SEARCH_END:
         _mdns_search_free(action->data.search_add.search);
         break;
+    case ACTION_BROWSE_ADD:
```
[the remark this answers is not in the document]
zwx1995esp: One for `_mdns_free_action` and one for `_mdns_execute_action`. I refer to the logic `ACTION_SEARCH_XXX`. Due to the previous logic, the function `_mdns_free_action` should be operated when call the function `mdns_free`, this should be avoid some potential memory leak.

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+    if (!action) {
+        HOOK_MALLOC_FAILED;
+        return ESP_ERR_NO_MEM;
+    }
+
+    action->type = type;
+    action->data.browse_sync.browse_sync = browse_sync;
+    if (xQueueSend(_mdns_server->action_queue, &action, (TickType_t)0) != pdPASS) {
+        free(action);
+        return ESP_ERR_NO_MEM;
```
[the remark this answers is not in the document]
zwx1995esp: Ok, but I refer to all other functions, for example in `_mdns_send_rx_action` when the queue send failed, it return the error `ESP_ERR_NO_MEM`. If this error code is not suitable, what's your suggestion? Should we change all other returns?

components/mdns/mdns.c
zwx1995esp: For this API designed, I also refer to the original implemented function `mdns_query_async_new`, and maybe in this function also no forwarding of error code types when `_mdns_send_search_action`. So for your comments, if we add some error returns, we should change the API declearation and it might be a break change. How about just add a error log here to show the action sent failed?

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+        }
+    }
+
+    browse->notifier = notifier;
+    return browse;
+}
+
+mdns_browse_t *mdns_browse_new(const char *service, const char *proto, mdns_browse_notify_t notifier)
+{
+    mdns_browse_t *browse = NULL;
```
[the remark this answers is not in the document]
zwx1995esp: I add this check at line:7027 but the logic seems wrong, I think `_str_null_or_empty(service) || _str_null_or_empty(proto)` should be ok.

components/mdns/mdns.c
zwx1995esp: This API is designed for only free a list item, not all list, so maybe I need change this API name and add some brefs for it.

zwx1995esp:
> The algorithm for doing this is described in the MDNS specification. Is it being followed? https://datatracker.ietf.org/doc/html/rfc6762#section-5.2
> 
> I just encountered this inability to monitor when trying to build a matter controller. esp-matter needs this feature.

Actually, I did not implement the feature: add the known answer in the query packets. This PR only support to query the target service with no known answers. But that might be an enhancement. You can have a try with this PR :)

zwx1995esp:
> The algorithm for doing this is described in the MDNS specification. Is it being followed? https://datatracker.ietf.org/doc/html/rfc6762#section-5.2
> 
> I just encountered this inability to monitor when trying to build a matter controller. esp-matter needs this feature.

No, I just send the query once and then listen for the service which has been queried.

jonsmirl:
For esp-matter with just need to do a query on something like _matterc._tcp and then monitor add/deletes. This is the event interface.  https://github.com/project-chip/connectedhomeip/blob/master/src/lib/dnssd/platform/DnssdBrowseDelegate.h

chshu:
@jonsmirl This MR introduces a continuous browse feature which is similar to `avahi-browse`, it will send the query once, then monitor the incoming responses, and trigger the callbacks immediately once the service changes.

It was required by Thread BR project, while it should also works for your Matter controller use case. Let us know If there are any functionalities missing.

components/mdns/mdns.c
```diff
@@ -57,6 +65,16 @@ static SemaphoreHandle_t _mdns_service_semaphore = NULL;
...
+static void _mdns_browse_result_add_srv(mdns_browse_t *browse, const char *hostname, const char *instance, const char *service, const char *proto,
+                                        uint16_t port, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+static void _mdns_browse_result_add_ip(mdns_browse_t *browse, const char *hostname, esp_ip_addr_t *ip,
+                                       mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+static void _mdns_browse_result_add_txt(mdns_browse_t *browse,  const char *instance, const char *service, const char *proto,
+                                        mdns_txt_item_t *txt, uint8_t *txt_value_len, size_t txt_count, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol,
+                                        uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+void debug_printf_browse_result(mdns_result_t *r_t, mdns_browse_t *b_t);
+void debug_printf_browse_result_all(mdns_result_t *r_t);
```
david-cermak, line 77:
```suggestion
static void debug_printf_browse_result(mdns_result_t *r_t, mdns_browse_t *b_t);
static void debug_printf_browse_result_all(mdns_result_t *r_t);
```

components/mdns/mdns.c
```diff
@@ -3511,14 +3534,25 @@ void mdns_parse_packet(mdns_rx_packet_t *packet)
     const uint8_t *content = data + MDNS_HEAD_LEN;
     bool do_not_reply = false;
     mdns_search_once_t *search_result = NULL;
+    mdns_browse_t *browse_result = NULL;
+    char browse_result_hostname[MDNS_NAME_BUF_LEN] = {0};
+    char browse_result_instance[MDNS_NAME_BUF_LEN] = {0};
+    char browse_result_service[MDNS_NAME_BUF_LEN] = {0};
+    char browse_result_proto[MDNS_NAME_BUF_LEN] = {0};
```
david-cermak, line 3541: This may take a lot of stack space. Could we allocate dynamically only if we're browsing (or maybe ifdef it out if browsing not enabled)?

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+
+    uint8_t i, j;
+    for (i = 0; i < MDNS_MAX_INTERFACES; i++) {
+        for (j = 0; j < MDNS_IP_PROTOCOL_MAX; j++) {
+            _mdns_search_send_pcb(&search, (mdns_if_t)i, (mdns_ip_protocol_t)j);
+        }
+    }
+}
+
+static esp_err_t _mdns_browse_sync_result_find_or_add_result(mdns_browse_sync_t *sync_browse, mdns_result_t *r)
```
david-cermak, line 7137: could we find a better name? `_mdns_browse_sync_result_find_or_add_result()` doesn't sound very descriptive, maybe something just `add_browse_result()` and comment that if the result's already there we don't have to add it...

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+                }
+                if (!r->addr) {
+                    r->addr = addr_add;
+                    prv = addr_add;
+                } else {
+                    prv->next = addr_add;
+                    prv = addr_add;
+                }
+                r_addr = r_addr->next;
+            }
```
david-cermak, line 7369: this loop looks largely suspicious, but after looking at the code for some time, it seems correct. could we use `copy_addresss_list()` instead?

components/mdns/mdns.c
```diff
@@ -6797,4 +6926,604 @@ void mdns_debug_packet(const uint8_t *data, size_t len)
...
+}
+
+static esp_err_t _mdns_copy_address_in_previous_result(mdns_result_t *result_list, mdns_result_t *r)
+{
+    mdns_ip_addr_t *r_addr;
+    mdns_ip_addr_t *prv = NULL;
+    while (result_list) {
+        if (!_str_null_or_empty(result_list->hostname) && !_str_null_or_empty(r->hostname) && !strcasecmp(result_list->hostname, r->hostname) &&
+                result_list->ip_protocol == r->ip_protocol && result_list->addr) {
+            // If there is a same hostname in previous hostname, we need to copy the address here.
```
david-cermak, line 7353: Question: Why we'd find a hostname in the current result with no address?

david-cermak (commented):
Thanks for posting these changes and implementing browse support. This LGTM! I like the way you treat browsing separately from the searching and other features. It may be helpful to reuse some code from the searching routines, but this change is less invasive and more compact :+1: 

just left few nitpicks and of course we need to test the changes properly.

components/mdns/mdns.c
zwx1995esp: Because one host may pubilsh more than one `_service._protocal` with different instance names, and in the result, we might record them in separated place.

For example, host will publish these mdns service:
```
hostname.test._test._udp
hostname.test1._test._udp
hostname.test2._test._udp
```
When we need to add the `hostname.test1._test._udp` to the result, we need copy the address from the result `hostname.test._test._udp`.

components/mdns/mdns.c
zwx1995esp: Yeah, they are the same logic, I will use the function `copy_address_list` instead.

components/mdns/mdns.c
zwx1995esp: Changed the logic, now using `malloc` to allocate the memory when used.

zwx1995esp:
> just left few nitpicks and of course we need to test the changes properly.

Hi, @david-cermak @gabsuren . I do these tests manually in my locally.

And in my local test, I test for these scenarios manually:
1. Start browsing with the service `_test._udp` on S3, and start/stop publishing these services via linux host using `avahi`
```
(S-A) avahi-publish-service test _test._udp 12347 test=4444 dn="for_ci_br_test1010"
(S-B) avahi-publish-service aaa _test._udp 20617 zwx=esp dn="for_ci_br_test"
```
case-1: Start browsing with the service `_test._udp`, start (S-A), stop (S-A), Stop browsing with the service `_test._udp` (check the heap usage during the test).
case-2: Start browsing with the service `_test._udp`, start (S-A), start (S-B), stop (S-A), stop (S-B), Stop browsing with the service `_test._udp`  (check the heap usage during the test).
case-3: Start browsing with the service `_test._udp`, start (S-A), start (S-B), stop (S-B), stop (S-A), Stop browsing with the service `_test._udp`  (check the heap usage during the test).
case-4: start (S-A), start (S-B), start browsing with the service `_test._udp, stop (S-B), stop (S-A), Stop browsing with the service `_test._udp`  (check the heap usage during the test).
2. Start browsing with the service `_test._udp` and `_testtt._udp`, and start/stop publishing these services via linux host using `avahi`
```
(S-A) avahi-publish-service test _test._udp 12347 test=4444 dn="for_ci_br_test1010"
(S-B) avahi-publish-service aaa _test._udp 20617 zwx=esp dn="for_ci_br_test"

(S-C) avahi-publish-service testxxx _testtt._udp 1234 test=1235 dn="for_ci_br_test"
```
case-1: Start browsing with the service `_test._udp`, Start browsing with the service `_testtt._udp`, start (S-A), stop (S-A), , start (S-C), stop (S-C), Stop browsing with the service `_test._udp`(check the heap usage during the test).
case-2:  Start browsing with the service `_test._udp`, Start browsing with the service `_testtt._udp`, start (S-A), start (S-B), start (S-C), stop (S-A), stop (S-B), stop (S-C), Stop browsing with the service `_test._udp`  (check the heap usage during the test).
3. The API `mdns_browse_new` and  `mdns_browse_delete`
case-1:  Start browsing with the service `_test._udp` for several times, check the heap usage during the test.
case-2: Stop browsing with the service `_test._udp` for several times, check the heap usage during the test.
case-3: Start browsing with the service `_test._udp`, then sop browsing with the service `_test._udp`. Rerun for several times,  check the heap usage during the test.

All the test case seems work fine in my local test env.

components/mdns/mdns.c
david-cermak: Oh, I see, it's actually not a hostname, but kind of a host-namespace of a service.
Thanks for explaining!

david-cermak (approved):
LGTM!

components/mdns/include/mdns.h
```diff
@@ -830,6 +836,28 @@ esp_err_t mdns_unregister_netif(esp_netif_t *esp_netif);
  */
 esp_err_t mdns_netif_action(esp_netif_t *esp_netif, mdns_event_actions_t event_action);
 
+/**
+ * @brief   Browse mDns for a service `_service._proto`.
```
gabsuren, line 840: Nitpick: mDns -> mDNS

gabsuren (approved):

components/mdns/include/mdns.h
zwx1995esp: Oh, thanks for this correction, will fix it asap.

PR #529, opened 2024-03-13
fix(mdns): add lock for some common apis (IDFGH-12338)

zwx1995esp:
This PR fixes a thread-safe risk, some common apis do not take a look when operate the static global variable `_mdns_server` of mdns.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -72,6 +72,7 @@ static bool _mdns_append_host_list(mdns_out_answer_t **destination, bool flush, 
 static void _mdns_remap_self_service_hostname(const char *old_hostname, const char *new_hostname);
 static esp_err_t mdns_post_custom_action_tcpip_if(mdns_if_t mdns_if, mdns_event_actions_t event_action);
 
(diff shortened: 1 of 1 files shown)


chshu (approved):

components/mdns/mdns.c
```diff
@@ -6329,8 +6341,14 @@ esp_err_t mdns_service_remove_all(void)
 /*
  * MDNS QUERY
  * */
-
 void mdns_query_results_free(mdns_result_t *results)
```
zwx1995esp, line 6344: A question: should we consider this API be called in other task? This API might also called in the notifer of mdns search once callback, this should be a mdns context, but the user might now be call this API directly in this callback, so a more safe way is that we double check the task handle, if not a mdns task, we might need to take a lock. What's your idear @david-cermak ?

zwx1995esp:
Hi @david-cermak I wrapped the API `mdns_query_results_free` with taking and giving the mdns service lock. PTAL.

david-cermak (approved):
LGTM! Thanks for the fixes!

Issue #536, opened 2024-03-22
Response Issue in CMUX Mode (IDFGH-12431)


brhmrsln:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

**Environment:**
* Hardware: ESP32-S3
* ESP-IDF
* Module: Quectel MC60 (configured in CMUX mode)
* Connection: UART

**Problem Description:**
We are utilizing the Quectel MC60 module in conjunction with the ESP32-S3, employing the ESP-MODEM component for communication. Our application involves connecting to the internet via the MC60's all-in-one solution and periodically requesting NMEA sentences.

**Issue:**
In CMUX mode, after successfully establishing an internet connection and receiving an IP address, we attempt to fetch GLL NMEA sentences using the command "AT+QGNSSRD" periodically. The MC60 appears to respond correctly with the requested sentence followed by an "OK" acknowledgment. However, In the initial response, we receive only "nmeasentence" without the preceding "OK". Subsequent requests then yield an "OK" response that seems to belong to the previous request, leading to a cycle where each NMEA sentence request is followed by an "OK" from the prior request, effectively causing empty responses and misaligned confirmations.

**Observations:**
The entire response (including "OK") is present in the UART buffer, suggesting that the issue might be with how the response is processed or retrieved.
We employ `esp_modem::dce_commands::generic_get_string` to capture the response.
The issue consistently occurs when the ESP system clock is set to either 240 MHz or 160 MHz.
Interestingly, adding a delay at the beginning of the bool `DTE::command_cb::process_line` function mitigates the issue.
At an 80 MHz ESP system clock frequency, the problem does not occur, and the system behaves as expected, receiving both the NMEA sentence and the "OK" acknowledgment without issues.
Upon inspecting the communication with a logic analyzer on the MC60 side, it is observed that the module transmits the complete response (sentence + "OK") as expected, indicating that the UART transmission from the MC60 is functioning correctly.

**Request:**
[...]

PR #540, opened 2024-03-26
fix(modem): Fixed UART task to check for buffered data periodically

david-cermak:
If we're missing an UART DATA event, or more data than we read are coming from one event, we would never get another event with the buffered data.

## Solution

Checking the data periodically (every 100ms) if no event's coming.

## Closes

https://github.com/espressif/esp-protocols/issues/536

### DIFF
--- a/components/esp_modem/src/esp_modem_uart.cpp
+++ b/components/esp_modem/src/esp_modem_uart.cpp
@@ -160,6 +160,11 @@ void UartTerminal::task()
                 ESP_LOGW(TAG, "unknown uart event type: %d", event.type);
                 break;
             }
+        } else {
+            uart_get_buffered_data_len(uart.port, &len);
+            if (len && on_read) {
+                on_read(nullptr, len);
+            }
         }
     }
 }


components/esp_modem/src/esp_modem_uart.cpp
```diff
@@ -160,6 +160,11 @@ void UartTerminal::task()
                 ESP_LOGW(TAG, "unknown uart event type: %d", event.type);
                 break;
             }
+        } else {
+            uart_get_buffered_data_len(uart.port, &len);
+            if (len && on_read) {
+                on_read(nullptr, len);
```
euripedesrocha, line 166: If I understand correctly, you are calling the read callback with null buffer to consume the data, it wouldn't be better to handle the situation locally?

david-cermak: It's more convenient this way, since the terminal class doesn't have to **own** any buffer. we just leave it to upper layers to use the buffer *they own* for reading. (plus practical consideration: if we handle this locally some bytes may theoretically come after reading locally and could be missed in the actual callback)

euripedesrocha (approved):

PR #543, opened 2024-04-02
fix(wifi_remote): Publish missing esp_wifi_remote and add to README

david-cermak:
Just to publish the same version to component manager. Doesn't need a bump commit or newer version, since it's been tagged and released here with `v.0.1.12`


gabsuren (approved):

david-cermak:
Thanks for the review, Suren!

PR #545, opened 2024-04-05
ci(common): Add clang tidy check to esp-protocols

david-cermak:
## TODO

- [x] ~~Make a reusable workflow and publish here https://github.com/espressif/github-actions~~ -> Created IDF-9900 to track this


david-cermak:
@euripedesrocha @suren-gabrielyan-espressif PTAL, I'm thinking of resolving the *TODO* list above separately (creating a tracker and fixing later), so that we can have some static analysis running for our components.

euripedesrocha (approved):

PR #547, opened 2024-04-09
Reduce maximum wait time for get_operator_name() to fix hangups

0xFEEDC0DE64:
Is there a reason why we have to freeze the task for 75 seconds here?

We have had lots of hangups and it seems like the answer is typically there after ~100ms, not 75000ms


david-cermak:
This command usually takes some time to complete, depending on the modem, but I see that this is more or less configurable on modern devices (e.g. Timeout set with `ATS7`).

Maybe we could add an optional timeout parameter, that would default to those 75seconds?

PR #550, opened 2024-04-12
fix(websocket): Fix locking issues of `esp_websocket_client_send_with_exact_opcode` IDF-9380

gabsuren:
* Fix locking issues of `esp_websocket_client_send_with_exact_opcode` API
* Extended examples to cover more cases
* Added new config CONFIG_ESP_WS_CLIENT_ENABLE_DYNAMIC_BUFFER for testing
* Added new target `release-v5.2`


components/esp_websocket_client/examples/target/pytest_websocket.py
```diff
@@ -150,10 +166,30 @@ def test_recv_long_msg(dut, websocket, msg_len, repeats):
...
+    # Checks if the WebSocket can accurately reconstruct a message sent in several smaller parts.
+    def test_fragmented_txt_msg(dut):
         dut.expect('Received=' + 32 * 'a' + 32 * 'b')
         print('Fragmented data received')
 
+    # Test for receiving fragmented binary messages:
+    # Ensures the WebSocket can handle and reconstruct binary data sent in fragments correctly.
+    def test_fragmented_binary_msg(dut):
+        # Creating the bytearray with specified pattern
+        data = bytearray([0] * 128 + [1] * 128).hex()
```
david-cermak, line 189: the `data` variable seems to be string, not binary if I'm not mistaken?

components/esp_websocket_client/examples/target/pytest_websocket.py
```diff
@@ -26,11 +26,18 @@ def get_my_ip():
     return IP
 
 
-class WebsocketTestEcho(WebSocket):
+def is_binary_strict(data):
+    return all(x in (0, 1) for x in data)
```
david-cermak, line 30: Is this check supposed to test if  the received data is binary? Not sure if it's intended but it seems that it checks if the data contained only `'1'` and `'0'` as characters (so the ascii values of 30 and 31 dec)

david-cermak (approved):
LGTM, just some concerns on testing the binary APIs.

PR #555, opened 2024-04-22
feat(websocket): adding support for `keep_alive_enable` when using WSS transport (IDFGH-11457)

gabsuren:


david-cermak (approved):
LGTM, do we need a test for that?

gabsuren:
> LGTM, do we need a test for that?

TBH I am not sure, as the test should check the connection over a period of time, which can impuct CI time

PR #557, opened 2024-04-22
Adds missing configuration fields

euripedesrocha:
Some aspects of the underlying mqtt client were not set.

### DIFF
--- a/components/esp_mqtt_cxx/esp_mqtt_cxx.cpp
+++ b/components/esp_mqtt_cxx/esp_mqtt_cxx.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -7,12 +7,12 @@
 #include <string>
 #include <algorithm>
 #include <stdexcept>
+#include <variant>
 
 #include "mqtt_client.h"
 #include "esp_log.h"
 
 #include "esp_mqtt.hpp"
-#include "esp_mqtt_client_config.hpp"
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


gabsuren (approved):
LGTM, just a few explanatory lines will be good to have for the functions

euripedesrocha:
> LGTM, just a few explanatory lines will be good to have for the functions

For which functions?

components/esp_mqtt_cxx/esp_mqtt_cxx.cpp
```diff
@@ -133,6 +133,22 @@ esp_mqtt_client_config_t make_config(BrokerConfiguration const &broker, ClientCr
...
+    mqtt_client_cfg.session.protocol_ver = config.session.protocol_ver;
+    mqtt_client_cfg.session.disable_keepalive = config.session.disable_keepalive;
+    mqtt_client_cfg.network.reconnect_timeout_ms = config.connection.reconnect_timeout_ms;
+    mqtt_client_cfg.network.timeout_ms = config.connection.network_timeout_ms;
+    mqtt_client_cfg.network.disable_auto_reconnect = config.connection.disable_auto_reconnect;
+    mqtt_client_cfg.network.refresh_connection_after_ms = config.connection.refresh_connection_after_ms;
+    mqtt_client_cfg.task.priority = config.task.task_prio;
+    mqtt_client_cfg.task.stack_size = config.task.task_stack;
+    mqtt_client_cfg.buffer.size = config.buffer_size;
+    mqtt_client_cfg.buffer.out_size = config.out_buffer_size;
```
david-cermak, line 151: could we use the visitor paradigm as for the broker options?

david-cermak: something like: `config_session(mqtt_client_cfg, config.session);` ?

david-cermak (approved):
LGTM, just a question about different config sections...

components/esp_mqtt_cxx/esp_mqtt_cxx.cpp
euripedesrocha: Not exactly because here we have used the `struct` fields as is in the old mqtt client config struct. I'll do an update of the wrapper as a follow-up, but want to provide the fix first.

PR #559, opened 2024-04-25
feat(mdns): add check of instance when handling PTR query (IDFGH-12701)

gytxxsy:
1. Upon receiving a PTR query, when matching question and answer entries, if an instance exists in the answer entry, it will also be considered as one of the check items. This change is to avoid mistakenly deleting certain question entries. For example, if device A hosts services B with the different instances (ins1~ins5). When the device receives a PTR query, with the question querying service B, and the answer list already contains ins1.Service B and ins2.Service B, then when matching this question entry and answer entries, the question entry should not be deleted.

2. Check the answer entries in the query when creating answer entries for query response. If the answer entry already exists, then give up creating that entry.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1871,6 +1871,7 @@ static void _mdns_create_answer_from_parsed_packet(mdns_parsed_packet_t *parsed_
     packet->id = parsed_packet->id;
 
     mdns_parsed_question_t *q = parsed_packet->questions;
+    uint32_t out_record_nums = 0;
     while (q) {
         shared = q->type == MDNS_TYPE_PTR || q->type == MDNS_TYPE_SDPTR || !parsed_packet->probe;
         if (q->type == MDNS_TYPE_SRV || q->type == MDNS_TYPE_TXT) {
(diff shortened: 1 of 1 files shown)


gytxxsy:
@zwx1995esp @chshu

components/mdns/mdns.c
```diff
@@ -1871,21 +1871,44 @@ static void _mdns_create_answer_from_parsed_packet(mdns_parsed_packet_t *parsed_
...
+                                break;
+                            }
+                        } else if (!service->service->instance && !r->host) {
+                            if (_mdns_service_match(service->service, r->service, r->proto, NULL) && r->ttl > (MDNS_ANSWER_PTR_TTL / 2)) {
+                                is_record_exist = true;
+                                break;
+                            }
+                        }
+                        r = r->next;
+                    }
```
david-cermak, line 1904: Maybe I didn't fully understand the change, but could we tell if the `record_exists` earlier when parsing the packet?

Why do we have to alloc and copy the record entries from parsed answers and compare them later?

david-cermak: I probably understand now, as at the time of parsing you cannot tell if you're about to create answer?

components/mdns/mdns.c
```diff
@@ -3831,6 +3877,44 @@ void mdns_parse_packet(mdns_rx_packet_t *packet)
...
                     }
+                    if (service) {
+                        mdns_parsed_record_t *record = malloc(sizeof(mdns_parsed_record_t));
+                        if (!record) {
+                            HOOK_MALLOC_FAILED;
+                            goto clear_rx_packet;
+                        }
+                        record->next = parsed_packet->records;
+                        parsed_packet->records = record;
+                        record->type = MDNS_TYPE_PTR;
```
david-cermak, line 3888: should we also say that the record type is `MDNS_ANSWER`. Not used anywhere, but for the sake of people reading the code?

components/mdns/mdns.c
gytxxsy: Yes, since the `instance` attribute exists, a single question may correspond to multiple answers. Therefore, when creating a response, it's necessary to check whether there is already a corresponding entry in the answers list of the query. If there is, then no new entry for response should be created (to avoid wasting resources).

It's not reasonable to perform this check during parsing because even if you know that certain entries already exist in the answers of the query during parsing, you still need to create and maintain a `existing entries list`, and compare it with the entries being attempted to be created when creating the response.

gytxxsy: It's more accurate to say that the current `Allocate space and store answer records` is essentially maintaining a `existing entries list`.

components/mdns/mdns.c
gytxxsy: Yes, it's a greate suggestion. Thanks.

components/mdns/mdns.c
david-cermak: > If we want to perform filtering during the parsing, then we should compare the entries existing in the device list with those in the query answer, and then creating and maintaining a list of entries to be created in the response, which will be used when creating the response. However, I think it looks strange and not reasonable.

Oh, okay. I was just curious if we could simplify that and avoid that extra allocation, but you're right that would probably lead to more complications. Thanks for explaining,

david-cermak (approved):
Thanks, LGTM.

you can ignore the clang warnings on `memcpy()` for now -- we'd have to address it globally, later.

gabsuren (approved):

PR #569, opened 2024-05-10
fix(mdns): free txt value len (IDFGH-12817)

DejinChen:
Free the new malloced txt_value_len instead of the txt_value_len in search result. If there exits txt entry in query results, the origin free processing will lead a crash due to multi free.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -4739,7 +4739,7 @@ free_txt:
         free((char *)(txt[i].value));
     }
     free(txt);
-    free(r->txt_value_len);
+    free(txt_value_len);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


chshu:
@gabsuren @david-cermak PTAL. 

https://github.com/espressif/esp-protocols/pull/522 intended to fix a memory leak issue but introduced this issue.

Could you please help to release mdns 1.3.1 version with this fix? It's a critial bug found in both the Matter and Thread BR applications.

gabsuren (approved):

gabsuren:
@chshu release `mdns-v1.3.1` is already available 
https://components.espressif.com/components/espressif/mdns

PR #573, opened 2024-05-15
feat(websocket): allow using external TCP transport handle (IDFGH-12825)

huming2207:
Hi there,

This PR was originally done last year (May 2023), but I forgot to submit it... 😅 

Anyway, this PR is for adding support to let WebSocket client to take external handles, similar to my another PR for MQTT client, see:

- https://github.com/espressif/esp-mqtt/blob/master/include/mqtt_client.h#L339
- https://github.com/espressif/esp-mqtt/pull/260

This is very useful for scenarios where: 

1. The application requires proxies to access network (via my [esp_tcp_transport_addons](https://github.com/huming2207/esp_tcp_transport_addons) hack for HTTP proxy access, or a SOCK4 proxy via the `transport_socks_proxy.c` in `tcp_transport` component) ;
2. The application somehow requires external TCP/IP stack on a modem via serial port, see [this example](https://github.com/espressif/esp-protocols/tree/c9439bd3d57f4941c980476f9961ff04687a9f7a/components/esp_modem/examples/modem_tcp_client)

Currently for my company, we are using my HTTP proxy library mentioned above to achieve HTTP proxy access feature. But we are currently using my modified `esp-websocket-client` and we wish to merge this into mainline soon instead. 

Please provide suggestion if there's any and I will modify accordingly. Thanks.

Regards,
Jackson

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -96,6 +96,7 @@ typedef struct {
     bool                        use_global_ca_store;
     bool                        skip_cert_common_name_check;
     esp_err_t                   (*crt_bundle_attach)(void *conf);
+    esp_transport_handle_t      ext_transport;
 } websocket_config_storage_t;
 
 typedef enum {
@@ -677,6 +678,7 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
     client->config->client_key_len = config->client_key_len;
     client->config->skip_cert_common_name_check = config->skip_cert_common_name_check;
     client->config->crt_bundle_attach = config->crt_bundle_attach;
+    client->config->ext_transport = config->ext_transport;
 
     if (config->uri) {
         if (esp_websocket_client_set_uri(client, config->uri) != ESP_OK) {
@@ -1100,9 +1102,13 @@ esp_err_t esp_websocket_client_start(esp_websocket_client_handle_t client)
         ESP_LOGE(TAG, "The client has started");
         return ESP_FAIL;
     }
-    if (esp_websocket_client_create_transport(client) != ESP_OK) {
-        ESP_LOGE(TAG, "Failed to create websocket transport");
-        return ESP_FAIL;
+
+    client->transport = client->config->ext_transport;
+    if (!client->transport) {
+        if (esp_websocket_client_create_transport(client) != ESP_OK) {
+            ESP_LOGE(TAG, "Failed to create websocket transport");
+            return ESP_FAIL;
+        }
     }
 
     if (xTaskCreate(esp_websocket_client_task, client->config->task_name ? client->config->task_name : "websocket_task",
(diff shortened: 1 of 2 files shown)


huming2207:
cc @suren-gabrielyan-espressif and @david-cermak

huming2207:
Any updates on this...?

huming2207:
any updates...? @gabsuren

gabsuren:
@huming2207 Thank you for your contribution, and we apologize for the delayed response. We are looking into it and will update you soon.

huming2207:
@gabsuren any updates? it's been a month...

euripedesrocha:
@huming2207 this is in our pipeline for evaluation as Suren mentioned. 
I looked into the code changes and in the current state we'll need some changes before accepting it. 
You are passing only the parent transport. We should have the actual transport that will be used since it is more flexible. Some users might want to have their own implementation of the websocket transport level and if only the parent customization is needed it is easy to compose before sending to websocket client. 

Also we need you to install the pre-commit hook as explained in our [contributing guide](https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md#submitting-a-pr) and fix the issues.

huming2207:
Hi @gabsuren 

> You are passing only the parent transport. We should have the actual transport that will be used since it is more flexible.

Just to clarify, it seems like `esp-mqtt` did the same like I did, except it calls `esp_transport_destroy()` at deinit. Do you mean I should add the `esp_transport_destroy()`? Or something else?

 > Also we need you to install the pre-commit hook as explained in our [contributing guide](https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md#submitting-a-pr) and fix the issues.

Sure, I will do that.

Regards,
Jackson

huming2207:
Also, just to be clear, this PR is solely for adding **parent** transport support, so that we are able to connect our stuff over a HTTP proxy provided by our partner companies. We don't need to, and better off not to touch the WS transport itself. Because that's not our use case at all.

euripedesrocha:
> Just to clarify, it seems like esp-mqtt did the same like I did, except it calls esp_transport_destroy() at deinit. Do you mean > I should add the esp_transport_destroy()? Or something else?

Not exactly the same way. MQTT client uses the external transport as is, not the base parent, as your proposal. In MQTT client, we don't even construct the `transport_list`

About the cleanup, we want to keep the transport valid as long as the client is in use and simplify user task on tracking the lifetime of the transport, but it is not what I'm suggesting here. We might even go back on that decision for mqtt client. 

> Also, just to be clear, this PR is solely for adding parent transport support, so that we are able to connect our stuff over a HTTP proxy provided by our partner companies. We don't need to, and better off not to touch the WS transport itself. Because that's not our use case at all.

I understand that this is your use case, and to be honest, probably the most common use case but, we need to provide the feature for the general case. And given that the workaround of constructing the websocket transport is just 1 or 2 extra lines it is better to do it that way.

huming2207:
> Not exactly the same way. MQTT client uses the external transport as is, not the base parent, as your proposal. In MQTT client, we don't even construct the `transport_list`
> 
> About the cleanup, we want to keep the transport valid as long as the client is in use and simplify user task on tracking the lifetime of the transport, but it is not what I'm suggesting here. We might even go back on that decision for mqtt client.

Ah crap, yes I misread the code. 

I will modify the code accordingly,

huming2207:
Hi @gabsuren @euripedesrocha 

I've committed the changes. Please re-review. Thanks.

gabsuren (approved):

euripedesrocha (approved):

huming2207:
Thanks for the approval, but it seems like the CI is failing because of lacking heaeders in the logging library? Should I provide a workaround or wait till you fix it?

PR #574, opened 2024-05-20
fix(mdns): remove same protocol services with different instances

DejinChen:
### Problem
ESP-Matter may publish services with the same protocol and service type but different instance names. e.g., ''a.matter._tcp" and "b.matter._tcp". When service info changed, we have to remove all "matter._tcp" service through `mdns_service_remove(const char *service_type, const char *proto)` and re-add new service. But now, the the different instance name services can not be removed at the same time. And re-adding will lead a **memory leak** problem for the unremoved services with the same subtype.

### Change overview:
1. Modify remove services action to remove the same instance name services. 
2. Verify that the service already has the same subtype before adding.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -5200,28 +5200,57 @@ static void _mdns_execute_action(mdns_action_t *action)
         break;
     case ACTION_SERVICE_DEL:
         a = _mdns_server->services;
-        if (action->data.srv_del.service) {
-            if (_mdns_server->services == action->data.srv_del.service) {
-                _mdns_server->services = a->next;
-                _mdns_send_bye(&a, 1, false);
-                _mdns_remove_scheduled_service_packets(a->service);
-                _mdns_free_service(a->service);
-                free(a);
-            } else {
-                while (a->next && a->next != action->data.srv_del.service) {
-                    a = a->next;
+        mdns_srv_item_t *b = a;
+        if (action->data.srv_del.instance) {
+            while (a) {
+                if (_mdns_service_match_instance(a->service, action->data.srv_del.instance,
+                                                 action->data.srv_del.service, action->data.srv_del.proto,
+                                                 action->data.srv_del.hostname)) {
+                    if (_mdns_server->services != a) {
+                        b->next = a->next;
+                    } else {
+                        _mdns_server->services = a->next;
+                    }
+                    _mdns_send_bye(&a, 1, false);
+                    _mdns_remove_scheduled_service_packets(a->service);
+                    _mdns_free_service(a->service);
+                    free(a);
+                    break;
                 }
-                if (a->next == action->data.srv_del.service) {
-                    mdns_srv_item_t *b = a->next;
-                    a->next = a->next->next;
-                    _mdns_send_bye(&b, 1, false);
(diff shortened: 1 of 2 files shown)


components/mdns/mdns.c
```diff
@@ -6401,6 +6430,16 @@ esp_err_t mdns_service_subtype_add_for_host(const char *instance_name, const cha
     if (!s) {
         return ESP_ERR_NOT_FOUND;
     }
+
+    mdns_subtype_t *srv_subtype = s->service->subtype;
+    while (srv_subtype) {
+        if (strcmp(srv_subtype->subtype, subtype) == 0) {
+            // The same subtype has already been added
+            return ESP_OK;
```
david-cermak, line 6438: nitpick: Could we return something else if the subtype already exits? Maybe `ESP_ERR_INVALID_ARG` as other mdns APIs?

components/mdns/mdns.c
```diff
@@ -5200,28 +5200,57 @@ static void _mdns_execute_action(mdns_action_t *action)
...
-                _mdns_server->services = a->next;
-                _mdns_send_bye(&a, 1, false);
-                _mdns_remove_scheduled_service_packets(a->service);
-                _mdns_free_service(a->service);
-                free(a);
-            } else {
-                while (a->next && a->next != action->data.srv_del.service) {
-                    a = a->next;
+        mdns_srv_item_t *b = a;
+        if (action->data.srv_del.instance) {
```
david-cermak, line 5204: Isn't this a breaking change? The API `mdns_service_remove_for_host()` simply removed the service in the past, whether `instance` was set or not, correct?

components/mdns/mdns.c
DejinChen: done

components/mdns/mdns.c
DejinChen: When `instance` is not set, the service will be removed if matched by `_mdns_service_match()`. And if set, matched by `_mdns_service_match_instance`.

david-cermak: That's not what I meant. The changes in this PR make perfect sense.

My only concern is about backward compatibility, so if someone removed services with *some* instance set, it just removed everything. Now it will remove only that single instance, correct?

But this is probably an edge usecase and we can treat this is a bugfix on API level.
Anyway, we have quite a big suite of test cases, so I think we should check if there's a case like that...

components/mdns/mdns.c
david-cermak: thanks!

components/mdns/mdns.c
DejinChen: Yes, this change may result in a different removing result than before...
So should i also add a test case for this?

david-cermak: > So should i also add a test case for this?

No need. The test cases I was talking about are not part of this repo.

I think we can just merge it and ask our Q/A team to test it.

DejinChen: OK.

david-cermak (approved):
LGTM, just please fix the CI.

@gabsuren can help you with the AFL host test, probably something missing on the mocking layer.

PR #578, opened 2024-05-27
[eppp]: Make some common esp_netif options configurable

david-cermak:
## Adding these netif options to eppp-config to make them configurable
- netif description -- used in common examples
- netif route_prio -- in case we need more netifs with different default GW

# Changelog

### Bug Fixes

- Make some common netif options configurable ([a5582ad](https://github.com/espressif/esp-protocols/commit/a5582ad))

### DIFF
--- a/components/eppp_link/.cz.yaml
+++ b/components/eppp_link/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(eppp): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py eppp_link
   tag_format: eppp-v$version
-  version: 0.1.0
+  version: 0.1.1
   version_files:
   - idf_component.yml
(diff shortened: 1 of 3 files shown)


components/eppp_link/eppp_link.c
```diff
@@ -703,7 +706,7 @@ void eppp_deinit(esp_netif_t *netif)
 
 esp_netif_t *eppp_init(eppp_type_t role, eppp_config_t *config)
 {
-    esp_netif_t *netif = netif_init(role);
+    esp_netif_t *netif = netif_init(role, config);
```
euripedesrocha, line 709: Shall we verify if `role` and `config` are not `NULL`?

david-cermak: Have added null checks to the public APIs:
* `eppp_init()`
* `eppp_open()`

PR #582, opened 2024-05-30
[wifi-remote]: Fix server race when receiving command and posting events

david-cermak:
Fixes race condition of simultaneously marshaling WiFi/IP event and receiving command from client.
* 9e13870 fix(wifi_remote): Lock server before marshaling events --> Quick and simple fix, but may block default event task
* 30a5262 Fix server event/command race condition using eventfd --> More elaborate and correct fix

### DIFF
--- a/components/esp_wifi_remote/eppp/wifi_remote_rpc_server.cpp
+++ b/components/esp_wifi_remote/eppp/wifi_remote_rpc_server.cpp
@@ -15,6 +15,8 @@
 #include "eppp_link.h"
 #include "wifi_remote_rpc_params.h"
 #include "lwip/apps/snmp.h"
+#include "esp_vfs.h"
+#include "esp_vfs_eventfd.h"
 
 extern "C" esp_netif_t *wifi_remote_eppp_init(eppp_type_t role);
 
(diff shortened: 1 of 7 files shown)


components/esp_wifi_remote/eppp/wifi_remote_rpc_server.cpp
```diff
@@ -32,7 +34,74 @@ const unsigned char key[] = "-----BEGIN PRIVATE KEY-----\n" CONFIG_ESP_WIFI_REMO
...
+    api_id type;
+    int32_t id;
+    esp_wifi_remote_eppp_ip_event *ip_data{nullptr};
+    bool clean_ip_data{true};
+    esp_err_t create_ip_data()
+    {
+        ip_data = new (std::nothrow) esp_wifi_remote_eppp_ip_event;
+        return ip_data ? ESP_OK : ESP_ERR_NO_MEM;
+    }
+    ~Events()
```
euripedesrocha, line 47: Could you explain the usage you expect here? The `struct` has only default constructed data and depending on state the destructor might leak the allocated `ip_data`.

david-cermak: The idea is using RAII even when en-queuing and later de-queuing the instance.

> the destructor might leak the allocated ip_data.

* we construct `ip_data` only when needed (if we get the appropriate `IP_EVENT`)
* we put this instance to the queue and only after (successful) en-queuing we mark this as `DO-NOT-FREE`, so the desctructor *leaks* the data, as they will be used afterwards...
* we get the instance from the queue with the default state, so now the destructor *free*s the data as expected

(briefly commented here https://github.com/espressif/esp-protocols/pull/582/files#diff-e2d9eebe2ae3236f3c8b216ed4f69dd878089e5f7c1b2fe3fb9a23c238651b8fR61)

euripedesrocha (approved):

PR #587, opened 2024-06-04
fix(mdns): Fix of mdns afl tests

gabsuren:


gabsuren:
@david-cermak thank you for a quick review

PR #604, opened 2024-06-24
Allow to add Delay to CMUX DISC (IDFGH-15937)

thiker5:
Some Modems require a delay before sending disconnect, added OPTION ESP_MODEM_CMUX_DELAY_AFTER_DLCI_DISCONNECT to allow configuring that


thiker5:
Some Modems require a delay before sending disconnect, added OPTION ESP_MODEM_CMUX_DELAY_AFTER_DLCI_DISCONNECT to allow configuring that

PR #606, opened 2024-06-26
[mosquitto]: Simple WIP support for localhost mosquitto

david-cermak:
:warning: **very WIP**


components/mosq/examples/broker/main/broker.c
```diff
@@ -0,0 +1,110 @@
...
+        break;
+    case MQTT_EVENT_UNSUBSCRIBED:
+        ESP_LOGI(TAG, "MQTT_EVENT_UNSUBSCRIBED, msg_id=%d", event->msg_id);
+        break;
+    case MQTT_EVENT_PUBLISHED:
+        ESP_LOGI(TAG, "MQTT_EVENT_PUBLISHED, msg_id=%d", event->msg_id);
+        break;
+    case MQTT_EVENT_DATA:
+        ESP_LOGI(TAG, "MQTT_EVENT_DATA");
+        printf("TOPIC=%.*s\r\n", event->topic_len, event->topic);
```
gabsuren, line 56: printf -> ESP_LOGI ?

gabsuren:
I see CI `build-target` tests, what about `run-target` tests?

components/mosq/port/callbacks.c
```diff
@@ -0,0 +1,48 @@
...
+#include "mosquitto_broker.h"
+#include "memory_mosq.h"
+#include "mqtt_protocol.h"
+#include "send_mosq.h"
+#include "util_mosq.h"
+#include "utlist.h"
+#include "lib_load.h"
+
+
+int mosquitto_callback_register(
```
euripedesrocha, line 18: Will we make it more flexible to use the callbacks in the future?

components/mosq/port/config.c
```diff
@@ -0,0 +1,244 @@
...
+    mosquitto__free(config->persistence_file);
+    config->persistence_file = NULL;
+    config->persistent_client_expiration = 0;
+    config->queue_qos0_messages = false;
+    config->retain_available = true;
+    config->set_tcp_nodelay = false;
+    config->sys_interval = 10;
+    config->upgrade_outgoing_qos = false;
+
+//    config__cleanup_plugins(config);
```
kostaond, line 97: Cleanup

components/mosq/examples/broker/README.md
```diff
@@ -0,0 +1,48 @@
+# MQTT broker example
+
+## Overview
+
+This example runs a TCP broker on a specified host and port.
+
+### How to use this example
+
+Choose the preferred connection in the project configuration menu, build and run the example.
+After it connects vias the specified network interface, it start the mosquitto broker, which by default listens on the connected interface.
```
alisitsyn, line 10: A small typo "vias"
"After it connects via the specified network interface, it starts the mosquitto broker, which by default listens on the connected interface." ?

components/mosq/examples/broker/main/broker.c
david-cermak: will use esp logs

david-cermak:
> I see CI `build-target` tests, what about `run-target` tests?

I will add a target test before first release

components/mosq/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+# Changelog
+
+## [2.0.28~0](https://github.com/espressif/esp-protocols/commits/mosq-v2.0.28_0)
+
+### Features
+
+- Simple WIP support for localhost mosquitto ([2e7233d8](https://github.com/espressif/esp-protocols/commit/2e7233d8))
```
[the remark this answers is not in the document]
david-cermak: I'll update it. it already supports remote clients

components/mosq/port/callbacks.c
david-cermak: we can extend it in future, but this version doesn't support plugins

components/mosq/port/config.c
```diff
@@ -0,0 +1,244 @@
...
+    }
+    if (config->log_file) {
+        mosquitto__free(config->log_file);
+        config->log_file = NULL;
+    }
+}
+
+const char *mosquitto_client_username(const struct mosquitto *client)
+{
+    return "my_client";
```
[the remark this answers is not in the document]
david-cermak: I'll double check if it's not used in  the broker impl,
If so I'd supply `return NULL; ` and comment that it's not used

components/mosq/port/config.c
```diff
@@ -0,0 +1,244 @@
...
+    while (endptr > str && isspace((int)endptr[0])) {
+        endptr[0] = '\0';
+        endptr--;
+    }
+    return str;
+}
+
+// Dummy definition of fork() to work around IDF warning: " warning: _fork is not implemented and will always fail"
+// fork() is used in mosquitto.c to deamonize the broker, which we do not call.
+pid_t fork (void)
```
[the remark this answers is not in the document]
david-cermak: I will add `abort()`

components/mosq/port/config.c
david-cermak: will do

components/mosq/examples/broker/README.md
david-cermak: will fix

david-cermak:
1) Merge this to master first before creating a release
2) Component name `mosquitto` 
3) Test locally scenario with 5 clients publishing each second and running over weekend (on esp32)
 - document the usecase (memory cunsumption)
4) Add target test
5) Mention what is supported and what not

components/mosq/port/config.c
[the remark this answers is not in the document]
david-cermak: This is actually used in the security layers for client identification.
Not used in this version, but fix this to return client id:
```cpp
const char *mosquitto_client_username(const struct mosquitto *client)
{
    return client->id;
}
```

components/mosq/port/config.c
david-cermak: fixed

components/mosq/examples/broker/README.md
david-cermak: fixed and updated the README. Thanks!

components/mosq/port/callbacks.c
david-cermak: I will not support plugins in the near future, but I'd like to implement callbacks in user configs, which would be useful to monitor whenever 
* client connects/disconnects
* provide a custom means to reject connection (based on IP, client id, user, password....)
* message receive -- with client id, topic, payload.

(experimented with that in https://github.com/david-cermak/esp-protocols/commit/d24019a133c959d56a02e2196fc20fa9c5b81957)

but that's still won't happen in the first published version

david-cermak:
> > I see CI `build-target` tests, what about `run-target` tests?

Target tests have been added for designated IDF branches and they pass: https://github.com/espressif/esp-protocols/actions/runs/10846059208/job/30098329113?pr=606#step:4:160

david-cermak:
@euripedesrocha @alisitsyn @kostaond @espressif-abhikroy Thanks for your review. This PR has been updated and target tests added. Could you please take another look and approve if okay with you?

components/mosquitto/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Supported Options
+
+The Espressif port supports a limited set of options (with plans to add more in future releases). These options can be configured through a structure passed to the `mosq_broker_start()` function. For detailed information on available configuration options, refer to the [API documentation](api.md).
+
+## API
+
+### Starting the Broker
+
+To start the broker, call the `mosq_broker_start()` function with a properly configured settings structure. The broker operates in the context of the calling task and does not create a separate task.
```
euripedesrocha, line 13: Why not create a separate task to run? This would make the usage inline with other components we have.

components/mosquitto/examples/broker/main/broker.c
```diff
@@ -0,0 +1,92 @@
...
+{
+    ESP_ERROR_CHECK(nvs_flash_init());
+    ESP_ERROR_CHECK(esp_netif_init());
+    ESP_ERROR_CHECK(esp_event_loop_create_default());
+    ESP_ERROR_CHECK(example_connect());
+
+    struct mosq_broker_config config = { .host = CONFIG_EXAMPLE_BROKER_HOST, .port = CONFIG_EXAMPLE_BROKER_PORT };
+
+#if CONFIG_EXAMPLE_BROKER_RUN_LOCAL_MQTT_CLIENT
+    xTaskCreate(mqtt_app_start, "mqtt_client", 4096, &config, 4, NULL);
```
euripedesrocha, line 88: Instead of spawning a task, have you considered adding the task delay on the before connect event handler?

euripedesrocha (approved):

components/mosquitto/README.md
david-cermak: This can be added later, if needs be.

I'd like to include this "task-less" option, because it's simply more flexible. Otherwise we'd start the task in the lib, but had to expose 
* task memory allocation capabilities
* task affinity
* priority / stack size
* ...something else?

components/mosquitto/examples/broker/main/broker.c
david-cermak: Good point, thanks! Removed the extra task and using before connect event to stall the mqtt task,

components/mosquitto/README.md
```diff
@@ -1,3 +1,32 @@
...
+
+It's recommended to analyze the stack size needed for the task, but in general, the broker requires at least 4 kB of stack size.
```
alisitsyn, line 15: Is this possible to configure the stack size through the kconfig value?

david-cermak: The broker runs in the calling task, users can easily adjust that programatically. Related discussion: https://github.com/espressif/esp-protocols/pull/606#discussion_r1761092636

alisitsyn (approved):

components/mosquitto/README.md
david-cermak: I'll address this in a followup PR:
* rename this API to `mqtt_broker_run()` to run the broker within current task
* use `mosq_broker_start()` `mosq_broker_stop()` to start/stop a task with the broker.

PR #607, opened 2024-07-03
feat(websocket): interruptable wait timeout

erkia:
Currently, websocket automatic reconnect delay is implemented using simple `vTaskDelay(client->wait_timeout_ms / 2 / portTICK_PERIOD_MS)`. This has two drawbacks:

 - changing wait timeout while delay is already active has an unpredictable result
 - calling `esp_websocket_client_stop()` while delay is active will block until (half) reconnect delay is passed, which is especially annoying when delay is long.

This pull requests uses task notifications to wake up websocket task from reconnect delay when necessary.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -471,6 +471,9 @@ static esp_err_t stop_wait_task(esp_websocket_client_handle_t client)
     }
 
     client->run = false;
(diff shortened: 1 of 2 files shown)


gabsuren:
Addressed in https://github.com/espressif/esp-protocols/pull/1085

gabsuren:
@erkia thank you for your contribution!

PR #608, opened 2024-07-05
feat(websocket): adding support for `if_name` when using WSS transport

johanstokking:
This is a follow up on https://github.com/espressif/esp-protocols/pull/555; `if_name` as not set for WSS while it is for WS. This fixes that. Thanks for considering.


gabsuren (approved):

gabsuren:
@johanstokking thank you for your contribution. LGTM

PR #609, opened 2024-07-08
[examples]: Make multi-netif example working with DNS_PER_DEFAULT_NETIF

david-cermak:
Latest IDF supports `CONFIG_ESP_NETIF_SET_DNS_PER_DEFAULT_NETIF` config option which restores global DNS server information from "per-netif" records whenever a new netif is selected for routing (set as default).

### DIFF
--- a/examples/esp_netif/multiple_netifs/README.md
+++ b/examples/esp_netif/multiple_netifs/README.md
@@ -19,6 +19,14 @@ This example demonstrates working with multiple different interfaces with differ
   * It tries to reconfigure DNS server if host name resolution fails
   * It tries to manually change the default interface if connection fails
 
+### Handling DNS server across interfaces
+
+This example also demonstrates how DNS servers could be handled on network interface level, as lwIP used global DNS server information.
+
+All network interfaces store their DNS info upon acquiring an IP in the internal structure (in the application code) and the DNS servers are restored if host name resolution fails.
+
+This functionality is handled in IDF (supported from v5.3) automatically if `CONFIG_ESP_NETIF_SET_DNS_PER_DEFAULT_NETIF` is enabled, the DNS server info keeps updating per network interface in IDF layers. This examples uses the IDF functionality if `CONFIG_ESP_NETIF_SET_DNS_PER_DEFAULT_NETIF=1`.
+
 ### Hardware Required
 
 To run this example, it's recommended that you have an official ESP32 Ethernet development board - [ESP32-Ethernet-Kit](https://docs.espressif.com/projects/esp-idf/en/latest/hw-reference/get-started-ethernet-kit.html).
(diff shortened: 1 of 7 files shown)


examples/esp_netif/multiple_netifs/main/multi_netif_main.c
```diff
@@ -69,13 +69,15 @@ void app_main(void)
...
-        wifi_init(WIFI_PRIO),
-        ppp_init(PPP_PRIO),
+        example_eth_init(ETH_PRIO),
+        example_wifi_init(WIFI_PRIO),
+        example_ppp_init(PPP_PRIO),
     };
     size_t num_of_ifaces = sizeof(ifaces) / sizeof(ifaces[0]);
 
     while (true) {
+        // For demonstration purposes we clear DNS table every iteration to exercise
```
euripedesrocha, line 79: Could we try to move this out of a loop and have it set twice only? One with the regular expected way users would do and after that we clean and do as demo? We could consider keep the loop as is as long as we have once the expected way clearly stated.

examples/esp_netif/multiple_netifs/README.md
```diff
@@ -19,6 +19,9 @@ This example demonstrates working with multiple different interfaces with differ
   * It tries to reconfigure DNS server if host name resolution fails
   * It tries to manually change the default interface if connection fails
 
+This example demonstrate how DNS servers could be handled on network interface level, as lwIP used global DNS server information. All network interfaces store the DNS info upon acquiring an IP in the internal structure and the DNS servers are restored if host name resolution fails.
```
euripedesrocha, line 22: It might be worth to add a subsection for the DNS part to improve reading.

examples/esp_netif/multiple_netifs/sdkconfig.defaults
```diff
@@ -1,2 +1,5 @@
+# You can use CONFIG_ESP_NETIF_SET_DNS_PER_DEFAULT_NETIF
```
euripedesrocha, line 1: Do we expect this to be the default for users? If so, we should enable it here and point in README how to disable.

examples/esp_netif/multiple_netifs/main/multi_netif_main.c
david-cermak: I didn't want to complicate it too much.
But I see your point, I've added a config option and explained the purpose. This is useful for users (even if they just copy-paste the code) to enable it for testing and disable for release.

examples/esp_netif/multiple_netifs/sdkconfig.defaults
david-cermak: No, we can't. It's not supported on all IDF versions.

examples/esp_netif/multiple_netifs/README.md
david-cermak: Reworded and extended (made a section out of this paragraph) PTAL again

euripedesrocha (approved):

PR #612, opened 2024-07-10
[modem]: Add support for guessing mode

david-cermak:
* Supports only DATA and PPP mode

* [x] Add CMUX modes

### DIFF
--- a/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
@@ -66,6 +66,18 @@ public:
     int write(DTE_Command command);
 
     /**
+     * @brief send data to the selected terminal, by default (without term_id argument)
+     * this API works the same as write: sends data to the secondary terminal, which is
+     * typically used as data terminal (for networking).
+     *
+     * @param data Data pointer to write
+     * @param len Data len to write
+     * @param term_id Terminal id: Primary if id==0, Secondary if id==1
+     * @return number of bytes written
+     */
+    int send(uint8_t *data, size_t len, int term_id = 1);
+
+    /**
      * @brief Reading from the underlying terminal
      * @param d Returning the data pointer of the received payload
      * @param len Length of the data payload
--- a/components/esp_modem/src/esp_modem_dce.cpp
+++ b/components/esp_modem/src/esp_modem_dce.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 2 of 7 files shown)


components/esp_modem/src/esp_modem_dce.cpp
```diff
@@ -191,4 +262,114 @@ modem_mode DCE_Mode::get()
...
+
+/**
+ * This namespace contains probe packets and expected replies on 3 different protocols,
+ * the modem device could use (as well as timeouts and mode ids for synchronisation)
+ */
+namespace probe {
+
+namespace ppp {
+// Test that we're in the PPP mode by sending an LCP protocol echo request and expecting LCP echo reply
+const uint8_t lcp_echo_request[] = {0x7e, 0xff, 0x03, 0xc0, 0x21, 0x09, 0x01, 0x00, 0x08, 0x99, 0xd1, 0x35, 0xc1, 0x8e, 0x2c, 0x7e };
```
euripedesrocha, line 279: Nitpick: This could be a constexpr std::array.

components/esp_modem/src/esp_modem_dce.cpp
```diff
@@ -191,4 +262,114 @@ modem_mode DCE_Mode::get()
...
+const uint8_t sabm0_reply[] = {0xf9, 0x03, 0x73, 0x01};
+const int mode = 1 << 0;
+const int timeout = 200;
+}
+};
+
+modem_mode DCE_Mode::guess_unsafe(DTE *dte, bool with_cmux)
+{
+    // placeholder for reply and its size, since it could come in pieces, and we have to cache
+    // this is passed to the lambda as references.
```
euripedesrocha, line 306:
```suggestion
    // this is captured by the lambda by reference.
```

components/esp_modem/src/esp_modem_dce.cpp
```diff
@@ -191,4 +262,114 @@ modem_mode DCE_Mode::get()
...
+        }
+        auto reply_size = std::min((size_t)sizeof(reply) - reply_pos, len);
+        ::memcpy(reply + reply_pos, data, reply_size);
+        reply_pos += reply_size;
+        ESP_LOG_BUFFER_HEXDUMP("esp-modem: guess mode data:", reply, reply_pos, ESP_LOG_DEBUG);
+
+        // Check whether the response resembles the "golden" reply (for these 3 protocols)
+        if (reply_pos >= sizeof(probe::ppp::lcp_echo_reply_head)) {
+            // check for initial 2 bytes
+            uint8_t *ptr = static_cast<uint8_t *>(memmem(reply, reply_pos, probe::ppp::lcp_echo_reply_head, 2));
```
euripedesrocha, line 325: Build a `string_view` from the received buffer and use provided algorithms might lead to simpler code here.

components/esp_modem/include/cxx_include/esp_modem_dte.hpp
```diff
@@ -65,6 +65,18 @@ class DTE : public CommandableIf {
...
     int write(DTE_Command command);
 
+    /**
+     * @brief send data to the selected terminal, by default (without term_id argument)
+     * this API works the same as write: sends data to the secondary terminal, which is
+     * typically used as data terminal (for networking).
+     *
+     * @param data Data pointer to write
+     * @param len Data len to write
+     * @param term_id Terminal id: Primary if id==0, Secondary if id==2
```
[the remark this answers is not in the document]
david-cermak: This was wrong, fixed in the doxy comments. the secondary terminal is id==1

components/esp_modem/src/esp_modem_dce.cpp
david-cermak: used std::array

components/esp_modem/src/esp_modem_dce.cpp
david-cermak: any idea how to do it without reinterpret cast?
```cpp
    std::string_view reply_view(reinterpret_cast<const char*>(reply), reply_size);
    
    // Create a string_view for the bytes to compare against
    std::string_view head_view(reinterpret_cast<const char*>(ppp::lcp_echo_reply_head.data()), 2);

    // Check if the reply starts with the first 2 bytes of lcp_echo_reply_head
    return reply_view.starts_with(head_view);
```

david-cermak: The thing is that I'm actually testing if the current protocol is binary or ascii, so I think the best option is to just use the old good memmem()...

euripedesrocha (approved):

PR #613, opened 2024-07-12
feat(mdns): check if the txt items is changed when browsing

gytxxsy:
When browsing, it is necessary to check for new TXT records. If the TXT records are changed, an update operation should be performed.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -7521,6 +7521,21 @@ static mdns_browse_t *_mdns_browse_find_from(mdns_browse_t *b, mdns_name_t *name
     return b;
 }
 
+static bool is_txt_item_in_list(mdns_txt_item_t txt, uint8_t txt_value_len, mdns_txt_item_t *txt_list, uint8_t *txt_value_len_list, size_t txt_count)
+{
+    for (size_t i = 0; i < txt_count; i++) {
+        if (strcmp(txt.key, txt_list[i].key) == 0) {
+            if (txt_value_len == txt_value_len_list[i] && memcmp(txt.value, txt_list[i].value, txt_value_len) == 0) {
+                return true;
+            } else {
+                // The key value is unique, so there is no need to continue searching.
+                return false;
+            }
+        }
+    }
+    return false;
+}
+
 /**
  * @brief  Called from parser to add TXT data to search result
  */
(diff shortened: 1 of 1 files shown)


components/mdns/mdns.c
```diff
@@ -7521,6 +7521,21 @@ static mdns_browse_t *_mdns_browse_find_from(mdns_browse_t *b, mdns_name_t *name
     return b;
 }
 
+static bool is_txt_item_exist(mdns_txt_item_t txt, uint8_t txt_value_len, mdns_txt_item_t *txt_list, uint8_t *txt_value_len_list, size_t txt_count)
```
david-cermak, line 7524: nitpick, suggest we give it a better name...
```suggestion
static bool is_txt_item_in_list(mdns_txt_item_t txt, uint8_t txt_value_len, mdns_txt_item_t *txt_list, uint8_t *txt_value_len_list, size_t txt_count)
```

components/mdns/mdns.c
```diff
@@ -7541,9 +7556,20 @@ static void _mdns_browse_result_add_txt(mdns_browse_t *browse, const char *insta
...
                 !_str_null_or_empty(r->service_type) && !strcasecmp(service, r->service_type) &&
                 !_str_null_or_empty(r->proto) && !strcasecmp(proto, r->proto)) {
+            bool should_update = false;
             if (r->txt) {
+                // Check if txt changed
+                if (txt_count != r->txt_count) {
+                    should_update = true;
+                } else {
+                    for (size_t txt_index = 0; txt_index < txt_count; txt_index++) {
+                        if (!is_txt_item_exist(txt[txt_index], txt_value_len[txt_index], r->txt, r->txt_value_len, r->txt_count)) {
```
david-cermak, line 7566: are the txt items in the list supposed to be ordered the same way as in the record? If so, maybe we can give it an index to start the next `for()` loop in the function from there. That way we'd save some iterations for most common usecases, where nothing changes and we'd have to go over the list twice.

david-cermak (approved):
LGTM, just two notes/questions

components/mdns/mdns.c
gytxxsy: TXT entries do not require a strict order, and entries in the old list and the new list may have the same content but in a different order. Therefore, it is necessary to traverse the entire list.

components/mdns/mdns.c
gytxxsy: It's a great suggestion. Thanks.

components/mdns/mdns.c
david-cermak: What I meant was that perhaps a tiny optimization for the most common usecase where we receive the PTR of the same service (with the same TXT) record could be done, as in this case we'd get the TXT records in the same order.

I know that they could be arranged differently, but we can handle this common case faster if we start searching from the last outer index in the inner loop (also considering that the same service gets announced multiple times) ... but that would only be beneficial if we have like hunder(s) TXT record...

anyway, just a thought.

PR #615, opened 2024-07-15
fix(mdns): Fix add/remove selfhosted services while hostname changes

david-cermak:
Reported by Bangbang and Thread team: "the mdns_service_add() function doesn't handle properly a case where hostname is changed between subsequent mdns_service_add_for_host() function call"

## Issue

`mdns_service_add()` makes a copy of the local hostname and calls using the local copy `mdns_service_add_for_host()`. When mdns's hostname is updated the local copy gets out of sync.

## Fix

Use `mdns_service_add_for_host(hostname=NULL)` for all self hosted services. Note that this the way this API should work and  is already documented:

https://github.com/espressif/esp-protocols/blob/906e447193710c1772305a218a9d0b315a549de4/components/mdns/include/mdns.h#L257

We use the local hostname only when the MDNS API is locked or inside the mdns task.


gabsuren (approved):

components/mdns/mdns.c
```diff
@@ -6075,55 +5880,36 @@ esp_err_t mdns_instance_name_set(const char *instance)
...
-        free(action);
-        return ESP_ERR_NO_MEM;
-    }
+    item->next = _mdns_server->services;
+    _mdns_server->services = item;
+    _mdns_probe_all_pcbs(&item, 1, false, false);
+    MDNS_SERVICE_UNLOCK();
 
     size_t start = xTaskGetTickCount();
     size_t timeout_ticks = pdMS_TO_TICKS(MDNS_SERVICE_ADD_TIMEOUT_MS);
```
zwx1995esp, line 5915: Question: For now, mdns will insert the service in the `_mdns_server` list directly, is this checking still needed?

david-cermak: Good catch, thanks! No longer necessary -> have removed.

common_components/linux_compat/freertos/freertos_linux.c
```diff
@@ -145,6 +157,13 @@ QueueHandle_t xSemaphoreCreateRecursiveMutex(void)
 
 void vTaskDelete(TaskHandle_t *task)
 {
+    for (int i = 0; i < s_threads; ++i) {
+        if (task == s_notifiers->id) {
```
gabsuren, line 161: The same question https://github.com/espressif/esp-protocols/pull/618#discussion_r1692587260

david-cermak: good point, already resolved and fixed in https://github.com/espressif/esp-protocols/pull/618

components/mdns/mdns.c
david-cermak: Fixed

euripedesrocha (approved):

components/mdns/mdns.c
```diff
@@ -6075,73 +5880,45 @@ esp_err_t mdns_instance_name_set(const char *instance)
...
+    mdns_service_t *s = NULL;
+
+    ESP_GOTO_ON_FALSE(_mdns_can_add_more_services(), ESP_ERR_NO_MEM, err, TAG, "Cannot add more services");
 
     mdns_srv_item_t *item = _mdns_get_service_item_instance(instance, service, proto, hostname);
-    MDNS_SERVICE_UNLOCK();
-    if (item) {
-        return ESP_ERR_INVALID_ARG;
-    }
+    ESP_GOTO_ON_FALSE(!item, ESP_ERR_INVALID_ARG, err, TAG, "Already exists");
```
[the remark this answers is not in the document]
david-cermak: Fixed, thanks!

components/mdns/mdns.c
```diff
@@ -6075,73 +5880,45 @@ esp_err_t mdns_instance_name_set(const char *instance)
...
 
-    mdns_service_t *s = _mdns_create_service(service, proto, hostname, port, instance, num_items, txt);
-    if (!s) {
-        return ESP_ERR_NO_MEM;
-    }
+    s = _mdns_create_service(service, proto, hostname, port, instance, num_items, txt);
```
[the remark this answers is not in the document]
david-cermak: could be moved inside, but we would need another one here anyway. that't the way those `ESP_GOTO_ON_FALSE()` work, they just accumulate error messages, and there's no way around it. (not that super-useful for low level coding)

I'd prefer leaving the error handling here, WDYT?

david-cermak:
Thanks for reviewing this Euripedes, please check my answers to the comments. 
Note that I also dissected one unrelated fix from the last commit:
* [fix(mdns): Fix mdns_delegate_hostname_add() to block until done](https://github.com/espressif/esp-protocols/pull/615/commits/2c1b16617eeb45195a6670ebafeb3602a8b42cb0)

PR #622, opened 2024-07-25
[eppp]: Support for transport via Ethernet (phy or emac)

david-cermak:
- [x] Using a real phy chip, via a standard Ethernet cable

---

- [ ] ~~Using a dummy phy, by means of EMAC to EMAC connection~~
- [x] ~~Added emac2emac example which demonstrates overriding (standard) Ethernet drivers with the custom one: [eth_dummy_phy](https://components.espressif.com/components/espressif/eth_dummy_phy)~~


components/eppp_link/README.md
```diff
@@ -55,3 +55,9 @@ Tested with WiFi-NAPT example
 
 * TCP - 9Mbits/s
 * UDP - 11Mbits/s
+
+### Ethernet
+
+- Internal EMAC with real PHY chip
+    * TCP - 5Mbits/s
+    * UDP - 8Mbits/s
```
kostaond, line 63: I'm quite surprised with such poor performance. Do you have idea what could be the bottle neck and why it isn't at least as good as SDIO?

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,249 @@
...
+    gpio_config_t gpio_sink_cfg = {
+        .pin_bit_mask = (1ULL << CONFIG_EPPP_LINK_ETHERNET_SINK_READY_GPIO),
+        .mode = GPIO_MODE_OUTPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,
+        .intr_type = GPIO_INTR_DISABLE
+    };
+    gpio_config(&gpio_sink_cfg);
+    gpio_set_level(CONFIG_EPPP_LINK_ETHERNET_SINK_READY_GPIO, 0);
+    vTaskDelay(pdMS_TO_TICKS(STARTUP_DELAY_MS));
```
kostaond, line 161: I'm the author of the simple EMAC 2 EMAC example but I'm not sure about this delay now :) I need to revisit.

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,249 @@
...
+        portYIELD_FROM_ISR();
+    }
+}
+#else
+#define STARTUP_DELAY_MS 500
+#endif
+
+static esp_err_t eth_init(void)
+{
+#if CONFIG_EPPP_LINK_ETHERNET_CLK_SOURCE
```
kostaond, line 128: We need to clearly distinguish that this is only ESP32 workaround. I'll try to come up with something. I'm currently thinking to wrap all this workaround code to specific functions with prefix like `esp32_gpio0_workaround`. Users then can easily remove them if they don't need them.

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,249 @@
...
+            break;
+        }
+        vTaskDelay(pdMS_TO_TICKS(100));
+    }
+    ESP_RETURN_ON_FALSE(i <= 5, ESP_FAIL, TAG, "Ethernet driver install failed");
+#else
+    ESP_RETURN_ON_ERROR(esp_eth_driver_install(&config, &s_eth_handles[0]), TAG, "Ethernet driver install failed");
+#endif // CONFIG_EPPP_LINK_ETHERNET_CLK_SINK
+    return ESP_OK;
+}
```
kostaond, line 216: `esp_restart();` of CLK source is missing when sink is restart. That could cause the sink may not boot correctly when restarted.

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,249 @@
...
+    }
+    if (s_phy) {
+        s_phy->del(s_phy);
+        s_phy = NULL;
+    }
+    free(s_eth_handles);
+    s_eth_handles = NULL;
+}
+
+#else
```
kostaond, line 235: What about adding a comment that here we use `PHY chip` to make it clearer to users? And also add comment at the start of `#ifdef CONFIG_EPPP_LINK_ETHERNET_USE_DUMMY_PHY`. Or rename it to `EMAC to EMAC`?

components/eppp_link/README.md
```diff
@@ -11,12 +11,12 @@ We usually call this node a SLAVE microcontroller. The "HOST" microcontroller ru
...
-             SLAVE micro                                  HOST micro
-    \|/  +----------------+                           +----------------+
-     |   |                |          serial line      |                |
-     +---+ WiFi NAT PPPoS |=== UART / SPI / SDIO =====| PPPoS client   |
-         |        (server)|                           |                |
-         +----------------+                           +----------------+
+             SLAVE micro                                     HOST micro
+    \|/  +----------------+                               +----------------+
+     |   |                |       (serial) line           |                |
+     +---+ WiFi NAT PPPoS |=== UART / SPI / SDIO / ETH ===| PPPoS client   |
```
kostaond, line 17: We may need to improve the README with instruction how to use EMAC to EMAC. Or at least provide link to the component?

components/eppp_link/README.md
david-cermak: would probably get similar performance as SDIO, after some optimizations; tested just briefly with all defaults.
I'm planning on looking into it at some point.

components/eppp_link/eppp_eth.c
david-cermak: I think it's a good workaround but doesn't really fit into component definition. Thinking of dissecting that and moving to example(s)

components/eppp_link/eppp_eth.c
david-cermak: 100% agree, I'll also think about it.

components/eppp_link/eppp_eth.c
david-cermak: Good point. In case of `wifi_remote` would could leave this logic on the host (master) side and let this side control the slave. But this is a generic component (on level lower than wifi-remote) and each node (server/client) is more or less generic too (peer nodes).

Cannot wait in this task (could be user's task or eppp-task), would have to create a small task to handle this case...

components/eppp_link/README.md
david-cermak: Yes, I'll improve the docs and add sections for each transport.

kostaond:
> @kostaond Thanks taking a look at the emac2emac option. I must say that I concur with your assessment and think that this option is probably not a viable transport option for `eppp` at this moment.
> 
> My suggestion is that we drop the emac2emac option in this PR, but allow the Ethernet init/start functionality to be extended (with callbacks or `__weak` functions). Then we can add emac2emac as an example and add appropriate comments about the GPIO0 workaround and the synchronization and reset stuff...

Agree, if you need to move forward, drop the emac2emac. It's currently too much polluted with the GPIO0 workaround. Presence or non presence of the workaround must be super clear to users because the most probable use case is to use ESP32 with WiFi connected to some other chip without WiFi.

david-cermak:
> It's currently too much polluted with the GPIO0 workaround. Presence or non presence of the workaround must be super clear to users because the most probable use case is to use ESP32 with WiFi connected to some other chip without WiFi.

Removed the emac2emac from the component and added it as an extra example with the specific configuration, so the GPIO0 workaround is in the app code only. (Moreover It demonstrates using any custom Ethernet driver)

components/eppp_link/examples/emac2emac/README.md
```diff
@@ -0,0 +1,12 @@
+
+# EPPP link with EMAC to EMAC transport layer
+
+This example runs a symmetrical server-client eppp application with iperf component using Ethernet transport layer, with a customized Ethernet driver from [eth_dummy_phy](https://components.espressif.com/components/espressif/eth_dummy_phy) component.
+
+Please refer to the component documentation for more information about the principle of operation and the actual physical connection between nodes.
+
+## How to use this example
+
+* Choose `CONFIG_EXAMPLE_NODE_SERVER` for the device connected as **RMII CLK Source Device**
```
euripedesrocha, line 10: We should add some advice here regarding clk source. 
@kostaond the conditions for clk generation if Wi-Fi is used in the device applies here as well, right?

components/eppp_link/examples/emac2emac/main/app_main.c
```diff
@@ -0,0 +1,201 @@
...
+        portYIELD_FROM_ISR();
+    }
+}
+#else
+#define STARTUP_DELAY_MS 500
+#endif
+
+esp_err_t eppp_transport_ethernet_init(esp_eth_handle_t *handle[])
+{
+    *handle = malloc(sizeof(esp_eth_handle_t));
```
euripedesrocha, line 58: Isn't `esp_eth_handle_t` a type alias for `void*`? Do we need to allocate here?

components/eppp_link/examples/emac2emac/main/app_main.c
```diff
@@ -0,0 +1,201 @@
...
+        .mode = GPIO_MODE_INPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_ENABLE,
+        .intr_type = GPIO_INTR_ANYEDGE
+    };
+    gpio_config(&gpio_source_cfg);
+    TaskHandle_t task_handle = xTaskGetHandle(pcTaskGetName(NULL));
+    gpio_isr_handler_add(CONFIG_EXAMPLE_RMII_CLK_READY_GPIO, gpio_isr_handler, task_handle);
+    ESP_LOGW(TAG, "waiting for RMII CLK sink device interrupt");
+    ESP_LOGW(TAG, "if RMII CLK sink device is already running, reset it by `EN` button");
```
euripedesrocha, line 76: Should we make it part of the source device to automatically reset the device if the user enables it?

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,120 @@
...
+    size_t packet_len = head->len;
+    if (len >= packet_len) {
+        esp_err_t ret = esp_netif_receive(netif, buffer + ETH_HEADER_LEN, packet_len, NULL);
+        free(buffer);
+        return ret;
+    }
+    return ESP_FAIL;
+}
+
+__attribute__((weak)) esp_err_t eppp_transport_ethernet_init(esp_eth_handle_t *handle_array[])
```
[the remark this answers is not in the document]
david-cermak: I need the customization for the emac2emac usecase in this PR. (but customization is always useful, especially for Ethernet, where initialization is far from trivial)

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,120 @@
...
+{
+    ESP_RETURN_ON_ERROR(eppp_transport_ethernet_init(&s_eth_handles), TAG, "Failed to initialize Ethernet driver");
+    ESP_RETURN_ON_ERROR(esp_eth_update_input_path(s_eth_handles[0], receive, esp_netif), TAG, "Failed to set Ethernet Rx callback");
+    sscanf(CONFIG_EPPP_LINK_ETHERNET_OUR_ADDRESS, "%2" PRIu8 ":%2" PRIu8 ":%2" PRIi8 ":%2" PRIu8 ":%2" PRIu8 ":%2" PRIu8,
+           &s_our_mac[0], &s_our_mac[1], &s_our_mac[2], &s_our_mac[3], &s_our_mac[4], &s_our_mac[5]);
+
+    sscanf(CONFIG_EPPP_LINK_ETHERNET_THEIR_ADDRESS, "%2" PRIu8 ":%2" PRIu8 ":%2" PRIi8 ":%2" PRIu8 ":%2" PRIu8 ":%2" PRIu8,
+           &s_their_mac[0], &s_their_mac[1], &s_their_mac[2], &s_their_mac[3], &s_their_mac[4], &s_their_mac[5]);
+    esp_eth_ioctl(s_eth_handles[0], ETH_CMD_S_MAC_ADDR, s_our_mac);
+    ESP_RETURN_ON_ERROR(esp_event_handler_register(ETH_EVENT, ESP_EVENT_ANY_ID, event_handler, NULL), TAG, "Failed to register Ethernet handlers");
```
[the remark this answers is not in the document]
david-cermak: I wouldn't create a custom handler/even-loop for eppp as I don't think it's necessary at this point. we can always add that later.

What would be the purpose of these custom events?

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,120 @@
...
+void eppp_transport_deinit(void)
+{
+    esp_eth_stop(s_eth_handles[0]);
+    eppp_transport_ethernet_deinit(s_eth_handles);
+}
+
+esp_err_t eppp_transport_tx(void *h, void *buffer, size_t len)
+{
+    if (!s_is_connected) {
+        return ESP_OK;
```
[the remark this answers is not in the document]
david-cermak: Good point, will fix. (the error is reported back to TCP/IP, which in most cases causes tries and floods the driver, but I think that's probably the correct behaviour)

components/eppp_link/examples/emac2emac/README.md
david-cermak: I'll add a link to the document. Yes it applies here, but only for in this example and on ESP32 target

components/eppp_link/examples/emac2emac/main/app_main.c
david-cermak: yes, we do. that is the customization point of Ethernet init.

> Isn't esp_eth_handle_t a type alias for void*

That's because the `ethernet_init` component returns an array of ethernet handles.

components/eppp_link/examples/emac2emac/main/app_main.c
david-cermak: we would need some other signalling method to perform the reset of the other device, right? (I think @bogdankolendovskyy suggested a similar handshaking mechanism with GPIOs like that, but it was rejected by @kostaond for being "too complicated", correct?)

david-cermak:
@kostaond @euripedesrocha Thanks for your review, I've reworked this PR, so that it uses "just Ethernet" with customizable initialization. Then (in a separate commit) I've added the EMAC-2-EMAC example with all those workarounds and magic delays.

components/eppp_link/CMakeLists.txt
```diff
@@ -1,3 +1,16 @@
...
+endif()
+
+if(CONFIG_EPPP_LINK_DEVICE_SDIO)
+    set(transport_src eppp_sdio_slave.c eppp_sdio_host.c)
+endif()
+
+idf_component_register(SRCS eppp_link.c ${transport_src}
                     INCLUDE_DIRS "include"
-                    PRIV_REQUIRES esp_netif esp_driver_spi esp_driver_gpio esp_timer driver)
+                    PRIV_REQUIRES esp_netif esp_driver_spi esp_driver_gpio esp_timer driver esp_eth)
```
kostaond, line 11: I'm not sure if `driver` is not going to be obsolete soon. On the other hand, if you just removed it, it wouldn't work with older IDF version. Please consider handle it as https://github.com/espressif/esp-eth-drivers/blob/master/enc28j60/CMakeLists.txt#L3

components/eppp_link/README.md
kostaond: Have you looked at it?

components/eppp_link/examples/emac2emac/main/app_main.c
kostaond: The problem of Bogdan's proposal was the devices startup would have to be synchronized for **both** devices. The timing was my main concern. Maybe if it was performed in a loop - the first device would send pulse then wait for response. If response didn't come, then try again. 

I also didn't like GPIO direction switching but I don't see it as bad as I used to today.

And it would also be really good to separate these functions since it is just a workaround. It should not be part of `dummy_phy` but it should be separate part of example.

components/eppp_link/eppp_eth.c
```diff
@@ -0,0 +1,120 @@
...
+    if (!s_is_connected) {
+        return ESP_FAIL;
+    }
+    // setup Ethernet header
+    header_t *head = (header_t *)s_out_buffer;
+    memcpy(head->dst, s_their_mac, ETH_ADDR_LEN);
+    memcpy(head->src, s_our_mac, ETH_ADDR_LEN);
+    head->len = len;
+    // handle frame size: ETH_MIN_PAYLOAD_LEN <= len <=  ETH_MAX_PAYLOAD_LEN
+    size_t frame_payload_len = len < ETH_MIN_PAYLOAD_LEN ? ETH_MIN_PAYLOAD_LEN : len;
```
kostaond, line 114: I think it's not necessary needed since EMAC should add padding bits by itself.

components/eppp_link/examples/emac2emac/main/register_iperf.c
```diff
@@ -0,0 +1,179 @@
+/*
```
kostaond, line 1: Consider using https://components.espressif.com/components/espressif/iperf-cmd/versions/0.1.3

euripedesrocha (approved):

components/eppp_link/README.md
david-cermak: Yes, I think it's because of the PPP implementation in lwip (which is steam based, while Ethernet is frame based -- so we're effectively creating packets from packets again). I'm planning on addressing it using with [this idea](https://github.com/espressif/esp-protocols/pull/481) after we merge this PR

components/eppp_link/examples/emac2emac/main/app_main.c
david-cermak: That's why I added it as a separate example. But I think I'll just remove it, as I hear that it's too messy and packed with workarounds.

Maybe revisit it later and add ESP32P4-ESP32 example with emac-emac option.

kostaond: Drop it, we've already agreed on it :smile: https://github.com/espressif/esp-protocols/pull/622#issuecomment-2252332193

david-cermak: > Drop it, we've already agreed on it 😄 https://github.com/espressif/esp-protocols/pull/622#issuecomment-2252332193

Yeah, but I made significant changes after that and moved all the workarounds from components to that example.

david-cermak: Anyways, emac2emac exampled removed.

components/eppp_link/CMakeLists.txt
david-cermak: Thanks for the pointers. Have updated!

components/eppp_link/eppp_eth.c
david-cermak: removed, thanks!

kostaond (approved):

david-cermak:
Thanks for your review, Ondrej & Euripedes!

Issue #412, opened 2023-11-01
Destructor race in esp_websocket_client v1.1.0: context deleted under worker task before it exits (IDFGH-11360)


DaStoned:
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I've been using esp_websocket_client 1.1.0 together with a GSM module (SIM7070G) connected via USB. Microcontroller is an ESP32S3 with 16 MB Flash and 8 MB SPIRAM, the PCB is a custom one with the GSM module on it. ESP IDF version 5.1.1.

There are rather frequent crashes (probability is around 10-20% in my project) when the Internet connection dies. Crashes occur because worker task `esp_websocket_client_task()` accesses deleted heap memory behind pointer `client` before exiting. So there is a race between this worker task exiting and its context being deleted with the same heap area re-used by another task. The sequence of events leading to a crash is following:

1. My code creates a new WS client where `disable_auto_reconnect` is set to `true` (otherwise the issue is masked). This malloc-s the client context from heap. It then connects to a WS server and happily exchanges some data.
2. At a random moment the GSM module stops communicating data (they have a nasty habit of occasionally doing that). 
3. The WS worker task `esp_websocket_client_task()` is in the process of reading data from the server using `esp_websocket_client_recv()`, which  fails due to the communication blackout. It aborts the connection by calling `esp_websocket_client_abort_connection()` [here](https://github.com/espressif/esp-protocols/blob/7c5a832821f0fac53cb9ae54e11a16a5a90b4297/components/esp_websocket_client/esp_websocket_client.c#L987C35-L987C35): 
4. `esp_websocket_client_abort_connection()` does two things relevant to this issue. Firstly it sets the client state to `WEBSOCKET_STATE_WAIT_TIMEOUT`, secondly it dispatches event `WEBSOCKET_EVENT_DISCONNECTED` to my code. 

Things now branch off into two different tasks, as my code is not allowed to kill the WS client in the same task where it received `WEBSOCKET_EVENT_DISCONNECTED`. My event handler posts a message to another thread to start the process of cleaning up and re-connecting the Websocket connection (remember, auto-reconnect is disabled). I'll describe what happens in the sequence of events that I observe from logs:

[...]

PR #626, opened 2024-08-12
Fix race condition when client set to not reconnect

euripedesrocha:
Fix: #412


david-cermak (approved):
LGTM!

FWIW we already support destroying asynchronously, so the usecase described in https://github.com/espressif/esp-protocols/issues/412 would not be necessary (still possible though), see:

https://github.com/espressif/esp-protocols/blob/e425a3c504e308ada7e8f594239c24ecb033d3b6/components/esp_websocket_client/include/esp_websocket_client.h#L228-L238

johanstokking:
Thanks for this @euripedesrocha.

PR #627, opened 2024-08-13
Make example to use certificate bundle

euripedesrocha:
To easy maintenance, makes the example on websocket client to use certificate bundle by default.


david-cermak (commented):
I agree that using the cert bundle is a better choice when working with a public broker

PR #629, opened 2024-08-14
esp_websocket: fix a crash when network anomaly (IDFGH-13461)

shootao:
when disconnect the wifi, the   esp_websocket_client_task continue to called 
esp_transport_poll_read to read, but the default the time is 1s,  then the read had timeout will called esp_websocket_client_abort_connection to destory some handle, at this time, esp_websocket_client_send_with_exact_opcode still blocked.
when esp_websocket_client_send_with_exact_opcode  had timeout, will trigger crash ， because handle(erro handle) already release


david-cermak (approved):
Thanks for the fix!

could you please click trough the CLA agreement so we can accept your contribution?

david-cermak:
Thank you for the update! I noticed that there are now two commits from different authors, and the pre-commit test didn’t pass.

Could you please review our [contribution guidelines](https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md) and update this PR accordingly? Specifically, it would be great if you could:
* Fix the commit messages
* Squash the two commits into one
* Rebase the branch (we recently resolved some CI issues on the master branch)

If you’d prefer, just let me know -- I can take care of the chores and push your work in a separate PR, while still crediting your original commit.

david-cermak:
@shootao Thanks again for the fixes. will merge by means of https://github.com/espressif/esp-protocols/pull/684

PR #630, opened 2024-08-15
feat(websocket): propagate esp_tls stack error and cert verify flags

johanstokking:
Even though the `esp_tls_stack_err` and `esp_tls_cert_verify_flags` are get, they are not propagated through event data. This fixes that.

I came across this because I noticed that an Mbed TLS error code that I was interested in (`MBEDTLS_ERR_SSL_ALLOC_FAILED`) was correctly logged by the component, but `esp_tls_stack_err` was uninitialized.


david-cermak (approved):
Thanks for fixing this issue!

These errors are even used(!) in examples, referring and printing some old/invalid values.

PR #632, opened 2024-08-16
[CI]: Fixes with recent IDF updates 

david-cermak:
* [Examples]: Ignore nearly full partition on C6 builds
* [Websocket]: cherry-picked @euripedesrocha 's fix from https://github.com/espressif/esp-protocols/pull/627 and updated to still use the generated self-signed certs when connecting to the python test server


euripedesrocha (approved):

david-cermak:
Thanks for the review!

PR #633, opened 2024-08-16
Feat/websocket: begin end thread events (IDFGH-13507)

bryghtlabs-richard:
Add two new events to signal the start and end of websocket thread operation.

Also update the websocket event handler documentation, as it was missing a few.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -970,6 +970,7 @@ static void esp_websocket_client_task(void *pv)
 
     client->state = WEBSOCKET_STATE_INIT;
     xEventGroupClearBits(client->status_bits, STOPPED_BIT | CLOSE_FRAME_SENT_BIT);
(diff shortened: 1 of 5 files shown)


david-cermak (commented):
> [docs(websocket): improve websocket event list](https://github.com/espressif/esp-protocols/pull/633/commits/9cf41636634409bc460e39d90a3c0b390d58c617)

LGTM, thanks for contributing and fixing the docs!

> [feat(websocket): add events for begin/end thread](https://github.com/espressif/esp-protocols/pull/633/commits/d7fa24bc20efcf4adc155431582f8a8e948a2ee9)

What's the purpose of these new events, in general, could you just please briefly explain?

bryghtlabs-richard:
It's useful for encapsulating all the other events. In my case, I'm using to to configure a thread-local storage value and destroying it once done.

david-cermak (approved):

david-cermak:
> The new events are useful for encapsulating all the other events. In my case, I'm using to to configure a thread-local storage value and destroying it once done.

Makes sense, thanks for explaining!

PR #642, opened 2024-08-27
[mdns]: Fix recent issues on v1.4 reported by Coverity

david-cermak:
##    fix(mdns): Fixed dead-code reported by coverity
    
Fixes CID 467738: Logically dead code in mdns.c, _copy_mdns_txt_items
Introduced by probably by a merge confilict, as the fix was added in
two separate PRs, merging d4da9cb0 first and 8a690503 later

##     fix(mdns): Fix use after free reported by coverity
    
Fixes CID 467739: Use after free in mdns.c, mdns_service_remove_for_host
We should look only for one match in the service list, since if we
assume there could be aliases, we might free one and reference the
other.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -5956,11 +5956,6 @@ static mdns_txt_item_t *_copy_mdns_txt_items(mdns_txt_linked_item_t *items, uint
     for (mdns_txt_linked_item_t *tmp = items; tmp != NULL; tmp = tmp->next) {
         ret_index++;
     }
-    if (ret_index == 0) {
-        *txt_count = 0;
-        *txt_value_len = NULL;
-        return NULL;
-    }
     *txt_count = ret_index;
     if (ret_index == 0) {   // handle empty TXT
         *txt_value_len = NULL;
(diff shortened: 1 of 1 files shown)


components/mdns/mdns.c
```diff
@@ -5956,11 +5956,6 @@ static mdns_txt_item_t *_copy_mdns_txt_items(mdns_txt_linked_item_t *items, uint
     for (mdns_txt_linked_item_t *tmp = items; tmp != NULL; tmp = tmp->next) {
         ret_index++;
     }
-    if (ret_index == 0) {
```
euripedesrocha, line 5959: Isn't this covering the case for `items==NULL`? Since it is a static function, I expect that this case is covered in the call site.

david-cermak: It's not covered by the calling party (I think correctly) due to hierarchical reasons, as we can call this multiple times to make copies of txt(s) without needing to check for `NULL`.

Anyway, this fix is actually removing the redundant fix, which is already there, just few lines below:

https://github.com/espressif/esp-protocols/blob/25b3d5fd7b1d42ba52c8faf819080ebec9af860b/components/mdns/mdns.c#L5959-L5963

(fixing merge conflict)

euripedesrocha (approved):

Issue #624, opened 2024-07-26
Request for Additional Method in ESP Modem: Handling Larger AT Command Responses (IDFGH-13338)


srikanthpalvai:
### Is your feature request related to a problem?

When issuing the commands AT+CMGF=0 and AT+CMGL=4, the response results in more data than usual. To handle this, we need an additional method that takes max_len as an argument.

### Describe the solution you'd like.

This method allows the user to specify the maximum length of the output buffer, ensuring that larger responses can be handled appropriately.

extern "C" esp_err_t esp_modem_at_raw_ext(esp_modem_dce_t *dce_wrap, const char *cmd, char *p_out, const char *pass, const char *fail, int timeout, **size_t max_len**)
{
    if (dce_wrap == nullptr || dce_wrap->dce == nullptr) {
        return ESP_ERR_INVALID_ARG;
    }
    std::string out;
    auto ret = command_response_to_esp_err(dce_wrap->dce->at_raw(cmd, out, pass, fail, timeout));
    if ((p_out != NULL) && (!out.empty())) {
        strlcpy(p_out, out.c_str(), **max_len**);
    }
    return ret;
}

### Describe alternatives you've considered.

#define STR_MAX 1024

esp_modem_dce_t *dce_wrap;
char response[ESP_MODEM_C_API_STR_MAX];
esp_err_t ret = esp_modem_at_raw_ext(dce_wrap, "AT+CMGL=4", response, "OK", "ERROR", 1000, STR_MAX );
if (ret == ESP_OK) {
    // Process the response
}

srikanthpalvai:
@david-cermak 
Could you please have a look on it.

david-cermak:
Yes, I'll look into it, planning on some maintenance work and the v1.2 release in about two weeks.
Meantime, we're open for ideas/suggestions and PRs are welcome.

srikanthpalvai:
Hi @david-cermak

I was not able to create a pull request for esp-modem
but here is the patch file it will help you.
[0001-Add-max_len-parameter-to-esp_modem_at_raw_ext-for-bu.patch](https://github.com/user-attachments/files/16462228/0001-Add-max_len-parameter-to-esp_modem_at_raw_ext-for-bu.patch)

srikanthpalvai:
Hi @david-cermak 
I have fixed it with below logic. Added it in the project root CMakeLists.txt

[...]

Issue #184, opened 2022-12-01
How to increase size of AT commands response? (IDFGH-8868)


MRCBoss:
Hello. I'm using ESP32 + BG96, 1 serial channel, CMUX, esp-modem version 0.1.24. Data mode sends sensor data beyond MQTT (PPP) and I need command mode to get GNSS data. The AT+QGPSLOC command response is received truncated, because of ESP_MODEM_C_API_STR_MAX limit of 64. How could I change this limit without modifying the ESP_MODEM_C_API_STR_MAX definition inside esp_modem_c_api.cpp code?

david-cermak:
Hi @MRCBoss 

You can define it directly in the project makefile by adding these two lines after project definitions
```cmake
idf_component_get_property(esp_modem esp_modem COMPONENT_LIB)
target_compile_definitions(${esp_modem} PRIVATE  "ESP_MODEM_C_API_STR_MAX=128")
```

(e.g. for the modem-console example, place the above after this line:

https://github.com/espressif/esp-protocols/blob/457f8335bb876b8af80a65b096e5381f49f29878/components/esp_modem/examples/modem_console/CMakeLists.txt#L6

)

david-cermak:
Have you placed into the project makefile (not the `CMakeLists.txt` in the main folder) and just **after** the `project()`directive? This should work for correctly, at least on IDF v4.4 and v5.+.

Which IDF version are you using?

MRCBoss:
Yes, I've placed into the project CMakeLists.txt, but I had put it **before** the project() directive. Now I've put it **after** that directive and the same error ocurred. I'm using IDF 4.4.2.
The esp-modem managed component is used inside a custom component (not main), could this be related?.

david-cermak:
Oh, I see! In that case, we have to specify the **full name** of the component (including namespace *"espressif"*), which is `espressif__esp_modem`

The cmake commands should then look like:
```cmake
cmake_minimum_required(VERSION 3.5)

include($ENV{IDF_PATH}/tools/cmake/project.cmake)
project(your-project)

idf_component_get_property(esp_modem espressif__esp_modem COMPONENT_LIB)
target_compile_definitions(${esp_modem} PRIVATE  "ESP_MODEM_C_API_STR_MAX=128")
```

MRCBoss:
Now it works fine. Thank you @david-cermak !

PR #649, opened 2024-09-11
[modem]: Add ability to change ESP_MODEM_C_API_STR_MAX from Kconfig (IDFGH-13668)

embedcat:
## Description
This PR introduce ability to change ESP_MODEM_C_API_STR_MAX from Kconfig (idf.py menuconfig).

In esp_modem in C-API there is ESP_MODEM_C_API_STR_MAX which defines max size of AT response returned to user.
There is no obvious way to increase this size. Some workaround was found in #184, but I think use Kconfig option is more clear and straightforward.

## Related
#184
#624

---

## Checklist


david-cermak (approved):
Thanks for addressing this problem!
Agree that making this a config option is much cleaner then injecting via build system, LGTM!

PR #654, opened 2024-09-18
[modem]: Fix arguments names when spawn esp_modem_xxx declarations

embedcat:
## Description

This PR fixes arguments names which auto-declared by `DECLARE_ALL_COMMAND_APIS()` (`"include/generate/esp_modem_command_declare.inc"`). This time with clean commit history.

For example, it would be:
`esp_err_t esp_modem_send_sms(esp_modem_dce_t *dce, const char *number, const char *message);`
instead of:
`esp_err_t esp_modem_send_sms(esp_modem_dce_t *dce, const char *p1, const char *p2);`

Also, it does clear auto completions and tips in IDE (VS Code + ESP-IDF):


david-cermak (approved):
Thanks for the updates!

PR #655, opened 2024-09-18
[modem]: Update target test builds to use external Catch2

david-cermak:
Switched from the deprecated Catch framework (IDF/tools) to Catch2, as Catch will be removed in v6.0.
Note that Catch2 has higher memory requirements,
necessitating an increase in stack size and partition table. Additionally, Catch2 increases compilation time due to its larger footprint.


components/esp_modem/test/target/main/pppd_test.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2022-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2022-2023 Espressif Systems (Shanghai) CO LTD
```
euripedesrocha, line 2: Are we going back in time :)?

euripedesrocha (approved):

components/esp_modem/test/target/main/pppd_test.cpp
david-cermak: Thanks for noticing, fixed!

> Are we going back in time :)?

We are, actually. Created this commits few days back, but had to temporarily modify the test to merge URC before investigating target test failures + wanted to rule out the catch2 suspect :-)
Then, after reverting to this commit again, the copyright changed back and no pre-commit hook nor the CI ever noticed :shrug:

PR #643, opened 2024-08-30
[wifi_remote]: Make esp_hosted default RPC library

david-cermak:
Also updates https://github.com/espressif/esp-idf/blob/6e5414b6c4f265a0adfb56a15fbfbe6beb1f8373/examples/protocols/mqtt/tcp/README.md?plain=1#L63-L66

PR #660, opened 2024-09-18
[wifi_remote]: Make flat version structure across IDF releases

david-cermak:
Adds multiple IDF release branch support to a single version of `wifi-remote` to simplify release workflow of this component

- [x] Add tests with IDF examples, where `esp_wifi_remote` is used


david-cermak:
@euripedesrocha @mantriyogesh PTAL at this proposal, I'd like to continue with linear versioning of wifi-remote (and replace https://github.com/espressif/esp-protocols/pull/644 https://github.com/espressif/esp-protocols/pull/643 with this one)

still need to cleanup commit history and build related IDF examples.
(also wanted to check with IDF tools people on concerns they might have with flattening this)

components/esp_wifi_remote/examples/mqtt/sdkconfig.defaults
```diff
@@ -1,3 +1,4 @@
+CONFIG_ESP_WIFI_REMOTE_LIBRARY_EPPP=y
```
mantriyogesh, line 1: do we still wish to use eppp as default solution here?

david-cermak: I would still keep `eppp` as default for this example (since I wanted to build it with wifi capable chips as well, which doesn't seem to work with esp-hosted)

components/esp_wifi_remote/idf_v5.4/esp_wifi_remote_weak.c
```diff
@@ -0,0 +1,432 @@
+/*
```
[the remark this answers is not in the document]
david-cermak: > Any changes from application/hosted side needed for this?

No change should be necessary, as the paths have been updated in CMake as well, so all headers are still available from public path (locations of sources shouldn't cause any trouble)

mantriyogesh:
LGTM..

euripedesrocha (approved):

PR #671, opened 2024-10-08
feat(sockutls): Add initial support for socket helpers

david-cermak:
* initial support for socket helper library
* useful for porting linux based libraries where performance and memory is not an issue

:warning: Very **WIP**

## Supports:

* ifaddrs
* socketpair
* pipe
* getnameinfo


.pre-commit-config.yaml
```diff
@@ -61,8 +61,8 @@ repos:
   - repo: local
     hooks:
       - id: commit message scopes
-        name: "commit message must be scoped with: mdns, modem, websocket, asio, mqtt_cxx, console, common, eppp, tls_cxx, mosq"
-        entry: '\A(?!(feat|fix|ci|bump|test|docs|chore)\((mdns|modem|common|console|websocket|asio|mqtt_cxx|examples|eppp|tls_cxx|mosq)\)\:)'
+        name: "commit message must be scoped with: mdns, modem, websocket, asio, mqtt_cxx, console, common, eppp, wifi_remote, tls_cxx, mosq, sockutls"
+        entry: '\A(?!(feat|fix|ci|bump|test|docs)\((mdns|modem|common|console|websocket|asio|mqtt_cxx|examples|eppp|wifi_remote|tls_cxx|mosq|sockutls)\)\:)'
```
david-cermak, line 65:
```suggestion
        name: "commit message must be scoped with: mdns, modem, websocket, asio, mqtt_cxx, console, common, eppp, tls_cxx, mosq, sockutls"
        entry: '\A(?!(feat|fix|ci|bump|test|docs|chore)\((mdns|modem|common|console|websocket|asio|mqtt_cxx|examples|eppp|tls_cxx|mosq|sockutls)\)\:)'
```

components/sock_utils/include/ifaddrs.h
```diff
@@ -0,0 +1,56 @@
...
+#endif
+
+/**
+ * @brief Simplified version of ifaddr struct
+ */
+struct ifaddrs {
+    struct ifaddrs  *ifa_next;    /* Next item in list */
+    char            *ifa_name;    /* Name of interface */
+    struct sockaddr *ifa_addr;    /* Address of interface */
+    int ifa_flags;
```
kostaond, line 32: Should be `unsigned int` per https://man7.org/linux/man-pages/man3/getifaddrs.3.html

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+    freeifaddrs(ifaddr);
+    *ifap = NULL;
+    return ret;
+
+}
+
+int getifaddrs(struct ifaddrs **ifap)
+{
+    if (ifap == NULL) {
+        return -1; // Invalid argument
```
kostaond, line 63: Should we also set `errno = EINVAL;`

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+
+    while ((netif = esp_netif_next_unsafe(netif)) != NULL) {
+        ESP_GOTO_ON_FALSE((ifaddr = (struct ifaddrs *)calloc(1, sizeof(struct ifaddrs))),
+                          ESP_ERR_NO_MEM, err, TAG, "Failed to allocate ifaddr");
+        if (next_addr == NULL) {    // the first address -> attach the head
+            *ifap = ifaddr;
+        } else {
+            *next_addr = ifaddr;    // attach next addr
+        }
+        ESP_GOTO_ON_FALSE((ifaddr->ifa_name = strdup(esp_netif_get_ifkey(netif))),
```
kostaond, line 34: I've already experienced (PTP porting) minor discourse between usual interface names used on Linux and in `esp_netif`. For example, `eth0` is quite common on Linux but it's `ETH_DEF` for `esp_netif` and `eth0` is actually `if_desc` in `esp_netif`.

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+
+}
+
+int getifaddrs(struct ifaddrs **ifap)
+{
+    if (ifap == NULL) {
+        return -1; // Invalid argument
+    }
+
+    return esp_netif_tcpip_exec(getifaddrs_unsafe, ifap) == ESP_OK ? 0 : -1;
```
kostaond, line 66: Should we set `errno` based on returned value?

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+        if (next_addr == NULL) {    // the first address -> attach the head
+            *ifap = ifaddr;
+        } else {
+            *next_addr = ifaddr;    // attach next addr
+        }
+        ESP_GOTO_ON_FALSE((ifaddr->ifa_name = strdup(esp_netif_get_ifkey(netif))),
+                          ESP_ERR_NO_MEM, err, TAG, "Failed to allocate if name");
+        ESP_GOTO_ON_FALSE((addr_in = (struct sockaddr_in *)calloc(1, sizeof(struct sockaddr_in))),
+                          ESP_ERR_NO_MEM, err, TAG, "Failed to allocate addr_in");
+        ESP_GOTO_ON_ERROR(esp_netif_get_ip_info(netif, &ip), err, TAG, "Failed to allocate if name");
```
kostaond, line 38: Check the error log message. Seems to be unrelated.

kostaond, line 38: It seems that `esp_netif_get_ip_info` always return some IP address. But what is the IP address when e.g. link was UP, interface got IP address from DHCP server and now link is DOWN. I guess it's the old address. However, can we consider it valid address? This question is also connected with line 43 (`ifaddr->ifa_flags = IFF_UP;`), can we always blindly set interface UP?

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (next_addr == NULL) {
+        *ifap = NULL;       // no addresses found
+    } else {
+        *next_addr = NULL;  // terminate the list
+    }
+    return ret;
+
+err:
+    freeifaddrs(ifaddr);
```
kostaond, line 54: This points to last process record, so if a fail occurs in e.g. in second iteration, the previously allocated address records are not freed, if I understand it correctly.

components/sock_utils/src/gai_strerror.c
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <stdio.h>
+#include "gai_strerror.h"
+
+const char *gai_strerror(int ecode)
+{
+    static char str[32];
```
espressif-abhikroy, line 11: Returning a static buffer is not thread safe.

components/sock_utils/src/socketpair.c
```diff
@@ -0,0 +1,88 @@
...
+    ESP_GOTO_ON_FALSE(getsockname(listenfd, (struct sockaddr *)&ss, &ss_len) >= 0, -1, err, TAG, "getsockname failed");
+
+    sa->sin_family = AF_INET;
+    sa->sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+
+    fd1 = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    ESP_GOTO_ON_FALSE(fd1 != INVALID_SOCKET, -1, err, TAG, "Cannot create read side socket");
+    ESP_GOTO_ON_FALSE(set_nonblocking(fd1) == 0, -1, err, TAG, "Failed to set socket to nonblocking mode");
+    if (connect(fd1, (struct sockaddr *)&ss, ss_len) < 0) {
+        ESP_GOTO_ON_FALSE(errno == EINPROGRESS || errno == EWOULDBLOCK, -1, err, TAG, "Failed to connect fd1");
```
espressif-abhikroy, line 56: Here only two errors are handled. Can we handle any error?

components/sock_utils/test/host/main/test_sock_utils.cpp
```diff
@@ -0,0 +1,139 @@
...
+    esp_netif_destroy(esp_netif);
+    esp_netif_destroy(esp_netif2);
+}
+
+static void test_pipe(int read_fd, int write_fd)
+{
+    CHECK(read_fd  >= 0);
+    CHECK(write_fd >= 0);
+    CHECK(read(read_fd, buffer, sizeof(buffer)) < 0);
+    CHECK(write(write_fd, buffer, 1) > 0);
```
kostaond, line 89: Question, is size of the data to be passed over pipe limited? I mean, I presume lwip infrastructure is used for this pipe implementation if I understand it correctly. So my concern is if lwip does not limit it to e.g. maximum Ethernet frame size size.

components/sock_utils/src/socketpair.c
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (fd2 != INVALID_SOCKET) {
+        close(fd2);
+    }
+    return ret;
+}
+
+int pipe(int pipefd[2])
+{
+    if (socketpair(AF_UNIX, SOCK_STREAM, 0, pipefd) == -1) {
```
espressif-abhikroy, line 84: Will this implementation of `pipe` restrict the bidirectional use of `socketpair`?

components/sock_utils/test/host/main/test_sock_utils.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "catch2/catch_test_macros.hpp"
+#include "catch2/catch_session.hpp"
+
+static char buffer[64];
+
+static esp_netif_t *create_test_netif(const char *if_key)
+{
+    esp_netif_inherent_config_t base_cfg = ESP_NETIF_INHERENT_DEFAULT_WIFI_STA();
+    esp_netif_ip_info_t ip = { };
+    ip.ip.addr = ESP_IP4TOADDR(1, 2, 3, 4);
```
kostaond, line 18: If IP address is parameter evaluated that address is gotten correctly, I think we should have different address for each interface to distinguish among them.

components/sock_utils/src/socketpair.c
kostaond, line 84: Pipe seems to be blocking by default per https://man7.org/linux/man-pages/man7/pipe.7.html but `socketpair` sets sockets to non-blocking.

components/sock_utils/src/getnameinfo.c
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <fcntl.h>
+#include "lwip/sockets.h"
+#include "lwip/netdb.h"
+#include "ifaddrs.h"
+#include "esp_check.h"
+
+int getnameinfo(const struct sockaddr *addr, socklen_t addrlen,
```
alisitsyn, line 12: Is there any constraints to support `sockaddr_in6` and family `AF_INET6`?

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+
+static esp_err_t getifaddrs_unsafe(void *ctx)
+{
+    struct ifaddrs **ifap = ctx;
+    struct ifaddrs *ifaddr = NULL;
+    struct ifaddrs **next_addr = NULL;
+    struct sockaddr_in *addr_in = NULL;
+    esp_netif_t *netif = NULL;
+    esp_netif_ip_info_t ip;
+    int ret = ESP_OK;
```
alisitsyn, line 24: Should the type be `esp_err_t` instead ?

components/sock_utils/test/host/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.16)
+
+set(COMPONENTS main)
+
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+
+project(sockutls_host_test)
```
alisitsyn, line 7: Minor: typo `sockutils_host_test`?

components/sock_utils/include/ifaddrs.h
```diff
@@ -0,0 +1,56 @@
...
+#define freeifaddrs esp_freeifaddrs
+#endif
+
+/**
+ * @brief Simplified version of ifaddr struct
+ */
+struct ifaddrs {
+    struct ifaddrs  *ifa_next;    /* Next item in list */
+    char            *ifa_name;    /* Name of interface */
+    struct sockaddr *ifa_addr;    /* Address of interface */
```
alisitsyn, line 31: Should the `ifa_data - address-family- specific data` in spite that support for other family is not implemented? It is better to keep it for compatibility reason with linux API and set it to NULL.

david-cermak:
Name of the component is misleading, consider:
* porting
* linux
help people find this component

david-cermak:
Document how and where are the prototypes defined

david-cermak:
> Document how and where are the prototypes defined

Add to the IDF hints -- if we get pipe() linkage error -> please install sock-utils (or whatever w'll call it)

components/sock_utils/include/netdb_macros.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#ifndef NI_NUMERICHOST
```
alisitsyn, line 8: Minor: Can the 'NI_MAXHOST', 'NI_MAXSERV' be added here to specify the maximum sizes required for the buffers?
This will allow to avoid magic numbers.

components/sock_utils/src/ifaddrs.c
david-cermak: will do

components/sock_utils/src/ifaddrs.c
david-cermak: we also have lwip names, which are something like `eth0`

david-cermak: here we should return the lwip name

david-cermak: will fix it here

components/sock_utils/src/ifaddrs.c
david-cermak: yes, will set the errno

components/sock_utils/src/ifaddrs.c
david-cermak: i'll fix it to return address as well as status, if the netif is UP it will only set IFF_UP flag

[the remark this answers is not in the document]
david-cermak: will fix

components/sock_utils/src/ifaddrs.c
david-cermak: I'll fix it

components/sock_utils/src/gai_strerror.c
david-cermak: need to create a hardcoded list of errors.

david-cermak: consider the TLS here `__thread_local_static...`

components/sock_utils/src/getnameinfo.c
```diff
@@ -0,0 +1,48 @@
...
+        return EAI_BADFLAGS;
+    }
+
+    const struct sockaddr_in *sinp = (const struct sockaddr_in *) addr;
+
+    switch (addr->sa_family) {
+    case AF_INET:
+        // Handle numeric host address
+        if (flags & NI_NUMERICHOST) {
+            if (inet_ntop(AF_INET, &sinp->sin_addr, host, hostlen) == NULL) {
```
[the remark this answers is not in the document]
david-cermak: need to check for NULL

components/sock_utils/include/poll.h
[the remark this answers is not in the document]
david-cermak: add a comment

components/sock_utils/src/socketpair.c
```diff
@@ -0,0 +1,88 @@
...
+    socklen_t ss_len = sizeof(struct sockaddr_in);
+    int fd1 = INVALID_SOCKET;
+    int fd2 = INVALID_SOCKET;
+    int listenfd = INVALID_SOCKET;
+    int ret = 0; // Success
+
+    sa->sin_family = AF_INET;
+    sa->sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+    sa->sin_port = 0;
+    listenfd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
```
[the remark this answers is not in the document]
david-cermak: i'll return an error if UDP is passed and add into a limitation section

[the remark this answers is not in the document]
david-cermak: wil fix

components/sock_utils/test/host/main/test_sock_utils.cpp
```diff
@@ -0,0 +1,139 @@
...
+    base_cfg.if_key = if_key;
+    esp_netif_config_t cfg = { .base = &base_cfg, .driver = NULL, .stack = ESP_NETIF_NETSTACK_DEFAULT_WIFI_STA };
+    return esp_netif_new(&cfg);
+}
+
+TEST_CASE("esp_getnameinfo() for IPv4", "[sock_utils]")
+{
+    struct sockaddr_in sock_addr = {};
+    sock_addr.sin_family = AF_INET;
+    sock_addr.sin_port = 257; //  (0x0101: same number for LE and BE)
```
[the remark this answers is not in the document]
david-cermak: this is wrong! I'll fix it someway

components/sock_utils/test/host/main/test_sock_utils.cpp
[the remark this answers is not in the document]
david-cermak: yes, will add the scenario, write multiple times before reading

components/sock_utils/src/socketpair.c
david-cermak: we handle all other errors, but it's okay if the errno is either EINPROGRESS or EWOULDBLOCK

we can log the actual errno.

components/sock_utils/include/ifaddrs.h
david-cermak: i'll reconsider

components/sock_utils/src/ifaddrs.c
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return esp_netif_tcpip_exec(getifaddrs_unsafe, ifap) == ESP_OK ? 0 : -1;
+}
+
+void freeifaddrs(struct ifaddrs *ifa)
+{
+    while (ifa != NULL) {
+        struct ifaddrs *next = ifa->ifa_next;
+        if (ifa->ifa_name) {
```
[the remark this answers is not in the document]
david-cermak: will remove

components/sock_utils/include/netdb_macros.h
david-cermak: will use these number when allocating buffers in tests

.pre-commit-config.yaml
david-cermak: fixed

components/sock_utils/src/gai_strerror.c
david-cermak: used TLS

components/sock_utils/include/ifaddrs.h
david-cermak: fixed to use `unsigned int`

components/sock_utils/src/ifaddrs.c
david-cermak: Sets `errno` here and in other places.

components/sock_utils/src/ifaddrs.c
david-cermak: used lwip name here, which is typically: `st1`, `et2`

components/sock_utils/src/ifaddrs.c
david-cermak: Added a switch-case block to translate esp_err_t to errno

components/sock_utils/src/ifaddrs.c
david-cermak: fixed

david-cermak: setting the flag only if the netif is up (per `esp_netif_is_netif_up()`) and tested on wifi (up) +ethernet (down)

components/sock_utils/src/ifaddrs.c
david-cermak: fixed

components/sock_utils/src/getnameinfo.c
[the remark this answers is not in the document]
david-cermak: lwip's `inet_ntop()` actually addresses the nullcheck and return an error.
The real `getnameinfo()`, however, succeeds (returns 0) if `host` is NULL.
Added the condition to be in line with linux's `getnameinfo()` behaviour.

components/sock_utils/include/gai_strerror.h
```diff
@@ -0,0 +1,31 @@
...
+extern "C" {
+#endif
+
+#ifdef CONFIG_IDF_TARGET_LINUX
+// namespace with esp_ on linux to avoid duplication of symbols
+#define gai_strerror esp_gai_strerror
+#endif
+
+/**
+* @brief Returns a string describing a `getaddrinfo()` error code.
```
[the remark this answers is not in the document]
david-cermak: updated the comments

components/sock_utils/include/netdb_macros.h
david-cermak: used `NI_MAXSERV` for buffer size in tests

david-cermak: unfortunately I had to revert this change to pass the CI.
This is defined in newlib headers in IDF, and newlib is not supported on linux target.
I can use some system header (which works on my machine), but wouldn't work on build machine.

components/sock_utils/src/ifaddrs.c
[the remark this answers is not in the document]
david-cermak: removed

components/sock_utils/include/ifaddrs.h
david-cermak: I left the structure definition in the original shape, to keep it minimal.

components/sock_utils/src/socketpair.c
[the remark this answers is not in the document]
david-cermak: It seems like these arguments to `socketpair()` and `socket()` have slightly different meaning and we usually create socketpair for `AF_UNIX` family.

Have added a condition, that we support only this combination of family, type, protocol:
```
socketpair(AF_UNIX, SOCK_STREAM, 0, fd);
```
and return error (not-implemented) if anything else supplied

components/sock_utils/src/socketpair.c
david-cermak: I switched to blocking mode, so it's not needed anymore.
removed completely.

components/sock_utils/src/socketpair.c
david-cermak: Fixed to close the unwanted directions:
```cpp
    // Close the unwanted ends to make it a unidirectional pipe
    if (shutdown(pipefd[0], SHUT_WR) == -1 || shutdown(pipefd[1], SHUT_RD) == -1) {
        close(pipefd[0]);
        close(pipefd[1]);
        return -1;
    }
```
(and this works in lwip)

david-cermak: both pipe and socketpair is blocking by default.
Have fixed.

components/sock_utils/src/getnameinfo.c
david-cermak: Yes, this is the limitation of most functions here, as documented in the README

components/sock_utils/src/ifaddrs.c
david-cermak: fixed to use the correct type

components/sock_utils/test/host/CMakeLists.txt
david-cermak: fixed

components/sock_utils/test/host/main/test_sock_utils.cpp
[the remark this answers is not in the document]
david-cermak: changed to
```cpp
sock_addr.sin_port = htons(TEST_PORT_NUMBER);
```
as sock_addr fields are in network byte order

components/sock_utils/test/host/main/test_sock_utils.cpp
david-cermak: Added a note in the component README file. (`CONFIG_LWIP_TCP_SND_BUF_DEFAULT` which is 4 x MSS by default in IDF)

david-cermak:
> Name of the component is misleading, consider:
> 
> * porting
> * linux
>   help people find this component

other ideas (since we have the **linux** target):
* `net_linux_compat`
* `linux_sock_compat`

espressif-abhikroy (approved):

alisitsyn (approved):
LGTM.

kostaond (approved):
LGTM!

Issue #699, opened 2024-11-20
Is there a way to send AT commands while keeping the connection, without using CMUX? (IDFGH-14106)


doragasu:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I'm using a Quectel MC60 modem. In CMUX mode, I can get Internet connectivity and simultaneously send AT commands (I need to periodically get signal quality). The problem is that now I am optimizing power consumption, and the modem refuses to enter sleep mode if CMUX is active, so I am forced to use DATA mode.

While in DATA mode, sending commands to the modem does not work, and if I momentarily try switching to COMMAND mode, the connection is immediately dropped. When many years ago I made my own modem drivers, I was able to momentarily switch to command mode (using the "+++" escape sequence), send the AT command, and return to DATA mode without interrupting the connection. Is this possible in any way?

TL;DR: Is there a way to get the signal quality (or send any other AT command) when using DATA mode, without interrupting the connectivity?

david-cermak:
`esp_modem` doesn't support it ATM, but I find it quite helpful (also used it to perform OTA in the past).
I've put up a sketch PR: https://github.com/espressif/esp-protocols/pull/700, which allows you to pause a network connection and bring up the AT interface.

doragasu:
Thanks!

With that patch, invoking `pause(true)` also switches to command mode, right? Or should I manually switch after the pause?

david-cermak:
> With that patch, invoking pause(true) also switches to command mode, right? Or should I manually switch after the pause?

It sends the corresponding AT commands, yes, but the mode stays the same.

To test it, you need to switch to the data mode first. Then:
* dce->pause(true);
* dce->send_some_at(...);
* dce->pause(false);
and the network should be still functional (mode stays the same `modem_mode::DATA_MODE`).
(network data are discarded so it may happen that the connection get's dropped or disconnected in the meantime, e.g. if you're using TPC keepalive or MQTT pings...)

doragasu:
I see, thanks.

[...]

PR #700, opened 2024-11-21
[modem]: Add support for pausing netif

david-cermak:
* allows to temporarily pause the network to supply AT commands
* allows to resume the network
* Closes https://github.com/espressif/esp-protocols/issues/699


components/esp_modem/Kconfig
```diff
@@ -76,4 +76,11 @@ menu "esp-modem"
         help
             If enabled, APIs to add URC handler are available
 
+    config ESP_MODEM_ADD_DEBUG_LOGS
```
euripedesrocha, line 79: We could just document the tags and set the `ESP_LOG_BUFFER_HEXDUMP` to verbose level. User would only need to set the tag level to verbose in their code, making this config option unnecessary.

components/esp_modem/src/esp_modem_netif.cpp
```diff
@@ -99,6 +99,19 @@ void Netif::stop()
     signal.clear(PPP_STARTED);
 }
 
+void Netif::pause(bool do_pause)
```
euripedesrocha, line 102: While this is a simple API it could be split in 2. And we have a return path with a boolean but the function signature states it return void. 

I would suggest split it in `Netif::start`, `Netif::pause` and maybe provide a `Netif::stop`

components/esp_modem/Kconfig
david-cermak: Make sense, but I'm actually using these logs very often when supporting users with some new/unsupported device and update the library based on shared logs.

we could make it verbose, but the important logs would be just diluted in tons of other unrelated messages,
we could ask users to add log filtering in their applications, but that would again complicate things.

this is just a simple option they could turn on and share more-less compact logs.

components/esp_modem/src/esp_modem_netif.cpp
david-cermak: > I would suggest split it in Netif::start, Netif::pause and maybe provide a Netif::stop

we already have start/stop.

> And we have a return path with a boolean

we don't. it was a return path from the lambda.

I have split the function into `pause()` and `unpause()` to make it clear. PTAL again.

components/esp_modem/include/cxx_include/esp_modem_netif.hpp
```diff
@@ -54,6 +54,16 @@ class Netif {
...
 
+    /**
+     * @brief Pause the network interface
+     */
+    void pause();
+
+    /**
+     * @brief Resume the network interface
+     */
+    void unpause();
```
euripedesrocha, line 65: Resume is a better name.

euripedesrocha (approved):

PR #702, opened 2024-11-28
[modem]: Refine data -> command transition

david-cermak:
* Swaps `netif.stop()` with setting a callback (to avoid missing the closure)
* Adds an escape sequence after PPP exits
* Fixup retry command mode to add a newline delimiter before issuing a textual command


components/esp_modem/include/cxx_include/esp_modem_dce_module.hpp
```diff
@@ -77,7 +77,9 @@ class GenericModule: public ModuleIf {
                 if (set_command_mode() == command_result::OK) {
                     return true;
                 }
-                Task::Delay(1000); // Mandatory 1s pause after escape
+                // send a newline to delimit the escape from the upcoming sync command
```
euripedesrocha, line 80: With this the delay is unnecessary?

PR #704, opened 2024-12-02
fix(websocket): release client-lock during WEBSOCKET_EVENT_DATA (IDFGH-14545)

### LINKED ISSUE #625: WebSocket may cause a failure while the sempaphore lock in esp_websocket_client_task is in place and esp_websocket_client_send_text is used elsewhere (IDFGH-13387)
### Answers checklist.

- [X] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [X] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

The WebSocket has a problematic lock mechanism that hinders concurrent usage and the dispatch of receiving and sending data. While receiving data packets in a stream and simultaneously sending data, the lock in the esp_websocket_client_task at line:

`if (xSemaphoreTakeRecursive(client->lock, lock_timeout) != pdPASS) {`

can take an excessive amount of time to release, sometimes exceeding 10 seconds. This delay can cause the send operation to fail, which is not aligned with resilient transmission practices.

It appears that the task suffers from a circular non-delay issue, causing it to run continuously without interruption. In a production environment, there should be pauses within the semaphore lock cycle to enable proper multitasking and ensure that user-side operations can function correctly.

bryghtlabs-richard:
## Description

Prevent deadlocks when reserving locks in WEBSOCKET_EVENT_DATA handler, and lock is held by another thread sending a websocket message.

Fix high latency caused by writers serialized with WEBSOCKET_EVENT_DATA while calling esp_websocket_client_send(), even when TCP window has enough space for the entire message being queued to send.

Multiple writers are still serialized at fragment boundaries, but only with other writers and websocket error updates.

## Related

Fixes #625.

## Testing

I tested with a WebSocket echo server running on the same WiFi access point, using the target-example with the following modifications:

* AppMain thread tries to send 20 unfragmented frames, waiting 100ms between frames.
* WebSocket client is patched to record the time taken to send each frame.
* WEBSOCKET_EVENT_DATA handler calls `vTaskDelay(configTICK_RATE_HZ);` to simulate application-processing.

Without this patch, the AppMain thread experiences long delays, often either almost 1 second, and sometimes much longer(like 6 seconds) depending on which cores the AppMain and WebSocketClient thread run. The AppMain thread is unable to send at 10Hz, even though the TCP send buffer is never full.

With this patch, the AppMain thread experiences no delays writing websocket frames, and is able to send at 10Hz, as long as the TCP send buffer is not filled.

Examples logs:
[test_before_fix.txt](https://github.com/user-attachments/files/17984140/test_before_fix.txt)
[test_before_fix2.txt](https://github.com/user-attachments/files/17984184/test_before_fix2.txt)
[test_with_fix.txt](https://github.com/user-attachments/files/17984139/test_with_fix.txt)

## Checklist

- [X] Documentation is updated as needed.
- [X] Code is well-commented, especially in complex areas.
- [X] Git history is clean — commits are squashed to the minimum necessary.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1071,12 +1071,6 @@ static void esp_websocket_client_task(void *pv)
                 break;
             }
             client->ping_tick_ms = _tick_get_ms();
-
-            if (esp_websocket_client_recv(client) == ESP_FAIL) {
-                ESP_LOGE(TAG, "Error receive data");
-                esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
-                break;
-            }
             break;
         case WEBSOCKET_STATE_WAIT_TIMEOUT:
 
@@ -1113,6 +1107,13 @@ static void esp_websocket_client_task(void *pv)
                 xSemaphoreTakeRecursive(client->lock, lock_timeout);
                 esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
                 xSemaphoreGiveRecursive(client->lock);
+            } else if (read_select > 0) {
+                if (esp_websocket_client_recv(client) == ESP_FAIL) {
+                    ESP_LOGE(TAG, "Error receive data");
+                    xSemaphoreTakeRecursive(client->lock, lock_timeout);
+                    esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
+                    xSemaphoreGiveRecursive(client->lock);
+                }
             }
         } else if (WEBSOCKET_STATE_WAIT_TIMEOUT == client->state) {
             // waiting for reconnecting...


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1086,6 +1080,14 @@ static void esp_websocket_client_task(void *pv)
                 xSemaphoreTakeRecursive(client->lock, lock_timeout);
                 esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
                 xSemaphoreGiveRecursive(client->lock);
+            } else if (read_select > 0) {
+                if (esp_websocket_client_recv(client) == ESP_FAIL) {
+                    ESP_LOGE(TAG, "Error receive data");
+                    xSemaphoreTakeRecursive(client->lock, lock_timeout);
+                    esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
+                    xSemaphoreGiveRecursive(client->lock);
+                    break;
```
bryghtlabs-richard, line 1089: this `break` is probably wrong

thakurpaudel (approved):

erwinschrodinger1:
There was similar issue when I tried the esp_websocket_client. This PR commit solves the issue of sending in the socket but I am not being able to get the message response from the server. 
image
The message data case is never called.

erwinschrodinger1:
Hello @bryghtlabs-richard. I tried using this after you did the another push. Then the same issue of unable to lock resembles which was solved after the first push. 
image

Thank you.

bryghtlabs-richard:
@erwinschrodinger1, if you try to send WS data with `timeout=0` while the WS client is locked by the receive thread, I believe `xSemaphoreTakeRecursive(client->lock, timeout=0)` will immediately fail. You might have better luck with a small, non-zero timeout. It is normal for the WS client receive thread to lock the `client->lock` for short periods while processing.

erwinschrodinger1:
@bryghtlabs-richard  how should I solve my error then? Use a small delay after sending message each time?

bryghtlabs-richard:
@erwinschrodinger1 , I wouldn't add a separate delay, but I would try a small timeout, maybe around 100ms.

If a thread tries to send while the client is locked, `xSemaphoreTakeRecursive()` will delay up to the timeout, and if the client is unlocked sooner then the sending thread can continue sooner.

If a thread tries to send while the client is unlocked, the sender can lock/send/unlock without delay.

erwinschrodinger1:
Sorry initially I mistook ticks with ms. Even after adding the timeout of around 10 second. The ws-client is not being able to lock. What can be the issue in it? 
image

bryghtlabs-richard:
I would guess the client is either not setup properly, or lock is taken but not released. I can't think of anything that would take 10 seconds. Its possible a thread has locked it and forgotten to unlock it, but I don't see how this change would do it.

The next thing I'd want to know - is `client->lock` locked? If so, by which line of code?

erwinschrodinger1:
Thank you so much @bryghtlabs-richard.

the issue was done by my handler which locked the WEBSOCKET_DATA event resulting in lock of client. It is was just me and my idiotic acts. Sorry.

bryghtlabs-richard:
No worries @erwinschrodinger1 , thanks for letting us know the cause was unrelated to this patchset.

glmfe (approved):
LGTM. @bryghtlabs-richard Thanks for the contribution!

PR #705, opened 2024-12-04
[mdns]: Workaround to bypass new kconfiglib 2.4.0

david-cermak:


david-cermak:
No needed anymore, the workaround has been deployed

PR #713, opened 2024-12-16
[sock-utils]: Bump 0.1 -> 0.2

david-cermak:
### Features

- Declare socketpar and gai-error via standard headers ([7f103572](https://github.com/espressif/esp-protocols/commit/7f103572))
- Add support for gethostname() ([f7c0b756](https://github.com/espressif/esp-protocols/commit/f7c0b756))

### DIFF
--- a/components/sock_utils/CHANGELOG.md
+++ b/components/sock_utils/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.2.0](https://github.com/espressif/esp-protocols/commits/sock_utils-v0.2.0)
+
+### Features
+
+- Declare socketpair and gai_strerror via standard headers ([b090a3cb](https://github.com/espressif/esp-protocols/commit/b090a3cb))
+- Add support for gethostname() ([f7c0b756](https://github.com/espressif/esp-protocols/commit/f7c0b756))
+
 ## [0.1.0](https://github.com/espressif/esp-protocols/commits/sock_utils-v0.1.0)
 
 ### Features
--- a/components/sock_utils/idf_component.yml
+++ b/components/sock_utils/idf_component.yml
@@ -1,4 +1,4 @@
-version: 0.1.0
+version: 0.2.0
 description: The component provides helper implementation of common system/socket utilities
 url: https://github.com/espressif/esp-protocols/tree/master/components/sock_utils
 dependencies:
--- a/components/sock_utils/.cz.yaml
+++ b/components/sock_utils/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(sockutls): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py sock_utils
   tag_format: sock_utils-v$version
-  version: 0.1.0
+  version: 0.2.0
   version_files:
   - idf_component.yml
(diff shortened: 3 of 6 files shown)


components/sock_utils/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: 0.1.0
+version: 0.2.0
 description: The component provides helper implementation of common system/socket utilities
 url: https://github.com/espressif/esp-protocols/tree/master/components/sock_utils
 dependencies:
```
igrr, line 4: Commenting on the lines below: looks like the dependency should now be on IDF >= 5.5, since IDF 5.0 doesn't know about LWIP_SOCKET_HAS_SOCKETPAIR.

david-cermak: Added additional notes to the component README:

**Notes**:

- **`*)`** `socketpair()` and `pipe()` are built on top of `lwIP` TCP sockets, inheriting the same characteristics. For instance, the maximum transmit buffer size is based on the `TCP_SND_BUF` setting.
- **`**)`** `socketpair()` and `gai_strerror()` are declared in sock_utils header files, the declaration is propagated to ESP-IDF from v5.5 to the official header files. If you're using older IDF version, you need to manually pre-include related header files from the sock_utils public include directory.
- **`***)`** `pipe()` is declared in compiler's `sys/unistd.h`.

PTAL again @igrr if it addresses your concern

igrr: > I mean there is a certain value of using this component even if LWIP_SOCKET_HAS_SOCKETPAIR isn't propagated.

Okay, I agree with that. I think with the notes added in the README this does address my comment, thank you!

igrr: Maybe just one nitpick,

> declared in sock_utils header files, the declaration is propagated to ESP-IDF from v5.5 to the official header files

could you mention the names of the header files for easier searching?

david-cermak: they're listed in table itself, last column: `declared in` (this is only the note below the text)

https://github.com/espressif/esp-protocols/blob/5dc38a88ad73807a7ae9b0c586116f1c8ebf10fc/components/sock_utils/README.md?plain=1#L12

igrr: Right, I'm sorry, I have overlooked that part. Everything is great now.

components/sock_utils/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.2.0](https://github.com/espressif/esp-protocols/commits/sock_utils-v0.2.0)
+
+### Features
+
+- Declare socketpar and gai-error via standard headers ([7f103572](https://github.com/espressif/esp-protocols/commit/7f103572))
```
euripedesrocha, line 7:
```suggestion
- Declare socketpair and gai-error via standard headers ([7f103572](https://github.com/espressif/esp-protocols/commit/7f103572))
```

euripedesrocha (approved):

components/sock_utils/CHANGELOG.md
david-cermak: Thanks for noticing, fixed separately (since this is a generated file).

PR #719, opened 2024-12-20
[mosq]:  Bump: v2.0.20 -> 2.0.20~1

david-cermak:
* Added missing API docs
* Replaced local port compat headers/sources and used `sock_utils` instead
* Added a CI job to perform some consistency checks:
  - all version numbers are the same
  - generated api.md is up to date


david-cermak:
Another reason is that the version 2.0.20 failed to upload correctly to the registry for some reason: https://components.espressif.com/components/espressif/mosquitto/versions/2.0.20?language=en 
(the package is corrupted and cannot be downloaded)
--
The plan is to yank the damaged version and use `2.0.20~1` instead.

components/mosquitto/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [2.0.20~1](https://github.com/espressif/esp-protocols/commits/mosq-v2.0.20_1)
```
kostaond, line 3: You also removed multiple files. Is it something users should be informed about?

kostaond (approved):
Just one minor comment, otherwise LGTM!

components/mosquitto/CHANGELOG.md
david-cermak: Yes, you're right. I removed those files after the bump commit so it didn't appear in the changelog.

Updated, Thanks!

PR #720, opened 2024-12-20
[sock-utils] Fix potential macro conflict

david-cermak:
Some macros could be defined already in standard  headers (depending on IDF and compiler version).
Need to pre-include the official headers (since the macros are defined unconditionally).

### DIFF
--- a/components/sock_utils/include/netdb_macros.h
+++ b/components/sock_utils/include/netdb_macros.h
@@ -5,6 +5,17 @@
  */
 #pragma once
 
+#include "sdkconfig.h"
+#ifndef CONFIG_IDF_TARGET_LINUX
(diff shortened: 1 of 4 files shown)


kostaond (approved):
LGTM!

components/sock_utils/include/netdb_macros.h
```diff
@@ -5,6 +5,16 @@
  */
 #pragma once
 
+#ifndef CONFIG_IDF_TARGET_LINUX
```
igrr, line 8: Do we need to `#include <sdkconfig.h>` before this check?

david-cermak: yes, we do! Thanks for catching this!

PR #727, opened 2025-01-07
fix(sockutls): Fix gai_strerror() impl to return const string

david-cermak:
Previous gai_strerror() returned numeric representation of the code, but used TLS storage, which might cause issues with stack sizes on all tasks in the system. Alternatively we can leave the storage to static only (which wouldn't be thread-safe) or we could one-time allocate and never free (which is wrong).

This option uses hardcoded strings for common error codes used in lwip. The disadvantage is that we might need to update the impl in the future when lwip adds more codes.


components/sock_utils/src/gai_strerror.c
```diff
@@ -1,17 +1,36 @@
...
 #include <stdio.h>
 #include "gai_strerror.h"
+#include "lwip/netdb.h"
 
-_Thread_local char gai_strerror_string[32];
+#define HANDLE_GAI_ERROR(code) \
+    case code: return #code;
 
-const char *gai_strerror(int ecode)
+const char *gai_strerror(int errcode)
```
kostaond, line 13: Function definition is still `const char *gai_strerror(int ecode);` in header file.

kostaond (approved):
LGTM!

components/sock_utils/src/gai_strerror.c
david-cermak: Thanks for noticing the inconsistency. Fixed the prototype in headers.

PR #730, opened 2025-01-10
[mdns]: Fixed some minor bugs

david-cermak:
## Github issues

* 9270a1ae fix(mdns): Move MDNS_NAME_BUF_LEN to public headers
* 77532f98 fix(mdns): Cleanup includes in mdns.c
* 5eddc9c8 fix(mdns): Allow advertizing service with port==0
* 980f3d36 fix(mdns): Fixed complier warning if MDNS_MAX_SERVICES==0

## Coverity issues

* a69123de fix(mdns): Fixed potential out-of-bound interface error
* 80a7b1fd fix(mdns): Fixed incorrect error conversion
* 411dee47 fix(mdns): Fixed potential overflow when allocating txt data

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2022 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -10,6 +10,7 @@
 extern "C" {
 #endif
 
+#include "sdkconfig.h"
 #include <esp_netif.h>
 
 #define MDNS_TYPE_A                 0x0001
@@ -20,6 +21,13 @@ extern "C" {
 #define MDNS_TYPE_OPT               0x0029
 #define MDNS_TYPE_NSEC              0x002F
 #define MDNS_TYPE_ANY               0x00FF
+
+#if defined(CONFIG_LWIP_IPV6) && defined(CONFIG_MDNS_RESPOND_REVERSE_QUERIES)
+#define MDNS_NAME_MAX_LEN           (64+4)                  // Need to account for IPv6 reverse queries (64 char address  + ".ip6" )
+#else
+#define MDNS_NAME_MAX_LEN           64                      // Maximum string length of hostname, instance, service and proto
+#endif
+#define MDNS_NAME_BUF_LEN           (MDNS_NAME_MAX_LEN+1)   // Maximum char buffer size to hold hostname, instance, service or proto
 
 /**
  * @brief   Asynchronous query handle
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -334,6 +334,9 @@ static mdns_host_item_t *mdns_get_host_item(const char *hostname)
 
 static bool _mdns_can_add_more_services(void)
 {
+#if MDNS_MAX_SERVICES == 0
+    return false;
+#else
     mdns_srv_item_t *s = _mdns_server->services;
     uint16_t service_num = 0;
     while (s) {
@@ -343,8 +346,8 @@ static bool _mdns_can_add_more_services(void)
             return false;
         }
     }
-
     return true;
+#endif
 }
 
 esp_err_t _mdns_send_rx_action(mdns_rx_packet_t *packet)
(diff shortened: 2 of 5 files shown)


components/mdns/mdns.c
```diff
@@ -334,6 +332,9 @@ static mdns_host_item_t *mdns_get_host_item(const char *hostname)
 
 static bool _mdns_can_add_more_services(void)
```
euripedesrocha, line 333: How often services are added? The number of services could be memorized to avoid the loop.

components/mdns/include/mdns.h
```diff
@@ -21,6 +21,13 @@ extern "C" {
 #define MDNS_TYPE_NSEC              0x002F
 #define MDNS_TYPE_ANY               0x00FF
 
+#if defined(CONFIG_LWIP_IPV6) && defined(CONFIG_MDNS_RESPOND_REVERSE_QUERIES)
```
euripedesrocha, line 24: Why make this public?

euripedesrocha (approved):

components/mdns/mdns.c
david-cermak: Usually once after the startup, but adding services also means certain time of probing and advertizing (about 6 seconds), so I would skip this optimizaiton.

components/mdns/include/mdns.h
david-cermak: since `mdns_hostname_get()` lets user to allocate a buffer this maximum size

components/mdns/include/mdns.h
```diff
@@ -21,6 +21,13 @@ extern "C" {
 #define MDNS_TYPE_NSEC              0x002F
 #define MDNS_TYPE_ANY               0x00FF
```
david-cermak, line 23: need `#include "sdkconfig.h"`

components/mdns/mdns.c
zwx1995esp, line 333: If return false, could we print some logs for the user to let them know the configuration `MDNS_MAX_SERVICES` is insufficient to add more services? For now, in the function [mdns_service_add_for_host](https://github.com/espressif/esp-protocols/blob/master/components/mdns/mdns.c#L5904) we will leave an error log for the user `Cannot add more services` and return an error `NO_MEMORY`. The user might see, `Oh, memory might be used up`. But if they print the heap usage debugging log, there might be a lot memory left. It might comfuse the user. Some issues from the user for asking this: https://github.com/espressif/esp-thread-br/issues/115

zwx1995esp:
Hi, @david-cermak  Another topic: I noticed that mDNS creates [a non-recursive mutex](https://github.com/espressif/esp-protocols/blob/master/components/mdns/mdns.c#L5405) and this kind of mutex can only be taken once by one task refer to [FreeRTOS](https://www.freertos.org/Documentation/02-Kernel/04-API-references/10-Semaphore-and-Mutexes/08-xSemaphoreCreateRecursiveMutex): 
image

As far as I know, the APIs with the prefix `_mdns_xxx` are implemented without acquiring and releasing the lock, and the APIs `mdns_xxx` are implemented with the lock acquired and released. So we can not call any `mdns_xxx` in mdns task. Is this designed deliberately? If we use a non-recursive mutex, then in mdns task, we can call both `mdns_xxx` and `_mdns_xxx` APIs.

components/mdns/mdns.c
```diff
@@ -4477,10 +4479,11 @@ void mdns_preset_if_handle_system_event(void *arg, esp_event_base_t event_base,
                 case IP_EVENT_GOT_IP6: {
                     ip_event_got_ip6_t *event = (ip_event_got_ip6_t *) event_data;
                     mdns_if_t mdns_if = _mdns_get_if_from_esp_netif(event->esp_netif);
-                    if (mdns_if < MDNS_MAX_INTERFACES) {
-                        post_mdns_enable_pcb(mdns_if, MDNS_IP_PROTOCOL_V6);
-                        post_mdns_announce_pcb(mdns_if, MDNS_IP_PROTOCOL_V4);
+                    if (mdns_if >= MDNS_MAX_INTERFACES) {
+                        return;
```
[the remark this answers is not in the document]
david-cermak: No no, we don't want en error log here, since the `MDNS_MAX_INTERFACES` that the interface id is invalid, i.e. we received on event from another netif (like PPP), unrelated to mdns.

components/mdns/mdns.c
david-cermak: Good point, have updated the error message to:
```cpp
                      "Cannot add more services, please increase CONFIG_MDNS_MAX_SERVICES (%d)", CONFIG_MDNS_MAX_SERVICES);
```

david-cermak:
> Hi, @david-cermak Another topic: I noticed that mDNS creates [a non-recursive mutex](https://github.com/espressif/esp-protocols/blob/master/components/mdns/mdns.c#L5405) and this kind of mutex can only be taken once by one task refer to....

Valid point, but won't address this in this PR. I'm going to work more on mdns in the near future though, so I'll reconsider.
the general idea is use non-recursive mutex, unless necessary, which up to this point was not, but that may change in the future.

david-cermak:
Thanks for the review @euripedesrocha and @zwx1995esp , I'll go ahead and merge this as is as I think I've addressed all the comments. I'll continue in https://github.com/espressif/esp-protocols/pull/732 which depends on this one

PR #732, opened 2025-01-13
[mdns]: Fixes clang-tidy warnings

david-cermak:
## Fixes:

* https://github.com/espressif/esp-protocols/security/code-scanning/45
* https://github.com/espressif/esp-protocols/security/code-scanning/111
* https://github.com/espressif/esp-protocols/security/code-scanning/108
* https://github.com/espressif/esp-protocols/security/code-scanning/75
* https://github.com/espressif/esp-protocols/security/code-scanning/66
* https://github.com/espressif/esp-protocols/security/code-scanning/61
* https://github.com/espressif/esp-protocols/security/code-scanning/60
* https://github.com/espressif/esp-protocols/security/code-scanning/58

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -2355,6 +2355,11 @@ static void _mdns_restart_pcb(mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol
         srv_count++;
         a = a->next;
     }
+    if (srv_count == 0) {
+        // proble only IP
+        _mdns_init_pcb_probe(tcpip_if, ip_protocol, NULL, 0, true);
+        return;
+    }
     mdns_srv_item_t *services[srv_count];
     size_t i = 0;
     a = _mdns_server->services;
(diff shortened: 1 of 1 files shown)


david-cermak:
@zwx1995esp PTAL, too

components/mdns/mdns.c
```diff
@@ -5346,7 +5351,7 @@ static void _mdns_service_task(void *pvParameters)
     for (;;) {
         if (_mdns_server && _mdns_server->action_queue) {
             if (xQueueReceive(_mdns_server->action_queue, &a, portMAX_DELAY) == pdTRUE) {
-                if (a && a->type == ACTION_TASK_STOP) {
+                if (a == NULL || a->type == ACTION_TASK_STOP) {
```
zwx1995esp, line 5354: Question: does `a == NULL` mean that an exception occurs? If yes, at this time, should we stop mdns timer? Should we free the lock and deinit the queue? How about asserting right now?

zwx1995esp (approved):
Just one question, LGTM overall.

components/mdns/mdns.c
david-cermak: It's definitely an invalid state, which should never happen -- i.e. typically happens during mem corrupt issues.
`assert()` is probably a better reaction then just stopping the task.
Have fixed, thanks for the suggestion!

euripedesrocha (approved):

PR #744, opened 2025-01-24
fix(mdns): Fix _mdns_append_fqdn excessive stack usage (IDFGH-14506)

### LINKED ISSUE #738: Recursion in mdns.c:_mdns_append_fqdn uses excessive stack space (IDFGH-14479)
### Answers checklist.

### General issue report

A client call to MDNS to remove a service from MDNS will send a bye packet resulting in a call to `_mdns_append_fqdn()` within the clients context. This function uses recursion to append the FDQN to the packet.
https://github.com/espressif/esp-protocols/blob/a83f1b6787f62314f2ac8f58755db916441909b6/components/mdns/mdns.c#L748
This function has some local variables in the full scope of the function. one of these happens to be a significantly large structure `mdns_name_t name`. The result is that this structure of ~300 bytes is allocated on the clients stack for every recursive iteration. I have seen at lease 4 levels of recursion in practise, which is a substantial burden on the calling task and rather wasteful or memory in an embedded environment.

As far as I can tell the each `mdns_name_t` structure is only used (and not always) within the while loop of the function and is no longer needed by the recursion call. Unfortunately because the variable is function scoped the compiler cannot tell that it is no longer needed and pop it from the stack before making the recursion call.

I suggest moving the `mdns_name_t name` declaration inside the loop so that it is better scoped and the stack space can be reclaimed.
This will significantly reduce the overall stack requirements for the function.
Here is a suggested patch:
```
diff -ur a/managed_components/espressif__mdns/mdns.c b/managed_components/espressif__mdns/mdns.c
--- q/managed_components/espressif__mdns/mdns.c	2025-01-21 11:07:03.107296749 +1030
+++ b/managed_components/espressif__mdns/mdns.c	2025-01-21 11:07:18.760049319 +1030
@@ -749,12 +749,12 @@
         //empty string so terminate
         return _mdns_append_u8(packet, index, 0);
     }
-    mdns_name_t name;
     static char buf[MDNS_NAME_BUF_LEN];
     uint8_t len = strlen(strings[0]);
     //try to find first the string length in the packet (if it exists)
     uint8_t *len_location = (uint8_t *)memchr(packet, (char)len, *index);
     while (len_location) {
+        mdns_name_t name;
         //check if the string after len_location is the string that we are looking for
         if (memcmp(len_location + 1, strings[0], len)) { //not continuing with our string
 search_next:
```
I do not have an specifics, but I did notice at least one other uses of recursion. So there may be other opportunities for stack usage reductions.

andrew-lifx:
Move "mdns_name_t name" declaration inside loop to move it out of scope and reclain stack space before recursion call.

## Description
Move "mdns_name_t name" declaration inside loop to move it out of scope and reclaim stack space before the recursion call.

## Related

Fixes #738 

## Testing

Tested using patched version on product build. Original was triggering a stack overflow. Stack overflow did not occur with patched version.

## Checklist

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -751,12 +751,12 @@ static uint16_t _mdns_append_fqdn(uint8_t *packet, uint16_t *index, const char *
         //empty string so terminate
         return _mdns_append_u8(packet, index, 0);
     }
-    mdns_name_t name;
     static char buf[MDNS_NAME_BUF_LEN];
     uint8_t len = strlen(strings[0]);
     //try to find first the string length in the packet (if it exists)
     uint8_t *len_location = (uint8_t *)memchr(packet, (char)len, *index);
     while (len_location) {
+        mdns_name_t name;
         //check if the string after len_location is the string that we are looking for
         if (memcmp(len_location + 1, strings[0], len)) { //not continuing with our string
 search_next:


david-cermak (approved):
Thanks for the PR!

All is good, just need to click through the CLA agreement in https://github.com/espressif/esp-protocols/pull/744#issuecomment-2611772801

PR #751, opened 2025-01-28
fix(websocket): wait for task on destroy (IDFGH-14533)

johanstokking:
## Description

This ensures that the task is awaited on destroy.

This fixes a race condition between two independent actions:

1. The WebSocket task detects a disconnect (e.g. read fails), initiates abort, sets `run` to false, posts the `DISCONNECTED` event, breaks, then posts the `FINISHED` events and deletes the task
2. The application destroys the WebSocket client (e.g. via an asynchronous invocation via receiving the disconnected event)

Currently, in `esp_websocket_client_destroy()` would immediately destroy all resources, because `run` is already `false`. However, the task may still be running, and it needs the resources to cleanly stop the task (e.g. post the `FINISHED` event).

This change ensures that the task has stopped. The task must be stopped before resources can be freed.

The `run` check in the destroyer is also racy, so the wait shouldn't happen in an `else`.

## Testing

I tested this locally.

## Checklist

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -449,6 +449,21 @@ static void destroy_and_free_resources(esp_websocket_client_handle_t client)
     client = NULL;
 }
 
+static esp_err_t stop_wait_task(esp_websocket_client_handle_t client)
+{
+    /* A running client cannot be stopped from the websocket task/event handler */
+    TaskHandle_t running_task = xTaskGetCurrentTaskHandle();
+    if (running_task == client->task_handle) {
+        ESP_LOGE(TAG, "Client cannot be stopped from websocket task");
+        return ESP_FAIL;
+    }
+
+    client->run = false;
+    xEventGroupWaitBits(client->status_bits, STOPPED_BIT, false, true, portMAX_DELAY);
+    client->state = WEBSOCKET_STATE_UNKNOW;
+    return ESP_OK;
+}
+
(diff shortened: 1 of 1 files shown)


johanstokking:
@gabsuren thanks for considering this one too

david-cermak (approved):
Nice, thanks for the fix!

So the real issue is actually using `(client->run)` to determine if the client is still running.
And we do the same thing in the `ws_client_stop()`, correct?
maybe checking `status_bits` for `STOPPED_BIT` would solve the issue, too, and we could even reduce some code duplication and do:
```cpp
esp_err_t esp_websocket_client_stop(esp_websocket_client_handle_t client)
{
    return check_and_stop_internal(client);
}

esp_err_t esp_websocket_client_destroy(esp_websocket_client_handle_t client)
{
    esp_err_t err = check_and_stop_internal(client);
    // ignore ESP_FAIL for some "already stopped" case
    destroy_and_free_resources(client);
    return err;
}
```
just an idea on possible improvement, but your current changes in this PR LGTM!

david-cermak (changes requested):
need to fix unit tests per: https://github.com/espressif/esp-protocols/actions/runs/13029806393/job/36348129127

this change broke the basic init-destroy usecase (since `STOPPED_BIT` is not set upon initialization),
and also, destroy is called on any unsuccessful init (so need to take into account that the event group might not be initialized yet)

david-cermak (approved):
Thanks for the updates!

PR #752, opened 2025-01-30
feat(websocket): add ability to reconnect after close (IDFGH-15455)

bryghtlabs-richard:
## Description

Add a flag to client configuration to enable reconnection after websocket closed cleanly. Needed to support protocols like Socket.IO, where some websocket closes issued by the server are reconnectable.

## Testing

Testing consists of issuing a websocket close from the server, and the client reconnecting.

## Checklist

- [X] 🚨 This PR does not introduce breaking changes.
- [X] Documentation is updated as needed.
- [X] Git history is clean — commits are squashed to the minimum necessary.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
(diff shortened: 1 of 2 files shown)


bryghtlabs-richard:
Thanks @euripedesrocha , I agree - a failure log in the console would be helpful if we ever cannot take the semaphore.

I also think we should move sending the `WEBSOCKET_EVENT_CLOSED` so it doesn't happen within the semaphore. Sending `WEBSOCKET_EVENT_DATA` with the semaphore held had caused a lot of deadlocks for me, and I worry about `WEBSOCKET_EVENT_CLOSED` also causing them less often.

Issue #754, opened 2025-02-01
mDNS resolution fails when requesting SRV and TXT records (IDFGH-14556)


tl-sl:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Recently in Home Assistant there have been issues when resolving mdns addresses of devices running on esp32.  After some debugging it seems that when requesting `SRV` and `TXT` records along with `A` and `AAAA` from `name.local` the devices fail to respond at all. I believe this is the default in Zeroconf. However if requesting only `A` and `AAAA` record then do correctly get a response.

This affects Arduino v3, but also tested on esp-idf v5.4 with latest mDNS component and using the query_advertise example which has same issue.

Here is a test script that reproduces the issues, running this script no answer is ever received.

```
import asyncio
import logging

from zeroconf import (
    DNSOutgoing,
    DNSQuestion,
)
from zeroconf.asyncio import AsyncZeroconf
from zeroconf.const import (
    _CLASS_IN,
    _FLAGS_QR_QUERY,
    _TYPE_A,
    _TYPE_AAAA,
    _TYPE_ANY,
    _TYPE_TXT,
    _TYPE_SRV,
)

logging.basicConfig(level=logging.DEBUG)
logging.getLogger("zeroconf").setLevel(logging.DEBUG)
name = "esp32-mdns.local."
lower_name = name.lower()


async def test_send_a_aaaa_query_lowercase_a_with_srv_txt():
    """Send a query with the A record type with the lowercase name."""
    aiozc = AsyncZeroconf()
    await aiozc.zeroconf.async_wait_for_start()
    outgoing = DNSOutgoing(_FLAGS_QR_QUERY)
    outgoing.add_question(DNSQuestion(lower_name, _TYPE_SRV, _CLASS_IN))
    outgoing.add_question(DNSQuestion(lower_name, _TYPE_TXT, _CLASS_IN))
    outgoing.add_question(DNSQuestion(lower_name, _TYPE_A, _CLASS_IN))
    outgoing.add_question(DNSQuestion(lower_name, _TYPE_AAAA, _CLASS_IN))
    aiozc.zeroconf.async_send(outgoing)
    await asyncio.sleep(3)

[...]

PR #758, opened 2025-02-04
[mdns]: Fix the responder to ignore only invalid queries

david-cermak:
not the entire packet, so we can still reply to next questions

Closes https://github.com/espressif/esp-protocols/issues/754


david-cermak:
@zwx1995esp PTAL, as well

euripedesrocha (approved):

bdraco:
Thanks for solving this 🥇

Issue #759, opened 2025-02-10
slip_custom_netif  demo not working


nart-schinackow-munoz:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Hi,

I am trying the slip_custom_netif and my esp32s3 keeps rebooting.
My Idea is to use the esp32 connected to a Linux pc via Serial using the "slattach -v -L -s 115200 -p slip /dev/ttyUSB1"

below is the log:

ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x2a (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce2820,len:0x1b34
load:0x403c8700,len:0x4
load:0x403c8704,len:0xfbc
load:0x403cb700,len:0x3820
entry 0x403c89e4
I (27) boot: ESP-IDF v5.3.2-dirty 2nd stage bootloader
I (27) boot: compile time Feb 10 2025 22:04:01
I (27) boot: Multicore bootloader
I (31) boot: chip revision: v0.2
I (34) boot: efuse block revision: v1.3
I (39) qio_mode: Enabling QIO for flash chip WinBond
I (44) boot.esp32s3: Boot SPI Speed : 80MHz
I (49) boot.esp32s3: SPI Mode       : QIO
I (54) boot.esp32s3: SPI Flash Size : 16MB
I (59) boot: Enabling RNG early entropy source...
I (64) boot: Partition Table:
I (68) boot: ## Label            Usage          Type ST Offset   Length
I (75) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (83) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (90) boot:  2 factory          factory app      00 00 00010000 00100000
I (98) boot: End of partition table
I (102) esp_image: segment 0: paddr=00010020 vaddr=3c040020 size=10968h ( 67944) map
I (121) esp_image: segment 1: paddr=00020990 vaddr=3fc94000 size=02ddch ( 11740) load
I (124) esp_image: segment 2: paddr=00023774 vaddr=40374000 size=0c8a4h ( 51364) load
[...]

PR #762, opened 2025-02-12
[examples]: Test the SLIP netif example on target as well

david-cermak:
## Problem

* `esp_netiff` by default uses lwip's `netif->state` to store it's internal data
* some non-typical lwip netifs also use this to store internal data (PPP, bridge, wireguard port)

## Solution

* to address this properly we need to fix `esp_netif` to either keep this reserved or add a Kconfig option to set `NETIF_DATA` -- will fix in IDF-12473
* here we provide a workaround to enable PPP netif to force `esp_netif` not to use the state field
* also adds target test

* Closes https://github.com/espressif/esp-protocols/issues/759


.github/workflows/examples_build-host-test.yml
```diff
@@ -51,3 +64,38 @@ jobs:
...
+  run_on_target:
+    # Skip running on forks since it won't have access to secrets
+    if: |
+      github.repository == 'espressif/esp-protocols' &&
+      ( contains(github.event.pull_request.labels.*.name, 'examples') || github.event_name == 'push' )
+    name: Slip example target test
+    needs: build_all_examples
+    strategy:
+      matrix:
+        idf_ver: ["release-v5.3", "latest"]
```
david-cermak, line 77: run on v5.4, too
```suggestion
        idf_ver: ["release-v5.3", "release-v5.4", "latest"]
```

euripedesrocha (approved):

.github/workflows/examples_build-host-test.yml
david-cermak: * added `v5.4` to build jobs
* replaced `v5.3` -> `v5.4` in run jobs.
(so we're building on all versions, but running only on latest)

PR #764, opened 2025-02-13
[mdns]: Allow allocate memory with configured caps

david-cermak:
Addresses IDF-11680: Add a config and a mechanism to support allocating memory from PSRAM for mDNS service.


david-cermak:
@zwx1995esp PTAL, not sure how to approach coexistence of `MDNS_TASK_MEMORY_ALLOC_FROM` and `MDNS_MEMORY_ALLOC_FROM` or define a dependency between?
so far added them as two separate configs, but open for discussion.

components/mdns/CMakeLists.txt
```diff
@@ -10,15 +10,22 @@ else()
     set(MDNS_CONSOLE "")
 endif()
 
+if (CONFIG_MDNS_MEMORY_ALLOC_CUSTOM_IMPL)
+    # User has to provide their own implementation of memory allocation functions
+    set(MDNS_MEMORY "")
```
zwx1995esp, line 15: Nits: Do we need an additional configuration option to specify the directory path for the user's implementation file? Users might face difficulties adding their files to the mDNS CMakeLists, as modifying the downloaded component is not allowed.

zwx1995esp (approved):
LGTM, Thanks a lot for such quick support.

zwx1995esp:
> Not sure how to approach coexistence of `MDNS_TASK_MEMORY_ALLOC_FROM` and `MDNS_MEMORY_ALLOC_FROM` or define a dependency between? so far added them as two separate configs, but open for discussion.

Hi, David, I think we should keep both configurations for the user. If the task is created in internal memory, it can save some time during system task switching.

components/mdns/CMakeLists.txt
david-cermak: > Do we need an additional configuration option to specify the directory path for the user's implementation file?

Not really, the idea here is to provide an implementation in an external component (like `main` or some other user side component) and set a link dependency to mdns (which can be done through project makefile)

But after thinking about it again, maybe an easier option would be to revert this config and just mark the mdns memory functions as `__weak`, WYT ?

zwx1995esp: Ok, agree with marking the mdns memory functions as `__weak` if macro `CONFIG_MDNS_MEMORY_ALLOC_CUSTOM_IMPL` is enabled.

david-cermak: ok, made the memory functions optionally `weak`.
also allowing to rewrite the task allocators, which is useful for setting some static reserved space for the task stack.

components/mdns/mdns_mem_caps.c
```diff
@@ -0,0 +1,95 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <string.h>
```
zwx1995esp, line 6: Nit: please inlcude the header file `sdkconfig.h` since the macro `CONFIG_XXX` is used in this file.

zwx1995esp (approved):

components/mdns/mdns_mem_caps.c
david-cermak: good point -- fixed, thanks!

PR #768, opened 2025-02-19
fix(modem): handle nullptr in DTE constructors to prevent invalid access (IDFGH-14688)

deodatomatheus:
## Description

The issue can be reproduced by setting a timeout for waiting for a USB connection (esp_modem_usb_term_config.timeou_ms). When the timeout occurs, the value of `primary_term` becomes null.

If either primary_term or secondary_term is `nullptr`, the set_command_callbacks function attempts to access invalid values. For example, in the first line of the function, `primary_term->set_read_cb([this](uint8_t *data, size_t len)`, the program tries to dereference primary_term, leading to a `LoadProhibited` exception and crashing the application.

To address this, I added a validation check in the DTE constructor to ensure the values are valid. Additionally, I handled the exception in the standard exception handler.

## Testing
I test the changes in a ESP32S3.

---

## Checklist

### DIFF
--- a/components/esp_modem/src/esp_modem_dte.cpp
+++ b/components/esp_modem/src/esp_modem_dte.cpp
@@ -14,37 +14,52 @@ using namespace esp_modem;
 
 static const size_t dte_default_buffer_size = 1000;
 
-DTE::DTE(const esp_modem_dte_config *config, std::unique_ptr<Terminal> terminal):
-    buffer(config->dte_buffer_size),
-    cmux_term(nullptr), primary_term(std::move(terminal)), secondary_term(primary_term),
-    mode(modem_mode::UNDEF)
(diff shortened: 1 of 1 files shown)


david-cermak (commented):
Thanks for the fix, LGTM!

Could you please just update the commit message prefix -> `fix(modem): ...`?
(or install the pre-commit hook)

david-cermak:
@deodatomatheus Please check also the CI failures, above,
we support a build with exception disabled, as well.

deodatomatheus:
I changed the code to use `ESP_MODEM_THROW_IF_FALSE(...) `instead of `throw std::invalid_argument(...)` to support building with exceptions disabled.

david-cermak (approved):
Thanks for the updates, LGTM!

PR #769, opened 2025-02-19
Add ESP_MDNS version keys

ascillato:
## Description

This PR aims to add version keys that can be retrieved from the firmware using this component. The version keys are needed so the firmware can tell which version of the MDNS component is using. This is required when doing firmware certification under some standards, for example [Measurement of Canada](https://ised-isde.canada.ca/site/measurement-canada/en/type-approval/apply-type-approval-electric-vehicle-supply-equipment).

The idea of the keys has been taken from other components that already have their version embedded into it like [esp_littleFS](https://github.com/joltwallet/esp_littlefs/blob/master/include/esp_littlefs.h)

I understand that adding these keys adds one more step when doing the maintenance of the component, but I hope is not a big hasle since it is benefit traceability.

## Related

None.

## Testing

```cpp
ESP_LOGI(TAG, "mDNS Ver: %s", ESP_MDNS_VERSION_NUMBER);
```

---

## Checklist


david-cermak (changes requested):
Thanks for this PR, it makes perfect sense to have a version number info in component headers.

This would however mean that we'd have to keep these in sync with the component version.

Suggest using idf cmake properties:
```cmake
...
# In mdns component makefile
idf_component_get_property(MDNS_VERSION ${COMPONENT_NAME} COMPONENT_VERSION)
target_compile_definitions(${COMPONENT_LIB} PUBLIC "-DESP_MDNS_VERSION_NUMBER=\"${MDNS_VERSION}\"")
```

ascillato:
@david-cermak 

Thanks for reviewing. Yes, that makes totally sense to automate the version key generation.

I tried to generate the KEY in the CMAKELISTS.TXT file but I could not make the key to propagate to the main() of a firmware using this MDNS component. So, what I propose instead is to let CMAKE read the version from the component registration and then writing that to the ESP_MDNS_VERSION_NUMBER key in mdns.h file.

Please, review this new modifications and let me know your thoughts.

components/mdns/CMakeLists.txt
```diff
@@ -34,7 +34,26 @@ if(${target} STREQUAL "linux")
...
+
+# Replace only the line that defines ESP_MDNS_VERSION_NUMBER.
+# The regex matches the start of the line with the #define ESP_MDNS_VERSION_NUMBER
+# followed by any characters except a newline.
+string(REGEX REPLACE "#define ESP_MDNS_VERSION_NUMBER[^\n]*"
+       "#define ESP_MDNS_VERSION_NUMBER \"${MDNS_VERSION}\""
+       PATCHED_CONTENT "${ORIGINAL_CONTENT}")
+
+# Write the patched content
+file(WRITE ${MDNS_H_FILE} "${PATCHED_CONTENT}")
```
david-cermak, line 59: Have you tried my original suggestion?
```suggestion
target_compile_definitions(${COMPONENT_LIB} PUBLIC "-DESP_MDNS_VERSION_NUMBER=\"${MDNS_VERSION}\"")
```

This should define the macro `ESP_MDNS_VERSION_NUMBER` from compiler command line. The important part is the **`PUBLIC`** keyword, meaning that this wouldn't be defined only in the component itself, but in all **public** dependencies (i.e. everything that includes `mdns.h`). So it should work from the main component.

components/mdns/include/mdns.h
```diff
@@ -13,6 +13,8 @@ extern "C" {
 #include "sdkconfig.h"
 #include <esp_netif.h>
 
+#define ESP_MDNS_VERSION_NUMBER "1.7.0"
+
```
david-cermak, line 17: need to remove this
```suggestion
```

components/mdns/CMakeLists.txt
david-cermak: > The key is not propagated to the rest of the files.

it will only be propagated to the component files which define dependency on mdns.

Does the macro expand in the `main` component for you?

> Please, advise how do you want to proceed.

Let's find out why the `target_compile_definitions()` doesn't work for you.
I don't think rewriting the mdns.h header from makefiles (on every expansion?) is acceptable.

ascillato: > Does the macro expand in the main component for you?
In my setup, the macro is not accessable from main().

> I don't think rewriting the mdns.h header from makefiles (on every expansion?) is acceptable.
No problem

ascillato:
@david-cermak

Thanks for reviewing.

I have applied your suggestions.

david-cermak (approved):
Thanks for the update, the current changes LGTM!

but please cleanup the commit history and update the commit message accordingly.
For details, read [contributing](https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md) guidelines.
(just enough to keep one commit with current change and `feat(mdns): Add version keys` as the commit message)

PR #771, opened 2025-02-21
feat(websocket): Add websocket HTTP redirect

glmfe:
## Description

 - Handle the HTTP 3xx status family and redirect the connection to the new host.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -64,6 +64,12 @@ static const char *TAG = "websocket_client";
 #define WS_OVER_TCP_SCHEME  "ws"
 #define WS_OVER_TLS_SCHEME  "wss"
 #define WS_HTTP_BASIC_AUTH  "Basic "
+#define WS_HTTP_REDIRECT(code) ((code >= 300) && (code < 400))
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
+// Features supported in 5.5.0
+#define WS_TRANSPORT_REDIRECT_HEADER_SUPPORT    1
+#endif
 
 const static int STOPPED_BIT = BIT0;
 const static int CLOSE_FRAME_SENT_BIT = BIT1;   // Indicates that a close frame was sent by the client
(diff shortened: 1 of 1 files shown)


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1024,10 +1029,33 @@ static void esp_websocket_client_task(void *pv)
...
+                                                   result, esp_err_to_name(error_handle->last_error), error_handle->esp_tls_error_code,
+                                                   error_handle->esp_tls_flags, client->error_handle.esp_ws_handshake_status_code, errno);
+                    }
+#else
                     esp_websocket_client_error(client, "esp_transport_connect() failed with %d, "
                                                "transport_error=%s, tls_error_code=%i, tls_flags=%i, esp_ws_handshake_status_code=%d, errno=%d",
                                                result, esp_err_to_name(error_handle->last_error), error_handle->esp_tls_error_code,
                                                error_handle->esp_tls_flags, client->error_handle.esp_ws_handshake_status_code, errno);
+#endif
                 } else {
```
david-cermak, line 1059: nitpick
```suggestion
                } else {
#endif
```

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1024,10 +1029,33 @@ static void esp_websocket_client_task(void *pv)
                 esp_tls_error_handle_t error_handle = esp_transport_get_error_handle(client->transport);
                 client->error_handle.esp_ws_handshake_status_code  = esp_transport_ws_get_upgrade_request_status(client->transport);
                 if (error_handle) {
+#ifdef ESP_TRANSPORT_WS_SUPPORTED_FEATURE_REDIRECT
+                    if (client->error_handle.esp_ws_handshake_status_code == WS_TRANSPORT_RESPONSE_REDIR) {
+                        client->config->port = 0;
+                        client->config->uri  = esp_transport_ws_get_redir_uri(client->transport);
+                        if (client->config->uri == NULL) {
+                            ESP_LOGE(TAG, "There are no redirect uri");
```
david-cermak, line 1037: question: will it continue with some default error or just abort the connection?

glmfe: Just changed the logic a little bit to be more clear, but we just continue with previous error approach (esp_websocket_client_error -> esp_websocket_client_abort_connection)

david-cermak (approved):
changes in this PR LGTM!

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1035,6 +1046,20 @@ static void esp_websocket_client_task(void *pv)
                 esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
                 break;
             }
+#if WS_TRANSPORT_REDIRECT_HEADER_SUPPORT
+            else if (WS_HTTP_REDIRECT(result)) {
+                // Redirecting to a new URI
+                client->config->port = 0;
+                client->config->uri  = esp_transport_ws_get_redir_uri(client->transport);
```
euripedesrocha: How long this will be valid? Shall we `strdup` here?

glmfe: refactored to a safer approach, thanks

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -598,6 +604,23 @@ static esp_err_t esp_websocket_client_create_transport(esp_websocket_client_hand
     return ESP_OK;
 }
 
+static void esp_websocket_client_prepare_transport(esp_websocket_client_handle_t client)
```
euripedesrocha: As this is an internal function there is little reason to test for valid transport as this must be validated on user facing APIs. 
The function also only set the port, I don't see the need for it. 
The name is also too generic. 

In summary I think this function should be removed.

glmfe: sure, no problem! Thanks

euripedesrocha (approved):

PR #772, opened 2025-02-24
feat(mdns): Add version keys

ascillato:
## Description

This PR aims to add version keys that can be retrieved from the firmware when using the mDNS component. The version keys are needed so the firmware can tell which version of the MDNS component is using. This is required when doing firmware certification under some standards, for example [Measurement Canada](https://ised-isde.canada.ca/site/measurement-canada/en/type-approval/apply-type-approval-electric-vehicle-supply-equipment).

Thanks for your time.

## Related

None.

## Testing

```cpp
ESP_LOGI(TAG, "mDNS Ver: %s", ESP_MDNS_VERSION_NUMBER);
```

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [x] 🚨 This PR does not introduce breaking changes.
- [x] All CI checks (GH Actions) pass.
- [x] Documentation is updated as needed.
- [x] Tests are updated or added as necessary.
- [x] Code is well-commented, especially in complex areas.
- [x] Git history is clean — commits are squashed to the minimum necessary.

PR #773, opened 2025-02-27
[mdns]: Bump 1.7.0 -> 1.8.0

david-cermak:
* Brings up https://github.com/espressif/esp-protocols/pull/772
---

### Features

- Add version keys ([e01e67e7](https://github.com/espressif/esp-protocols/commit/e01e67e7))

### Bug Fixes

- Reformat mdns sources per indent-cont=120 ([c7663cde](https://github.com/espressif/esp-protocols/commit/c7663cde))

### DIFF
--- a/components/mdns/.cz.yaml
+++ b/components/mdns/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(mdns): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py mdns
   tag_format: mdns-v$version
-  version: 1.7.0
+  version: 1.8.0
   version_files:
   - idf_component.yml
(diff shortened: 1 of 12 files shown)


components/mdns/.cz.yaml
```diff
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(mdns): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py mdns
   tag_format: mdns-v$version
-  version: 1.7.0
+  version: 1.7.1
```
euripedesrocha, line 6: Is the bumped version correct? We are adding anew feature so shouldn't it be 1.8.0?

david-cermak: yes, you're right, it should be `1.8.0` -> fixed

PR #780, opened 2025-03-14
[mdns]: Fix potential task delete race

david-cermak:
Need to wait for the task to be deleted before destroy its stack

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -5427,7 +5427,7 @@ static void _mdns_service_task(void *pvParameters)
         }
     }
     _mdns_service_task_handle = NULL;
-    vTaskDelete(NULL);
+    vTaskDelay(portMAX_DELAY);
 }
 
 static void _mdns_timer_cb(void *arg)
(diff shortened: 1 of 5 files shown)


components/mdns/mdns.c
```diff
@@ -5539,9 +5538,11 @@ static esp_err_t _mdns_service_task_stop(void)
             vTaskDelete(_mdns_service_task_handle);
             _mdns_service_task_handle = NULL;
         }
-        while (_mdns_service_task_handle) {
+        eTaskState state;
+        while ((state = eTaskGetState(_mdns_service_task_handle)) != eDeleted) {
```
kostaond, line 5542: Frankly speaking, it looks fishy... FreeRTOS doc specifies: `eDeleted (the tasks TCB is waiting to be cleaned up)`. It indicates to me that you need to call `eTaskGetState` in some precise point in time to get this value. Are we able to ensure it? What will happen when you call `eTaskGetState` and task has been already deleted? Seems like undefined behavior...

In addition, if the below condition is executed (the condition above `while`),
```c
        if (xQueueSend(_mdns_server->action_queue, &a, (TickType_t)0) != pdPASS) {
            vTaskDelete(_mdns_service_task_handle);
            _mdns_service_task_handle = NULL;
        }
``` 
you will call `eTaskGetState` with NULL reference which causes assert. That doesn't seem to be intended behavior.

kostaond: What about releasing the semaphore in the mdns task?
```c
// set task handle to NULL
// set global semaphore ref to NULL
// release semaphore
// delete semaphore
// delete task
```
However, that would be inconsistent with semaphore allocation...

david-cermak: you're right, while this change seems to work it is just wrong. I'll think of a better way.

> What about releasing the semaphore in the mdns task?

It's not about the semaphore, it's about the stack, which is allocated externally. ("by the book" solution would be to spawn a new thread to delete the task and cleanup its stack as it's done in FreeRTOS layers)

kostaond (approved):

Issue #715, opened 2024-12-17
[websocket]: ability to read connection upgrade headers


bryghtlabs-richard:
### Is your feature request related to a problem?

ESP WebSocket-client instances are unable to use cookie-based sticky sessions, because there is no interface to extract the upgrade response headers. 

Here's an example of what we'd like to use: https://docs.aws.amazon.com/prescriptive-guidance/latest/load-balancer-stickiness/alb-cookies-stickiness.html, which tries to set `lb_cookie` to remember which backend server the device connected to, but there's no way for the client to read the value of the `Set-Cookie: lb_cookie=<server-index-value>` header.

### Describe the solution you'd like.

When connecting, I'd like the ability to iterate HTTP response headers during the WEBSOCKET_EVENT_CONNECTED event handler.

Once I have this, I can save out the `lb_cookie` for use in future reconnections.

bryghtlabs-richard:
Will we need some sort of feature-flag from the ESP-IDF to know if `header_hook` is available?

glmfe:
Hi @bryghtlabs-richard ,

we might use the IDF version tag, I'm just checking whether this will be backported to 5.5.0

```
#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
//Code that refers to `header_hook`
#endif
```

PR #794, opened 2025-04-02
feat(websocket): Add ws get HTTP response headers

glmfe:
## Description
- Expose HTTP header from the ws upgrade.

## Related
- Fixes [#715](https://github.com/espressif/esp-protocols/issues/715)


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -25,6 +25,11 @@
 
 static const char *TAG = "websocket_client";
 
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
```
david-cermak, line 28: Are we going to backport this feature to v5.5 ?

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -25,6 +25,11 @@
...
+// Features supported in 5.5.0
+#define WS_TRANSPORT_STORE_RESPONSE_HEADERS    1
```
david-cermak, line 30: Maybe we should add this macro to the `ws_header` directly, so that clients know which "capabilities" the current transport has.

components/esp_websocket_client/examples/target/main/websocket_example.c
```diff
@@ -177,6 +178,8 @@ static void websocket_app_start(void)
 #if CONFIG_WS_OVER_TLS_SKIP_COMMON_NAME_CHECK
     websocket_cfg.skip_cert_common_name_check = true;
 #endif
+    websocket_cfg.response_headers = malloc(WS_HANDSHAKE_RESPONSE_HEADERS_MAX_SIZE);
```
david-cermak, line 181: Nitpick: We should `free()` the buffer at the end of the example.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -725,7 +736,8 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
     client->config->cert_common_name = config->cert_common_name;
     client->config->crt_bundle_attach = config->crt_bundle_attach;
     client->config->ext_transport = config->ext_transport;
-
+    client->config->response_headers = config->response_headers;
+    client->config->response_headers_len = config->response_headers_len;
```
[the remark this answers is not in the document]
glmfe: not needed, the current way we are setting the config str for ws and by calling esp_transport_ws_set_config will then call esp_transport_ws_set_headers (same as set_path, set_auth...)

components/esp_websocket_client/esp_websocket_client.c
glmfe: yes, probably to 5.5.1. But this was removed.

components/esp_websocket_client/esp_websocket_client.c
glmfe: Agree, I moved to ws_transport, thanks

components/esp_websocket_client/include/esp_websocket_client.h
```diff
@@ -134,6 +134,8 @@ typedef struct {
     size_t                      ping_interval_sec;          /*!< Websocket ping interval, defaults to 10 seconds if not set */
     struct ifreq                *if_name;                   /*!< The name of interface for data to go through. Use the default interface without setting */
     esp_transport_handle_t      ext_transport;              /*!< External WebSocket tcp_transport handle to the client; or if null, the client will create its own transport handle. */
+    char                        *response_headers;          /*!< WebSocket handshake response headers */
```
euripedesrocha, line 139:
```suggestion
    char                        *response_headers;          /*!< WebSocket handshake response headers buffer, user is responsible for manage lifetime. */
```

components/esp_websocket_client/esp_websocket_client.c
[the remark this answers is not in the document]
Bryght-Labs-Richard: Isn't `esp_transport_ws_set_headers ()` only for the request headers? Where do the response headers populated? I supposed I was expecting ws_connect() to be modified to support this.

components/esp_websocket_client/include/esp_websocket_client.h
vikramdattu: How is the user supposed to manage this? The allocation is done by esp_websocket_client:

```c
    // Allocate memory for response_headers if length is specified
    if (config->response_headers_len > 0) {
        client->config->response_headers = malloc(config->response_headers_len);
        ESP_WS_CLIENT_MEM_CHECK(TAG, client->config->response_headers, goto _websocket_init_fail);
        memset(client->config->response_headers, 0, config->response_headers_len);
        client->config->response_headers_len = config->response_headers_len;
    } else {
        client->config->response_headers = NULL;
        client->config->response_headers_len = 0;
    }
```
Also, how is this different from the ws->buffer at this point: https://github.com/espressif/esp-idf/blob/1bbf04cb4cf54d74c1fe21ed12dbf91eb7fb1019/components/tcp_transport/transport_ws.c#L326-L327
Can it be simply re-used (/raise and event with the pointer and the response_len)?

Bryght-Labs-Richard: >Also, how is this different from the ws->buffer at this point...Can it be simply re-used (/raise and event with the pointer and the response_len)?

You and I had the same thought :) There's [an implementation of this approach here](https://github.com/espressif/esp-idf/pull/16119), feel free to help review!

glmfe: @vikramdattu, Just to clarify, the comment was made before I pushed the latest changes. At that point, the approach required users to manage the buffer's lifetime. 
As Richard mentioned, we're now reviewing a different approach for this ticket.

PR #811, opened 2025-05-07
fix(websocket): fix setup logic in the external transport (IDFGH-15250)

huming2207:
## Description

Hi all,

I made a dumb mistake when I was submitting the PR: https://github.com/espressif/esp-protocols/pull/573

I somehow forgot two branch logic when setting up the websocket client transport. These two logic exist in our company's codebase and we have used it for a year already, but somehow was missing in the PR to this upstream codebase. Without these two conditions, the websocket client will create a `esp_transport` handle anyhow, no matter the `ext_transport` was provided or not. This will break the proxy support of course.

Anyway, here's the fix. Sorry again for the trouble I made.

Regards,
Jackson

## Related

PR related:

- https://github.com/espressif/esp-protocols/pull/573
- https://github.com/espressif/esp-protocols/pull/616

## Testing

These two logic exist in our company's codebase and we have used it for a year already in our products.

## Checklist

### DIFF
--- a/components/esp_modem/src/esp_modem_command_library.cpp
+++ b/components/esp_modem/src/esp_modem_command_library.cpp
@@ -635,17 +635,40 @@ command_result get_network_registration_state(CommandableIf *t, int &state)
     }
 
     constexpr std::string_view pattern = "+CEREG: ";
-    int state_pos_start;
-    int state_pos_end;
-    if (out.find(pattern) == std::string::npos || (state_pos_start = out.find(',')) == std::string::npos) {
+
+    // Check if pattern exists and find its position
+    size_t pattern_pos = out.find(pattern);
+    if (pattern_pos == std::string::npos) {
         return command_result::FAIL;
     }
 
-    if (out.find(pattern) == std::string::npos || (state_pos_end = out.find(',', state_pos_start)) == std::string::npos) {
-        if (std::from_chars(out.data() + state_pos_start, out.data() + out.size(), state).ec == std::errc::invalid_argument) {
-            return command_result::FAIL;
+    // Find the first comma after the pattern
+    size_t state_pos_start = out.find(',', pattern_pos);
+    if (state_pos_start == std::string::npos) {
+        return command_result::FAIL;
+    }
+
+    // Find the end of the state value - either a second comma or end of line
+    size_t state_pos_end = out.find(',', state_pos_start + 1);
+    if (state_pos_end == std::string::npos) {
+        // No second comma found, look for end of line characters
+        state_pos_end = out.find('\r', state_pos_start);
+        if (state_pos_end == std::string::npos) {
+            state_pos_end = out.find('\n', state_pos_start);
         }
-    } else if (std::from_chars(out.data() + state_pos_start, out.data() + state_pos_end, state).ec == std::errc::invalid_argument) {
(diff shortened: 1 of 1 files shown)


huming2207:
Sorry, we internally reviewed and we think we have more issues to fix. I will fix it first and then let you know we are ready.

huming2207:
Okay now I'm ready for review. 

This also fixes some potential memory corruptions. Previously if I use an externally created proxy handle that is not a `transport_ssl` or `transport_tcp` handle, and if some options like keep alive was set, the `esp_transport_ssl_set_*()` will be called and eventually corrupts the proxy handle context. Because the type is different.

huming2207:
Hello @glmfe , may I ask is there any feedback on this PR?

huming2207:
Any updates...? @glmfe

euripedesrocha:
Hi @huming2207, thanks for the contribution, and sorry for the delay. 

I need to double-check the code because we shouldn't be reaching that logic. 

Second and most important, when adding the external transport we should be adding the full transport, not only the parent. To be honest, I'm undecided on what is more confusing to users. 

At this point, I believe that the simple approach is that the external transport must be the full transport implementation, not only the parent for `transport_ws`.

huming2207:
Hi @euripedesrocha 

Thanks for the followup. 

> I need to double-check the code because we shouldn't be reaching that logic.

Nah I think the original implementation has already caused us heap corruptions. 

> At this point, I believe that the simple approach is that the external transport must be the full transport implementation, not only the parent for transport_ws.

Yea, and the original implementation will simply doesn't work and cause memory corruptions. I think we need something that actually work at least... 😅

Jackson

euripedesrocha:
@huming2207 I would like to understand your problem here. 

The original version of this feature seems to take the correct approach. 

When adding an external transport, I expect that the user needs to set all parameters of the transport and the library should work with what was added. For this library it means that you need to initialize the parent and pass it to websocket_ts or even implement the base transport completely. I agree that the first case of adding the parent is the most common. 

Can you share a small reproducer that causes the memory corruption?

huming2207:
Sorry @euripedesrocha I was a bit busy earlier as well 😅 

> When adding an external transport, I expect that the user needs to set all parameters of the transport and the library should work with what was added. For this library it means that you need to initialize the parent and pass it to websocket_ts or even implement the base transport completely. I agree that the first case of adding the parent is the most common.

> Can you share a small reproducer that causes the memory corruption? 

Ah yes I think you are right. I misread the code and the `esp_transport_tcp_set_***()` won't apply to the external transport because I already have a guard here anyway: https://github.com/espressif/esp-protocols/blob/f43dd5012fb1e5d2decb79bd9941b1771e3fea5f/components/esp_websocket_client/esp_websocket_client.c#L1159

But yes for https://github.com/espressif/esp-protocols/pull/811/commits/fb227494e0d89287c7d7448feae7575c4565aa27 , This one I think is necessary. Otherwise it won't actually use my external transport at all.

Sorry for the trouble. I will remove this commit: https://github.com/espressif/esp-protocols/pull/811/commits/76b475e56debe4823ccba931731df15401ebd801

huming2207:
Sorry I will come back on this a bit later. I think I might did something wrong in the beginning. I will come back to this maybe next week after I fix some other urgent issues.

PR #812, opened 2025-05-07
feat(eppp): Add support for TUN interface

david-cermak:
- [x] Fix all + cleanup transports
- [x] Implement factory pattern
- [x] Make netif type configurable (PPP vs. TUN)
- [x] Use ICMP for TUN netif connection
- [x] Relax deps criteria (mainly for Ethernet)

---
#### Followup PR:
* Support for channels

### DIFF
--- a/components/eppp_link/README.md
+++ b/components/eppp_link/README.md
@@ -1,23 +1,44 @@
 # ESP PPP Link component (eppp_link)
 
-The component provides a general purpose connectivity engine between two microcontrollers, one acting as PPP server (slave), the other one as PPP client (host).
-This component could be used for extending network using physical serial connection. Applications could vary from providing PRC engine for multiprocessor solutions to serial connection to POSIX machine. This uses a standard PPP protocol to negotiate IP addresses and networking, so standard PPP toolset could be used, e.g. a `pppd` service on linux. Typical application is a WiFi connectivity provider for chips that do not have WiFi
+The component provides a general purpose connectivity engine between two microcontrollers, one acting as PPP server, the other one as PPP client.
+This component could be used for extending network using physical serial connection. Applications could vary from providing PRC engine for multiprocessor solutions to serial connection to POSIX machine. This uses a standard PPP protocol (if enabled) to negotiate IP addresses and networking, so standard PPP toolset could be used, e.g. a `pppd` service on linux. Typical application is a WiFi connectivity provider for chips that do not have WiFi.
+Uses simplified TUN network interface by default to enable faster data transfer on non-UART transports.
 
 ## Typical application
 
 Using this component we can construct a WiFi connectivity gateway on PPP channel. The below diagram depicts an application where
 PPP server is running on a WiFi capable chip with NAPT module translating packets between WiFi and PPPoS interface.
-We usually call this node a SLAVE microcontroller. The "HOST" microcontroller runs PPP client and connects only to the serial line,
-brings in the WiFi connectivity from the "SLAVE" microcontroller.
+We usually call this node a communication coprocessor, or a "SLAVE" microcontroller.
+The main microcontroller (sometimes also called the "HOST") runs PPP client and connects only to the serial line,
+brings in the WiFi connectivity from the communication coprocessor.
 
 ```
-             SLAVE micro                                     HOST micro
+        Communication coprocessor                        Main microcontroller
     \|/  +----------------+                               +----------------+
      |   |                |       (serial) line           |                |
      +---+ WiFi NAT PPPoS |=== UART / SPI / SDIO / ETH ===| PPPoS client   |
          |        (server)|                               |                |
          +----------------+                               +----------------+
 ```
+
+## Configuration
+
+### Choose the transport layer
+
+Use `idf.py menuconfig` to select the transport layer:
+
+* `CONFIG_EPPP_LINK_UART` -- Use UART transport layer
+* `CONFIG_EPPP_LINK_SPI` -- Use SPI transport layer
+* `CONFIG_EPPP_LINK_SDIO` -- Use SDIO transport layer
+* `CONFIG_EPPP_LINK_ETHERNET` -- Use Ethernet transport
+  - Note: Ethernet creates it's own task, so calling `eppp_perform()` would not work
+  - Note: Add dependency to ethernet_init component to use other Ethernet drivers
+  - Note: You can override functions `eppp_transport_ethernet_deinit()` and `eppp_transport_ethernet_init()` to use your own Ethernet driver
+
+### Choose the network interface
+
+Use PPP netif for UART; Keep the default (TUN) for others
+
 
 ## API
 
--- a/components/eppp_link/eppp_eth.c
+++ b/components/eppp_link/eppp_eth.c
@@ -11,11 +11,18 @@
 #include "esp_check.h"
 #include "esp_event.h"
 #include "esp_mac.h"
+#include "esp_idf_version.h"
 #include "eppp_link.h"
 #include "eppp_transport.h"
 #include "esp_eth_driver.h"
-#include "ethernet_init.h"
 #include "esp_eth_spec.h"
+#include "eppp_transport_eth.h"
+// Use Ethernet Init component if available
+// (otherwise use just simple init/deinit with generic MAC/PHY)
+#if __has_include("ethernet_init.h")
+#define USE_ETHERNET_INIT_COMPONENT
+#include "ethernet_init.h"
+#endif
 
 typedef struct header {
     uint8_t dst[ETH_ADDR_LEN];
@@ -26,9 +33,56 @@ typedef struct header {
 static const char *TAG = "eppp_ethernet";
 static bool s_is_connected = false;
 static esp_eth_handle_t *s_eth_handles = NULL;
-static uint8_t s_out_buffer[ETH_MAX_PACKET_SIZE];
 static uint8_t s_their_mac[ETH_ADDR_LEN];
 static uint8_t s_our_mac[ETH_ADDR_LEN];
+
+#ifndef USE_ETHERNET_INIT_COMPONENT
+static esp_eth_handle_t s_handles[1] = { NULL };
+static esp_eth_mac_t *s_mac = NULL;
+static esp_eth_phy_t *s_phy = NULL;
+
+static void simple_deinit(esp_eth_handle_t *handle_array[])
+{
+    if (s_handles[0] != NULL) {
+        esp_eth_driver_uninstall(s_handles[0]);
+        s_handles[0] = NULL;
+    }
+    if (s_mac != NULL) {
+        s_mac->del(s_mac);
+        s_mac = NULL;
+    }
+    if (s_phy != NULL) {
+        s_phy->del(s_phy);
+        s_phy = NULL;
+    }
+}
+
+static esp_err_t simple_init(struct eppp_config_ethernet_s *config, esp_eth_handle_t *handle_array[])
+{
+    esp_err_t ret = ESP_OK;
+    eth_mac_config_t mac_config = ETH_MAC_DEFAULT_CONFIG();
+    eth_esp32_emac_config_t esp32_emac_config = ETH_ESP32_EMAC_DEFAULT_CONFIG();
+    esp32_emac_config.smi_gpio.mdc_num = config->mdc_io;
+    esp32_emac_config.smi_gpio.mdio_num = config->mdio_io;
+    s_mac = esp_eth_mac_new_esp32(&esp32_emac_config, &mac_config);
+    eth_phy_config_t phy_config = ETH_PHY_DEFAULT_CONFIG();
+    phy_config.phy_addr = config->phy_addr;
+    phy_config.reset_gpio_num = config->rst_io;
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+    s_phy = esp_eth_phy_new_generic(&phy_config);
+#else
+    s_phy = esp_eth_phy_new_ip101(&phy_config);
+#endif
+    esp_eth_config_t eth_config = ETH_DEFAULT_CONFIG(s_mac, s_phy);
+    ESP_GOTO_ON_ERROR(esp_eth_driver_install(&eth_config, &s_handles[0]), err, TAG, "Ethernet driver install failed");
+    *handle_array = s_handles;
+    return ESP_OK;
+err:
+    simple_deinit(handle_array);
+    return ret;
+}
+
+#endif // USE_ETHERNET_INIT_COMPONENT
 
 static void event_handler(void *arg, esp_event_base_t event_base,
                           int32_t event_id, void *event_data)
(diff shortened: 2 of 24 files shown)


components/eppp_link/README.md
```diff
@@ -19,6 +20,23 @@ brings in the WiFi connectivity from the "SLAVE" microcontroller.
          +----------------+                               +----------------+
 ```
 
+## Configuration
+
+### Choose the transport layer
+
```
euripedesrocha, line 27:
```suggestion

Use `idf.py menuconfig` to select the transport layer.

```

components/eppp_link/include/eppp_transport_eth.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * SPDX-FileCopyrightText: 2023-2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include "eppp_link.h"
+
+eppp_transport_handle_t eppp_eth_init(struct eppp_config_ethernet_s *config);
```
euripedesrocha, line 10: Could we unify the init/deinit pushing it to "TAG" the config struct as enum dispatching internally to the channels?

components/eppp_link/examples/slave/main/eppp_slave.c
```diff
@@ -127,18 +124,26 @@ void app_main(void)
...
+    config.spi.intr = 17;
+    config.spi.freq = 1000000;
 #elif CONFIG_EPPP_LINK_DEVICE_UART
     config.transport = EPPP_TRANSPORT_UART;
-    config.uart.tx_io = CONFIG_EXAMPLE_UART_TX_PIN;
-    config.uart.rx_io = CONFIG_EXAMPLE_UART_RX_PIN;
-    config.uart.baud = CONFIG_EXAMPLE_UART_BAUDRATE;
+    config.uart.tx_io = 18;
+    config.uart.rx_io = 17;
+//    config.uart.baud = CONFIG_EXAMPLE_UART_BAUDRATE;
```
euripedesrocha: Clean or uncomment?

components/eppp_link/eppp_eth.c
```diff
@@ -26,10 +33,55 @@ typedef struct header {
...
+            handles[0] = NULL;
+        }
+        if (mac != NULL) {
+            mac->del(mac);
+            mac = NULL;
+        }
+        if (phy != NULL) {
+            phy->del(phy);
+            phy = NULL;
+        }
```
kostaond: Is this some placeholder? Since now it's a dead code because `handles`, `mac` and `phy` is always initialized to `NULL`.

components/eppp_link/eppp_eth.c
```diff
@@ -65,23 +117,48 @@ static esp_err_t receive(esp_eth_handle_t h, uint8_t *buffer, uint32_t len, void
...
+    header_t *head = (header_t *)out_buffer;
+    memcpy(head->dst, s_their_mac, ETH_ADDR_LEN);
+    memcpy(head->src, s_our_mac, ETH_ADDR_LEN);
+    head->len = len;
+    // support only payloads with len <=  ETH_MAX_PAYLOAD_LEN
+    if (len > ETH_MAX_PAYLOAD_LEN) {
+        return ESP_FAIL;
+    }
+    memcpy(out_buffer + ETH_HEADER_LEN, buffer, len);
+    return esp_eth_transmit(s_eth_handles[0], out_buffer, len + ETH_HEADER_LEN);
```
kostaond: Transmit could be optimized by using `esp_eth_transmit_ctrl_vargs` function. If you use it, you can skip the whole `memcpy(out_buffer + ETH_HEADER_LEN, buffer, len);` operation. The Ethernet frame is then assembled only during copying to DMA.

Note that you can use `esp_eth_transmit_vargs` wrapper to ensure backward compatibility with older versions of IDF.

components/eppp_link/eppp_sdio.c
```diff
@@ -0,0 +1,86 @@
...
+esp_err_t eppp_sdio_host_init(struct eppp_config_sdio_s *config);
+esp_err_t eppp_sdio_slave_init(struct eppp_config_sdio_s *config);
+void eppp_sdio_slave_deinit(void);
+void eppp_sdio_host_deinit(void);
+
+esp_err_t eppp_perform(esp_netif_t *netif)
+{
+    struct eppp_handle *h = esp_netif_get_io_driver(netif);
+    if (h->stop) {
+        return ESP_ERR_TIMEOUT;
```
euripedesrocha, line 37: Why return `ESP_ERR_TIMEOUT`?

components/eppp_link/eppp_sdio.c
```diff
@@ -0,0 +1,86 @@
...
+    if (handle->is_host) {
+        return eppp_sdio_host_rx(netif);
+    } else {
+        return eppp_sdio_slave_rx(netif);
+    }
+}
+
+static esp_err_t post_attach(esp_netif_t *esp_netif, void *args)
+{
+    eppp_transport_handle_t h = (eppp_transport_handle_t)args;
```
euripedesrocha, line 49: Do we need to validate the pointer here? Or it will be validated by the caller?

components/eppp_link/eppp_sdio.c
```diff
@@ -0,0 +1,86 @@
...
+    return ESP_OK;
+}
+
+eppp_transport_handle_t eppp_sdio_init(struct eppp_config_sdio_s *config)
+{
+    __attribute__((unused)) esp_err_t ret = ESP_OK;
+    struct eppp_sdio *h = calloc(1, sizeof(struct eppp_sdio));
+    ESP_RETURN_ON_FALSE(h, NULL, TAG, "Failed to allocate eppp_handle");
+    h->parent.base.post_attach = post_attach;
+    h->is_host = config->is_host;
```
euripedesrocha, line 71: Do we need to check for `config != NULL`?

components/eppp_link/eppp_spi.c
```diff
@@ -0,0 +1,468 @@
...
+    bus_cfg.mosi_io_num = config->mosi;
+    bus_cfg.miso_io_num = config->miso;
+    bus_cfg.sclk_io_num = config->sclk;
+    bus_cfg.quadwp_io_num = -1;
+    bus_cfg.quadhd_io_num = -1;
+    bus_cfg.max_transfer_sz = TRANSFER_SIZE;
+    bus_cfg.flags = 0;
+    bus_cfg.intr_flags = 0;
+
+    // TODO: Init and deinit SPI bus separately (per Kconfig?)
```
euripedesrocha: Resolve TODO or create a follow-up task.

components/eppp_link/eppp_spi.c
```diff
@@ -0,0 +1,468 @@
...
+        if (h->outbound.len == 0) { // we're blocked if we don't have any data
+            h->blocked = SLAVE_BLOCKED;
+        } else {
+            h->blocked = SLAVE_WANTS_WRITE; // we notify the master that we want to write
+            esp_timer_start_once(h->timer, MIN_TRIGGER_US);
+        }
+    }
+}
+
+static void post_trans(spi_slave_transaction_t *trans)
```
euripedesrocha: Nitpick: Other functions use transaction, could we be consistent and make this `post_transaction` and also rename the argument?

components/eppp_link/eppp_uart.c
```diff
@@ -0,0 +1,213 @@
...
+        current_offset = 0;
+        return;
+    }
+
+    // Process while we have enough data for at least a header
+    while (current_offset >= sizeof(struct header)) {
+        // Verify magic number
+        if (head->magic != HEADER_MAGIC) {
+            // Try to recover by shifting one byte at a time
+            memmove(in_buf, in_buf + 1, current_offset - 1);
```
kostaond: It seems like inefficient way of how to find a new header's start, in my opinion. Too much copying involved. From `memmove` description:  "The objects may overlap: copying takes place as if the characters were copied to a temporary character array and then the characters were copied from the array to dest."

components/eppp_link/eppp_uart.c
```diff
@@ -0,0 +1,213 @@
...
+#define UART_BUF_SIZE (1024)
+#define MAX_PAYLOAD (1500)
+#define HEADER_MAGIC (0x7E)
+#define PACKET_SIZE (MAX_PAYLOAD + 4)
+
+struct header {
+    uint8_t magic;
+    uint8_t check;
+    uint16_t size;
+} __attribute__((packed));
```
kostaond, line 36: Is this a final solution? 8-bit magic number and 8-bit check based on length which is 1500B maximum does not provide enough combinations to be reliable enough, in my opinion.

components/eppp_link/eppp_eth.c
```diff
@@ -26,10 +33,55 @@ typedef struct header {
 static const char *TAG = "eppp_ethernet";
 static bool s_is_connected = false;
 static esp_eth_handle_t *s_eth_handles = NULL;
-static uint8_t s_out_buffer[ETH_MAX_PACKET_SIZE];
 static uint8_t s_their_mac[ETH_ADDR_LEN];
 static uint8_t s_our_mac[ETH_ADDR_LEN];
 
+#ifndef USE_ETHERNET_INIT_COMPONENT
+static esp_err_t simple_init_deinit(bool init, struct eppp_config_ethernet_s *config, esp_eth_handle_t *handle_array[])
```
alisitsyn: Is there any significant advantage of using this init-deinit instead of separate constructor-destructor?

components/eppp_link/include/eppp_transport_eth.h
david-cermak: we need to keep it separate for each transport

components/eppp_link/eppp_sdio.c
```diff
@@ -0,0 +1,86 @@
...
+void eppp_sdio_slave_deinit(void);
+void eppp_sdio_host_deinit(void);
+
+esp_err_t eppp_perform(esp_netif_t *netif)
+{
+    struct eppp_handle *h = esp_netif_get_io_driver(netif);
+    if (h->stop) {
+        return ESP_ERR_TIMEOUT;
+    }
+    struct eppp_sdio *handle = __containerof(h, struct eppp_sdio, parent);;
```
alisitsyn, line 39: Minor: remove extra semicolon

components/eppp_link/eppp_sdio.c
david-cermak: Need to document how the `eppp_perform()` is supposed to work and by returning `ESP_ERR_TIMEOUT` we ask to stop.

david-cermak: or pickup a better name

components/eppp_link/eppp_sdio.c
david-cermak: I'll check if add a null check if needed

components/eppp_link/eppp_sdio.c
david-cermak: will return an error

components/eppp_link/eppp_spi.c
david-cermak: will add a jira task

components/eppp_link/eppp_spi.c
david-cermak: will rename the function name and the argument

components/eppp_link/eppp_eth.c
david-cermak: I'll split this into two functions to make it more natural for readers.

david-cermak: I'll consider just swapping the init/deinit parts

components/eppp_link/eppp_eth.c
david-cermak: Thanks, will use it

components/eppp_link/eppp_uart.c
david-cermak: I'll fix it

components/eppp_link/eppp_uart.c
david-cermak: I'll calculate the probability and let you know

david-cermak: will use CRC instead

components/eppp_link/eppp_sdio.c
david-cermak: will fix

components/eppp_link/README.md
```diff
@@ -1,7 +1,8 @@
 # ESP PPP Link component (eppp_link)
 
 The component provides a general purpose connectivity engine between two microcontrollers, one acting as PPP server (slave), the other one as PPP client (host).
```
david-cermak:
```suggestion
The component provides a general purpose connectivity engine between two microcontrollers, one acting as PPP server, the other one as PPP client.
```

components/eppp_link/README.md
```diff
@@ -19,6 +20,23 @@ brings in the WiFi connectivity from the "SLAVE" microcontroller.
          +----------------+                               +----------------+
```
david-cermak, line 21: Change:
SLAVE micro -> communication coprocessor

HOST micro -> main micro

components/eppp_link/README.md
david-cermak: fixed

components/eppp_link/include/eppp_transport_eth.h
david-cermak: left to decouple link-time transport dependency

components/eppp_link/examples/slave/main/eppp_slave.c
david-cermak: cleaned-up throughout both projects and all transports -- using Kconfig options

components/eppp_link/eppp_sdio.c
david-cermak: Documented and used `ESP_ERR_TIMEOUT` (as I couldn't find a better name in the generic types)

Alternatively I can make an `enum {OK, FAIL, DONE}` -- but that would be a breaking change.

components/eppp_link/eppp_sdio.c
david-cermak: added checks everywhere, as users could place NULL ptr in `esp_netif_attach(netif, NULL <---- here);`

components/eppp_link/eppp_sdio.c
david-cermak: added an errorcheck

components/eppp_link/eppp_spi.c
david-cermak: created jira

components/eppp_link/eppp_spi.c
david-cermak: renamed

components/eppp_link/eppp_eth.c
david-cermak: reworked to the old style ini-deinit with global statics

components/eppp_link/eppp_eth.c
david-cermak: Checked the implementation of `esp_eth_transmit_ctrl_vargs()`, but it seems like it expect the whole Ethernet frames including headers, so you cannot transmit header and payload in two buffers, correct?
```
esp_eth_transmit_ctrl_vargs(eth_handle, &ts, 2, test_pkt, transmit_size));
// here `test_pkt[] = { eth-header, payload }`
```

I would still need to reassembly the frame here, since I got only TUN payload (starting from IP header) from the network (unlike standard network handlers, where we get Ethernet frames).
Could make some improvements, like "prepend" the headers to a pbuf, but that would again bring lwip deps here.

components/eppp_link/eppp_uart.c
david-cermak: reworked

components/eppp_link/eppp_sdio.c
david-cermak: fixed

components/eppp_link/README.md
david-cermak: rephrased

components/eppp_link/README.md
david-cermak: rephrased and used "communication + main micro" instead (with slave+host in parentheses, as they are used some times, too)

This change also introduces the change in SPI transport, so you can use SPI master and slave interchangeably, e.g. running SPI master on "slave micro" and SPI slave on "host micro" -- same with SDIO

components/eppp_link/eppp_uart.c
david-cermak: this should be safe, as we're checking the integrity only for the header (not the payload, which is evaluated on upper layers)

But this made me think that adding magic to the calculation doesn't bring any entropy and using CRC on three bytes is counter-productive. So my suggestion is that the check would only `XOR` upper and lower byte of the length.
In the upcoming update, it would also include "channel", improving the probability a bit.

components/eppp_link/README.md
```diff
@@ -1,24 +1,45 @@
...
 PPP server is running on a WiFi capable chip with NAPT module translating packets between WiFi and PPPoS interface.
-We usually call this node a SLAVE microcontroller. The "HOST" microcontroller runs PPP client and connects only to the serial line,
-brings in the WiFi connectivity from the "SLAVE" microcontroller.
+We usually call this node a communication coprocessor, or a "SLAVE" microcontroller.
+The main microcontroller (sometimes also called the "HOST") runs PPP client and connects only to the serial line,
+brings in the WiFi connectivity from the communication coprocessor.
 
 ```
-             SLAVE micro                                     HOST micro
+        Communication comprocessor                        Main microcontroller
```
euripedesrocha:
```suggestion
        Communication coprocessor                        Main microcontroller
```

euripedesrocha (approved):
Only a small typo to fix.

components/eppp_link/README.md
david-cermak: Thanks, updated!

components/eppp_link/eppp_eth.c
david-cermak: oh, it actually works. Thanks! fixed as suggested!

components/eppp_link/eppp_uart.c
david-cermak: @kostaond Per my calculations, it's about 1 in 11,400 packets of random data that could be wrongly identified as a valid packet header.

**Long story**
* P of finding magic in random data: `P(magic)=1/256`
* P of checksum: `P(check)=1501/65536`
* combined: `8.8×10^-5`, which corresponds to the above number

So I think we're more than safe, here.

PS: I'd still apply the XOR idea above, but will do it in the upcoming fix

kostaond: Yeah, XOR idea is definitely better than just simple sum as was used before because XOR should ensure more even distribution of possible combinations.

kostaond: It's still sum if I see correctly `uint8_t calculated_check = HEADER_MAGIC + head->size + (head->size >> 8);`

david-cermak: I meant to address it in my followup PR: https://github.com/espressif/esp-protocols/pull/814

but applied and tested.
Thanks for bringing it up, BTW, the P-calc above it a bit "biased", since the first part (magic) should be neglected, because some misalignment eventually happens in any stream operations, and to recover correctly we have only 97% chance :thinking:  (Arithmetic sums have one advantage -- easy to calculate probability, I'm not brave enough to update the formula now, with the `XOR` employed).

Thank you, resolving!

components/eppp_link/eppp_uart.c
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        // Got a complete packet, pass it to network
+        esp_netif_receive(netif, in_buf + buf_start + sizeof(struct header), payload_size, NULL);
+
+        // Advance start pointer past this packet
+        buf_start += total_packet_size;
+
+        // compact if we don't have enough space for 1x UART_BUF_SIZE
+        if (buf_start > (sizeof(in_buf) / 2) || (sizeof(in_buf) - buf_end) < UART_BUF_SIZE) {
```
kostaond, line 145: Just double check if the the watermark isn't set too low. If I understand it correctly it's half of the `in_buf`. Would it be possible to set it to e.g. 3/4?

david-cermak: Looks okay from my side, setting it too high, like 3/4, has one disadvantage, the next max read size must be adjusted to 1/4 of the buffer size = inefficiency.

kostaond: My concern is that we receive a frame with maximum size. Such frame is expected to be received in multiple chunks based on my understanding of implementation of `process_packet` function. Hence it's appears to me that it is possible to receive the last chunk of such frame along with start of the new frame...  Oh, I see, it should be OK, since  `UART_BUF_SIZE (1536)` > `ETH_MAX_PACKET_SIZE`, i.e. the header should still fit into the first half of the buffer.

So what about to `#define UART_BUF_SIZE(ETH_MAX_PACKET_SIZE + AT_LEAST_HEADER_SIZE)` to make it super clear.

david-cermak: Reworked the receive function a bit, again, saving 1x`UART_BUF_SIZE`.

> So what about to #define UART_BUF_SIZE(ETH_MAX_PACKET_SIZE + AT_LEAST_HEADER_SIZE) to make it super clear.

Updated, just note that it's not `ETH_MAX_PACKET_SIZE`, but the maximum payload (which in this case is TUN packet size, not Ethernet)

PR #814, opened 2025-05-26
[eppp-link]: Support for channels

david-cermak:
## Followup on https://github.com/espressif/esp-protocols/pull/812

* Adds support for custom channels
* Could be used to transmit/receive any data
* Demonstrated with channeling WiFi 802.11 frames from host to slave (+ one chat channel)

## Task list

- [x] Cleanup commit history
- [x] Documentation


components/eppp_link/plan.md
```diff
@@ -0,0 +1,28 @@
+# Plan: Make Number of Channels Configurable in eppp_link
```
euripedesrocha: Should we keep this?

david-cermak: I'll probably delete it, thanks for noticing!

PR #821, opened 2025-06-11
fix(modem): Consume buffer after handled URC

lolrobbe2:
## Description

Currently the urc_line handler function does not consume the buffer, this PR aims to fix this.

## Related

The PR uses the implementation discussed here: https://github.com/espressif/esp-protocols/pull/810

## Testing

All changes where tested useing the DPTechnics walter board using sequans GM02S IoT modem

---

## Checklist

### DIFF
--- a/components/esp_modem/src/esp_modem_dte.cpp
+++ b/components/esp_modem/src/esp_modem_dte.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -369,18 +369,23 @@ void DTE::on_read(got_line_cb on_read_cb)
 bool DTE::command_cb::process_line(uint8_t *data, size_t consumed, size_t len)
 {
 #ifdef CONFIG_ESP_MODEM_URC_HANDLER
+    command_result commandResult = command_result::FAIL;
     if (urc_handler) {
-        urc_handler(data, consumed + len);
+        commandResult = urc_handler(data, consumed + len);
     }
-    if (result != command_result::TIMEOUT || got_line == nullptr) {
-        return false;   // this line has been processed already (got OK or FAIL previously)
+    if (result != command_result::TIMEOUT && got_line == nullptr) {
+        return false; // this line has been processed already (got OK or FAIL previously)
     }
 #endif
     if (memchr(data + consumed, separator, len)) {
-        result = got_line(data, consumed + len);
+        result = got_line(data + consumed, consumed + len);
         if (result == command_result::OK || result == command_result::FAIL) {
             signal.set(GOT_LINE);
+#ifdef CONFIG_ESP_MODEM_URC_HANDLER
+            return commandResult == command_result::OK;
+#else
             return true;
+#endif
         }
     }
     return false;


david-cermak (commented):
Thanks for following up on this @lolrobbe2 !

This LGTM! I'll quickly test it and merge. Thank you!

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -369,18 +369,23 @@ void DTE::on_read(got_line_cb on_read_cb)
...
 {
 #ifdef CONFIG_ESP_MODEM_URC_HANDLER
+    command_result commandResult = command_result::FAIL;
     if (urc_handler) {
-        urc_handler(data, consumed + len);
+        commandResult = urc_handler(data, consumed + len);
     }
-    if (result != command_result::TIMEOUT || got_line == nullptr) {
-        return false;   // this line has been processed already (got OK or FAIL previously)
+    if (result != command_result::TIMEOUT && got_line == nullptr) {
```
david-cermak, line 376: The only remaining question is on this line: `||` -> `&&`, while the logic hasn't changed -- and should be the same as for the `CONFIG_ESP_MODEM_URC_HANDLER=n`, right?

david-cermak (approved):
LGTM, otherwise

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: Okay, let's merge as is and I'll me handle the case above differently (in a future PR)

dgrochowalski, line 376: I don't fully understand this change, isn't risky that we could enter rest part of the code which tries to call got_line, which will be null:
```
result = got_line(data + consumed, consumed + len);
```

shouldn't we as wee modify:
https://github.com/espressif/esp-protocols/blob/b14389487409546f29d9077dd3b0ae1a688b5f89/components/esp_modem/src/esp_modem_dte.cpp#L69

?

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -369,18 +369,23 @@ void DTE::on_read(got_line_cb on_read_cb)
...
+        return false; // this line has been processed already (got OK or FAIL previously)
     }
 #endif
     if (memchr(data + consumed, separator, len)) {
-        result = got_line(data, consumed + len);
+        result = got_line(data + consumed, consumed + len);
```
dgrochowalski, line 381: Shouldn't be
```
result = got_line(data + consumed, len);
```
?

This change is quite significant and I had few custom AT command which were "based on this". Change itself I think is really good, but needs to be checked carefully, especially for generic_get_string().

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -369,18 +369,23 @@ void DTE::on_read(got_line_cb on_read_cb)
...
 #ifdef CONFIG_ESP_MODEM_URC_HANDLER
+    command_result commandResult = command_result::FAIL;
     if (urc_handler) {
-        urc_handler(data, consumed + len);
+        commandResult = urc_handler(data, consumed + len);
     }
-    if (result != command_result::TIMEOUT || got_line == nullptr) {
-        return false;   // this line has been processed already (got OK or FAIL previously)
+    if (result != command_result::TIMEOUT && got_line == nullptr) {
+        return false; // this line has been processed already (got OK or FAIL previously)
```
dgrochowalski, line 377: shouldn't be here:
```
return commandResult == command_result::OK;
```

components/esp_modem/src/esp_modem_dte.cpp
```diff
@@ -369,18 +369,23 @@ void DTE::on_read(got_line_cb on_read_cb)
...
     }
 #endif
     if (memchr(data + consumed, separator, len)) {
-        result = got_line(data, consumed + len);
+        result = got_line(data + consumed, consumed + len);
         if (result == command_result::OK || result == command_result::FAIL) {
             signal.set(GOT_LINE);
+#ifdef CONFIG_ESP_MODEM_URC_HANDLER
```
dgrochowalski, line 384: Shouldn't be:
```
        if (result == command_result::OK || result == command_result::FAIL)
        {
            signal.set(GOT_LINE);
            return true;
        }
    }
#ifdef CONFIG_ESP_MODEM_URC_HANDLER
    return commandResult == command_result::OK;
#else
    return false;
#endif
```
?

dgrochowalski:
Sorry for late comments, but I was testing changes. Actually I am still testing them.

david-cermak:
> @david-cermak @lolrobbe2 Sorry for late comments, but I was testing changes. Actually I am still testing them.

@dgrochowalski Thanks for taking a look and testing it! Maybe I was too hasty to merge it, sorry for that.
I was going to handle some edge cases separately, so I'll take your comments and address them in a future PR -- thank you again for your inputs!

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: correct, this change might cause nullptr deref!

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: Yes, exactly!

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: need to revert this change

components/esp_modem/src/esp_modem_dte.cpp
david-cermak: depends, the idea was to keep the potential half-read buffer for another URC (if needed).
But updating this would be a breaking change, so I'd like to avoid it.

PR #834, opened 2025-07-01
[modem]: Fix to use compatible iterator types for std::search in new gcc

david-cermak:
Fixes: https://github.com/espressif/esp-protocols/actions/runs/15997026484/job/45122679598#step:5:64

~~GCC 15 and later require the 5-argument version of std::search when using raw pointers, as the 4-argument overload is only available for ForwardIterators (not raw pointers). This explicit predicate ensures compatibility with modern libstdc++~~

Fixed by replacing `search()` with `find()`


components/esp_modem/examples/modem_tcp_client/main/sock_commands_bg96.cpp
```diff
@@ -130,7 +130,12 @@ Responder::ret Responder::recv(uint8_t *data, size_t len)
     auto *recv_data = (char *)data;
     if (data_to_recv == 0) {
         const std::string_view head = "+QIRD: ";
-        auto head_pos = std::search(recv_data, recv_data + len, head.begin(), head.end());
+        // GCC 15 and later require the 5-argument version of std::search when using raw pointers,
+        // as the 4-argument overload is only available for ForwardIterators (not raw pointers).
+        auto head_pos = std::search(recv_data, recv_data + len, head.data(), head.data() + head.size(), [](char a, char b) {
```
euripedesrocha: As suggested in the other MR, move to use  `std::string_view` offers a simpler solution(but we can adopt it later).

euripedesrocha (approved):

components/esp_modem/examples/modem_tcp_client/main/sock_commands_bg96.cpp
david-cermak: Thanks, updated.

Fixed here, since I need to use another public broker to unblock the CI.

PR #837, opened 2025-07-07
[modem]: Extend tcp-client example

david-cermak:
- [x] Support for esp-at endpoint
- [ ] ~~Support for multiple connections~~ (in a followup PR)


components/esp_modem/examples/modem_tcp_client/main/espat_module/sock_module.hpp
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#include "esp_modem_config.h"
+#include "cxx_include/esp_modem_api.hpp"
+
+#pragma  once
+
+namespace sock_dce {
+
+class Module: public esp_modem::GenericModule {
```
euripedesrocha, line 14: This being tightly coupled to esp-at could be renamed to reflect that.

david-cermak: I use the same names for other implementations, so they can be easily replaced.

PR #841, opened 2025-07-10
fix Stuck when ‘disable_auto_reconnect = false’ and Call 'esp_websocket_client_close' when the connection fails and automatically reconnects (IDFGH-15854)

bigtreetech:
bug 复现步骤
1. 配置 `disable_auto_reconnect = false` 开启自动重连功能
2. 建立一个无法连接的client，此时后台任务会在后台重连，并且此时 `client->run` is `true`
3. 调用 `esp_websocket_client_close` 关闭此 clent 会堵塞在此函数中。


gabsuren:
@bigtreetech Thank you for the report. 

This has been addressed in https://github.com/espressif/esp-protocols/pull/1018

PR #849, opened 2025-07-16
[modem]: tcp-client example to support multiple connections

david-cermak:
* Follow upon https://github.com/espressif/esp-protocols/pull/837


components/esp_modem/examples/modem_tcp_client/main/generate/sock_dce.cpp
```diff
@@ -139,8 +183,8 @@ bool DCE::at_to_sock()
 bool DCE::sock_to_at()
 {
     ESP_LOGD(TAG,  "socket read: data available");
-    if (!signal.wait(IDLE, 1000)) {
-        ESP_LOGE(TAG,  "Failed to get idle");
+    if (!signal.wait(IDLE, 5000)) {
```
euripedesrocha: Should this be configurable?

components/esp_modem/examples/modem_tcp_client/main/generate/sock_dce.cpp
```diff
@@ -139,8 +183,8 @@ bool DCE::at_to_sock()
...
+        ESP_LOGE(TAG,  "Failed to get idle 2");
```
euripedesrocha: Is there another log message like this one? Could we make it easier for the location to be identified?

components/esp_modem/examples/modem_tcp_client/main/modem_client.cpp
```diff
@@ -23,14 +23,25 @@
 #include "tcp_transport_mbedtls.h"
 #include "tcp_transport_at.h"
 
-#define BROKER_URL "test.mosquitto.org"
+#define USE_TLS 0
```
euripedesrocha: This defines could be simplified by having the URL define directly without composing the macros. Also is the `USE_TLS` define a left over?

components/esp_modem/examples/modem_tcp_client/main/generate/sock_dce.cpp
david-cermak: Good point, added a constant

components/esp_modem/examples/modem_tcp_client/main/generate/sock_dce.cpp
david-cermak: This was a leftover, removed

components/esp_modem/examples/modem_tcp_client/main/modem_client.cpp
david-cermak: Added those configs to Kconfig, so users chould choose.

> Also is the USE_TLS define a left over?

It's not, as in this example the code is different for TCP and TLS connection (we use different tcp_transport() instances)

euripedesrocha (approved):

PR #862, opened 2025-08-12
fix(mbedtls_cxx): Enable mbedtls cookie support

david-cermak:
Fixing https://github.com/espressif/esp-protocols/actions/runs/16911809541/job/47915222746

Per v6.0 mbedtls updates


.github/workflows/tls_cxx__build.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         with:
           submodules: recursive
       - name: Build ${{ matrix.test.app }} with IDF-${{ matrix.idf_ver }}
+        env:
+          EXPECTED_WARNING: "unknown kconfig symbol"
```
kostaond: Would it be possible to make definition of `EXPECTED_WARNING` conditional? It is expected warning for IDF versions < 6

igrr:
```suggestion
          EXPECTED_WARNING: "unknown kconfig symbol 'MBEDTLS_SSL_COOKIE_C'"
```
or perhaps like this, since it won't allow _other_ unknown Kconfig options to creep in...

david-cermak: Thanks for your inputs! Added both

kostaond (approved):

Issue #851, opened 2025-07-17
[MODEM] disable certain features/ modules (IDFGH-15956)


lolrobbe2:
### Is your feature request related to a problem?

when using the modem just for sending AT commands it would be helpful to be able to disable the unused modem modules, or PPP for example as it just adds bloat to the final build.

### Describe the solution you'd like.

using KCONFIG add required `#ifdefs` for disabeling.

PR #863, opened 2025-08-14
[modem]: Support esp-modem use without PPP

david-cermak:
* Closes https://github.com/espressif/esp-protocols/issues/851

Supports building (and using) modem library without PPP mode.
It is already possible (see the [tcp-client](https://github.com/espressif/esp-protocols/tree/master/components/esp_modem/examples/modem_tcp_client) examples), but it only works with C++ interface.
This is more straight-forward approach.
It also assures PPP mode is "selected" for new projects that pull `esp_modem` component.

### DIFF
--- a/components/esp_modem/include/cxx_include/esp_modem_dce_factory.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_dce_factory.hpp
@@ -62,12 +62,16 @@ class Creator {
 public:
     Creator(std::shared_ptr<DTE> dte, esp_netif_t *esp_netif): dte(std::move(dte)), device(nullptr), netif(esp_netif)
     {
+#ifdef CONFIG_ESP_MODEM_USE_PPP_MODE
         ESP_MODEM_THROW_IF_FALSE(netif != nullptr, "Null netif");
+#endif
     }
 
     Creator(std::shared_ptr<DTE> dte, esp_netif_t *esp_netif, std::shared_ptr<T_Module> dev): dte(std::move(dte)), device(std::move(dev)), netif(esp_netif)
     {
+#ifdef CONFIG_ESP_MODEM_USE_PPP_MODE
         ESP_MODEM_THROW_IF_FALSE(netif != nullptr, "Null netif");
+#endif
     }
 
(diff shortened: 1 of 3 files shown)


components/esp_modem/include/cxx_include/esp_modem_dce_factory.hpp
```diff
@@ -62,12 +62,16 @@ class Creator {
 public:
     Creator(std::shared_ptr<DTE> dte, esp_netif_t *esp_netif): dte(std::move(dte)), device(nullptr), netif(esp_netif)
     {
+#ifdef CONFIG_ESP_MODEM_USE_PPP_MODE
         ESP_MODEM_THROW_IF_FALSE(netif != nullptr, "Null netif");
```
euripedesrocha, line 66: Other modes can be instantiated with a non initialized netif? Did you consider providing a default value for esp_netif parameter?

david-cermak: Yes, if user chooses not to use PPP, they could provide null netif.
Alternatively, we can create a separate constructor without netif or default netif param to null, but:
1) This would lead to more LOCs
2) Correct C++ solution is not needed as:

https://github.com/espressif/esp-protocols/blob/958ff6a584e3a106e594b8385868469b55005a70/components/esp_modem/examples/modem_tcp_client/main/command/sock_dce.cpp#L302-L305

(as mentioned above)

This is mainly needed for C-only API

euripedesrocha: You don't need to add a new constructor. You can have a default parameter in this one. But we can keep as is.

euripedesrocha (approved):

components/esp_modem/include/cxx_include/esp_modem_dce_factory.hpp
david-cermak: Oh okay, understand now -- that would basically turn the constructor into the code above:
```cpp
Factory::create(config, move(dte);
```
But still, defaulting the `netif` to null would make it unclear (for a very narrow scenario) as in most cases we want to **enforce** the parameter. This won't be very useful for C++ users either, as they could just **use** the factory per preferences.

Issue #864, opened 2025-08-17
eppp_link does not build unless ipv6 is enabled


bensuffolk:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### What component are you using? If you choose Other, provide details in More Information.

Other

### component version

1.0.0

### IDF version.

5.3.1

### More Information.

Unless you have IPV6 enabled in menu config (LWIP) There are several compile errors, but one such error is:

```
/Projects/Components/eppp_link/eppp_netif_tun.c:68:71: error: unknown type name 'ip6_addr_t'
   68 | static err_t tun_output_v6(struct netif *netif, struct pbuf *p, const ip6_addr_t *ipaddr)
```

PR #865, opened 2025-08-18
[eppp]: Support for IPv4-only mode

david-cermak:
Closes https://github.com/espressif/esp-protocols/issues/864

### DIFF
--- a/components/eppp_link/.cz.yaml
+++ b/components/eppp_link/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(eppp): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py eppp_link
   tag_format: eppp-v$version
-  version: 1.0.0
+  version: 1.0.1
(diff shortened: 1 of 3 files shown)


components/eppp_link/examples/slave/sdkconfig.defaults
```diff
@@ -4,3 +4,4 @@ CONFIG_LWIP_TCPIP_TASK_STACK_SIZE=4096
 CONFIG_LWIP_PPP_SUPPORT=y
 CONFIG_LWIP_PPP_SERVER_SUPPORT=y
 CONFIG_LWIP_PPP_VJ_HEADER_COMPRESSION=n
+CONFIG_LWIP_IPV6=n
```
euripedesrocha: Why make IPv4 only the default? Testing?

david-cermak: Yes, testing.
But I think I'd better make a separate `sdkconfig.ci` for that

PR #869, opened 2025-08-21
fix(CI): Fix test component dir after IDFv6.0 update

david-cermak:


components/eppp_link/test/test_app/CMakeLists.txt
```diff
@@ -1,7 +1,14 @@
 # The following four lines of boilerplate have to be in your project's CMakeLists
 # in this exact order for cmake to work correctly
 cmake_minimum_required(VERSION 3.16)
-set(EXTRA_COMPONENT_DIRS $ENV{IDF_PATH}/tools/unit-test-app/components)
+
+if (EXISTS $ENV{IDF_PATH}/tools/unit-test-app/components)
```
alisitsyn: @cermak, this is similar solution which would work good. In my case I check the IDF version instead:

```cmake
include($ENV{IDF_PATH}/tools/cmake/project.cmake)

set(EXTRA_COMPONENT_DIRS "../test_common")
if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER "5.5")
    list(APPEND EXTRA_COMPONENT_DIRS "$ENV{IDF_PATH}/tools/test_apps/components")
else()
    list(APPEND EXTRA_COMPONENT_DIRS "$ENV{IDF_PATH}/tools/unit-test-app/components")
endif()
```

alisitsyn (approved):
@david-cermak , the changes LGTM!

components/eppp_link/test/test_app/CMakeLists.txt
david-cermak: Thanks, I'll probably use the same method -- wasn't sure if we can `include($ENV{IDF_PATH}/tools/cmake/project.cmake)` after or before setting `EXTRA_COMPONENT_DIRS`

alisitsyn: > wasn't sure if we can include($ENV{IDF_PATH}/tools/cmake/project.cmake)

I don't think there can be an issue, this works.

PR #874, opened 2025-08-26
feat(websocket): Create specific Event for PONG message

### LINKED ISSUE #777: esp_websocket_client: filter client received PONG response (or add event...) (IDFGH-14804)
### Is your feature request related to a problem?

Currently when esp_websocket_client_recv() processes a PONG response it will be dispatched to the client event handler as a WEBSOCKET_EVENT_DATA with NULL data and zero length.

### Describe the solution you'd like.

I would like to propose 3 options:
1. Filter it out all together:
```
/**
* esp_websocket_client_recv()
*/
        if (client->last_opcode != WS_TRANSPORT_OPCODES_PONG) {
            esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DATA, client->rx_buffer, rlen);
        } else {
            ESP_LOGD(TAG, "Skipping event dispatch for PONG frame.");
        }
```
2. Add a specific WEBSOCKET_EVENT_PONG to esp_websocket_event_id_t to allow it to be properly identified and handled by the client event function:
```
        if (client->last_opcode != WS_TRANSPORT_OPCODES_PONG) {
            esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DATA, client->rx_buffer, rlen);
        } else {
            esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_PONG, client->rx_buffer, rlen);
        }
```
3. Add a define/config option for user choice.

### Describe alternatives you've considered.

See above

### Additional context.

I chose option 1 as I have no need to process PONG responses at an app level. But someone might find value in the distinct ability to process PONGs at an application level...

Just to note, the AsyncWebSocket (in EAPAsyncWebServer) bubbles up those events:
```
typedef enum {
  WS_EVT_CONNECT,
  WS_EVT_DISCONNECT,
  WS_EVT_PING,
  WS_EVT_PONG,
  WS_EVT_ERROR,
  WS_EVT_DATA
} AwsEventType;
```

### LINKED ISSUE #858: Websocket client doesn't handle transport timeout errors correctly (IDFGH-16202)
### Answers checklist.

### General issue report

Might be related to #795.

### What i've gathered

The websocket client uses esp_transport_ws for it's transport layer. When using the esp_transport_read() function, you can use a timeout. This is configureable in the websocket client as 'network_timeout_ms'. When this timeout is reached, the esp_transport_read() function returns 0. This is however also the case for length 0 messages like PING or CLOSE. This already makes it hard to detect a timed out message. It seems that when this timeout is reached, an error isn't correctly detected, and we handle the message as regular. After this happens, the websocket client keeps publishing offset/randomly fragmented data.

### Log

See log below. When the timeout is reached, the function returns 0. After this though, we don't handle this as a timeout and still emit the data event. This is what the client_handler is subscribed to. After this happens once, we enter a weird 'fragmented read' state where the client is confused about what bit is where. Also the websocket headers get shuffled around.

Funnily enough, this confused/fragmented read state continues untill 'E (294271) transport_ws: Error read data(0)' happens again. Then it recovers. This is clearly linked to the network_timeout_ms, as i can control the frequency of this error by enlongating/shortening the timeout time. Shorter timeout increases the frequency.

```
I (292621) client_handler: WS received data: opcode=1, len=63, data={"message":"type","data":{"data":{"point":3262,"point2":1952}}}
E (294271) transport_ws: Error read data(0)
E (294271) transport_ws: Error reading payload data(0)

-> len=0 && opcode=1 -> timeout reached but still publishes event

I (294271) client_handler: WS received data: opcode=1, len=0, data= 
E (294281) client_handler: on_data_event(467): Received unexpected text data

-> After this, the client publishes weirdly fragmented data:

I (295091) client_handler: WS received data: opcode=11, len=34, data={"message":"type","data":{"data"
W (295091) client_handler: WS unhandled op code 11
I (295101) client_handler: WS received data: opcode=10, len=30, data=:{"point":3262,"point2":1952}}}
I (295121) client_handler: WS received data: opcode=10, len=2, data=�U
I (295121) client_handler: WS received data: opcode=10, len=56, data={"message":"type","data":{"data":{"point":3262,"point2":
I (295131) client_handler: WS received data: opcode=14, len=7, data=1952}}}
W (295141) client_handler: WS unhandled op code 14
I (295611) client_handler: WS received data: opcode=14, len=2, data=�S
W (295611) client_handler: WS unhandled op code 14
I (295871) client_handler: WS received data: opcode=14, len=63, data={"message":"type","data":{"data":{"point":3262,"point2":1952}}}
W (295881) client_handler: WS unhandled op code 14

-> Another timeout

E (296401) transport_ws: Error read data(0)
E (296401) transport_ws: Error reading payload data(0)
I (296401) client_handler: WS received data: opcode=14, len=0, data=
W (296411) client_handler: WS unhandled op code 14

-> State is restored again

I (296711) app_remote: WS received data: opcode=1, len=63, data={"message":"type","data":{"data":{"point":3262,"point2":1952}}}
```

I had to obfuscate the actual data being send, so maybe the lengths don't match exactly.

glmfe:
## Description

- Separate the PONG message from generic payload, this can make it easier to identify when a timeout occur when receiving a payload

## Related
- Fixes #777
- Fixes #858

### DIFF
--- a/components/esp_websocket_client/examples/target/main/websocket_example.c
+++ b/components/esp_websocket_client/examples/target/main/websocket_example.c
@@ -79,6 +79,8 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
         break;
     case WEBSOCKET_EVENT_CONNECTED:
         ESP_LOGI(TAG, "WEBSOCKET_EVENT_CONNECTED");
+        // Optional: Send ping to keep the connection alive.
+        esp_websocket_client_send_with_opcode((esp_websocket_client_handle_t)handler_args, WS_TRANSPORT_OPCODES_PING, NULL, 0, portMAX_DELAY);
         break;
     case WEBSOCKET_EVENT_DISCONNECTED:
         ESP_LOGI(TAG, "WEBSOCKET_EVENT_DISCONNECTED");
@@ -115,6 +117,9 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
         ESP_LOGW(TAG, "Total payload length=%d, data_len=%d, current payload offset=%d\r\n", data->payload_len, data->data_len, data->payload_offset);
 
         xTimerReset(shutdown_signal_timer, portMAX_DELAY);
+        break;
+    case WEBSOCKET_EVENT_PONG:
+        ESP_LOGI(TAG, "WEBSOCKET_EVENT_PONG");
         break;
     case WEBSOCKET_EVENT_ERROR:
         ESP_LOGI(TAG, "WEBSOCKET_EVENT_ERROR");
--- a/components/esp_websocket_client/include/esp_websocket_client.h
+++ b/components/esp_websocket_client/include/esp_websocket_client.h
@@ -34,6 +34,7 @@ typedef enum {
     WEBSOCKET_EVENT_CONNECTED,      /*!< Once the Websocket has been connected to the server, no data exchange has been performed */
     WEBSOCKET_EVENT_DISCONNECTED,   /*!< The connection has been disconnected */
     WEBSOCKET_EVENT_DATA,           /*!< When receiving data from the server, possibly multiple portions of the packet */
+    WEBSOCKET_EVENT_PONG,           /*!< When receiving a PONG control frame from the server, Pong events might carry payload when present. */
     WEBSOCKET_EVENT_CLOSED,         /*!< The connection has been closed cleanly */
     WEBSOCKET_EVENT_BEFORE_CONNECT, /*!< The event occurs before connecting */
     WEBSOCKET_EVENT_BEGIN,          /*!< The event occurs once after thread creation, before event loop */
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -990,7 +990,11 @@ static esp_err_t esp_websocket_client_recv(esp_websocket_client_handle_t client)
             return ESP_OK;
         }
 
-        esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DATA, client->rx_buffer, rlen);
+        if (client->last_opcode == WS_TRANSPORT_OPCODES_PONG) {
+            esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_PONG, client->rx_buffer, rlen);
+        } else {
+            esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DATA, client->rx_buffer, rlen);
+        }
 
         client->payload_offset += rlen;
     } while (client->payload_offset < client->payload_len);
--- a/components/esp_websocket_client/examples/linux/main/websocket_linux.c
+++ b/components/esp_websocket_client/examples/linux/main/websocket_linux.c
@@ -30,6 +30,7 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
         break;
     case WEBSOCKET_EVENT_CONNECTED:
         ESP_LOGI(TAG, "WEBSOCKET_EVENT_CONNECTED");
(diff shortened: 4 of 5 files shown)


components/esp_websocket_client/examples/target/main/websocket_example.c
```diff
@@ -114,6 +115,12 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
 
         ESP_LOGW(TAG, "Total payload length=%d, data_len=%d, current payload offset=%d\r\n", data->payload_len, data->data_len, data->payload_offset);
 
+        xTimerReset(shutdown_signal_timer, portMAX_DELAY);
+        break;
+    case WEBSOCKET_EVENT_PONG:
+        ESP_LOGI(TAG, "WEBSOCKET_EVENT_PONG");
+        ESP_LOGI(TAG, "Received opcode=%d", data->op_code);
```
david-cermak: Do we need to print the opcode? It would be always `PONG`

components/esp_websocket_client/examples/target/main/websocket_example.c
```diff
@@ -114,6 +115,12 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
...
+        // Reset shutdown timer on PONG as well
```
david-cermak: Question: Why we're resting the timer after "PONG"? Would the test shutdown if we didn't?

components/esp_websocket_client/examples/target/main/websocket_example.c
```diff
@@ -79,6 +79,7 @@ static void websocket_event_handler(void *handler_args, esp_event_base_t base, i
         break;
     case WEBSOCKET_EVENT_CONNECTED:
         ESP_LOGI(TAG, "WEBSOCKET_EVENT_CONNECTED");
+        esp_websocket_client_send_with_opcode((esp_websocket_client_handle_t)handler_args, WS_TRANSPORT_OPCODES_PING, NULL, 0, portMAX_DELAY);
```
david-cermak, line 83: Suggest adding a comment, that this is only needed for testing. Otherwise people would copy-paste this to their project, which I think is not necessary.

components/esp_websocket_client/include/esp_websocket_client.h
```diff
@@ -34,6 +34,7 @@ typedef enum {
     WEBSOCKET_EVENT_CONNECTED,      /*!< Once the Websocket has been connected to the server, no data exchange has been performed */
     WEBSOCKET_EVENT_DISCONNECTED,   /*!< The connection has been disconnected */
     WEBSOCKET_EVENT_DATA,           /*!< When receiving data from the server, possibly multiple portions of the packet */
+    WEBSOCKET_EVENT_PONG,           /*!< When receiving a PONG control frame from the server */
```
david-cermak: Maybe add a note that the event also contains data.

glmfe: Done, thanks!

components/esp_websocket_client/examples/target/main/websocket_example.c
glmfe: Added, Thanks!

components/esp_websocket_client/examples/target/main/websocket_example.c
glmfe: you're right we dont need to reset the timer in this case, thanks

components/esp_websocket_client/examples/target/main/websocket_example.c
glmfe: Removed, thanks!

glmfe:
> I believe this should be marked as a breaking change. Although the event is new, how the users were dealing with the PONG message before?

Currently all PONGS are being directed to the Event "WEBSOCKET_EVENT_DATA"... so this might be a breaking change

PR #908, opened 2025-10-10
[modem]: Fix version v1.4.1 from `release-modem-v1.4` branch

david-cermak:
### Bug Fixes

- Remove unused kconfig ([97194507](https://github.com/espressif/esp-protocols/commit/97194507))
- Fix modem console dependencies ([0d17f19f](https://github.com/espressif/esp-protocols/commit/0d17f19f))
- Fix CI build issues with IDFv6.0 ([d65f8f69](https://github.com/espressif/esp-protocols/commit/d65f8f69))
- Address build issues ([b4520660](https://github.com/espressif/esp-protocols/commit/b4520660))
- Fix driver dependency issue on v6.0 ([26b3859c](https://github.com/espressif/esp-protocols/commit/26b3859c))

### DIFF
--- a/components/esp_modem/.cz.yaml
+++ b/components/esp_modem/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(modem): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py esp_modem
   tag_format: modem-v$version
-  version: 1.4.0
+  version: 1.4.1
   version_files:
   - idf_component.yml
--- a/components/esp_modem/CHANGELOG.md
+++ b/components/esp_modem/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [1.4.1](https://github.com/espressif/esp-protocols/commits/modem-v1.4.1)
+
+### Bug Fixes
+
+- Remove unused kconfig ([97194507](https://github.com/espressif/esp-protocols/commit/97194507))
+- Fix modem console dependencies ([0d17f19f](https://github.com/espressif/esp-protocols/commit/0d17f19f))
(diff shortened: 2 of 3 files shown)


.github/workflows/pre_commit_check.yml
```diff
@@ -1,45 +0,0 @@
-name: Check pre-commit rules
```
euripedesrocha, line 1: Why this is being removed?

.github/workflows/publish-docs-component.yml
```diff
@@ -55,9 +56,6 @@ jobs:
                 if [[ "${{ env.BUMP_COMPONENT }}" == "${comp}" ]]; then
                   echo "Building specific version of ${comp} (${{ env.BUMP_VERSION }})"
                   ./generate_docs ${{ env.BUMP_VERSION }}
-                else
```
euripedesrocha, line 58: Same here, why are you removing?

.github/workflows/pre_commit_check.yml
david-cermak: Just to simplify things [this](https://github.com/espressif/esp-protocols/pull/908/files#diff-b92f3ab0020dcbd84dc507e6898ab704b339cb947dfdbd4a74847c8c2c165fe7L24) translates to `origin/master` so this job fails. I can shim the script to work on this release branch, but found it too complicated. Pre-commit hooks are still working.

.github/workflows/publish-docs-component.yml
david-cermak: This builds the "latest" version (latest on master).
To address this "correctly" we'd have to create a new docs endpoint for this release branch, but I think it's not necessary and we can have the docs only for the "released" versions and for the "latest".

PR #909, opened 2025-10-14
Add esp_driver_uart dependency for new esp-idf (IDFGH-16663)

0xFEEDC0DE64:
newest esp-idf version requires additional dependencies to be added, otherwise it doesnt cmake anymore

### DIFF
--- a/components/esp_modem/CMakeLists.txt
+++ b/components/esp_modem/CMakeLists.txt
@@ -11,7 +11,7 @@ else()
                       src/esp_modem_uart.cpp
                       src/esp_modem_term_uart.cpp
                       src/esp_modem_netif.cpp)
-    set(dependencies driver esp_event esp_netif)
+    set(dependencies driver esp_driver_uart esp_event esp_netif)
 endif()
 
 


components/esp_modem/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ else()
                       src/esp_modem_uart.cpp
                       src/esp_modem_term_uart.cpp
                       src/esp_modem_netif.cpp)
-    set(dependencies driver esp_event esp_netif)
+    set(dependencies driver esp_driver_uart esp_event esp_netif)
```
david-cermak, line 14: LGTM for the newer IDF versions, but we need to support also the older one, so some ifdef-ing is necessary.

This is already fixed on [`v14.1`](https://github.com/espressif/esp-protocols/releases/tag/modem-v1.4.1), if you like you can cherry-pick https://github.com/espressif/esp-protocols/commit/26b3859cfb4a31a03400aeb67346c4aba1bedbe1 and we can merge it.

david-cermak:
> it feels like merge-requesting to esp-protocols is a much better experience than merge-requesting anything to esp-idf. 

Thank you, appreciate your kind feedback 👍 

> At esp-idf they started using AI bots to do even less reviews than before and a ci job is blocking almost all commit messages and usually merge requests take years 

Sorry about that, I'll try to help with developer's experience in IDF too, understand the effort PR authors put into fixing, contributing and sharing their work and the frustration when the only feedback is an automated `"you're absolutely right"`-kind of message...

PR #913, opened 2025-10-17
ci(common): Ignore ethernet warning

glmfe:
## Description
 - Currently the "latest" docker image is outdated causing some warnings regarding the eth driver configs.


.github/workflows/websocket__build-target-test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Build
     strategy:
       matrix:
-        idf_ver: ["release-v5.0", "release-v5.1", "release-v5.2", "release-v5.3", "latest"]
+        idf_ver: ["latest"]
```
david-cermak: I think we should keep the active IDF releases. at least for the build steps

gabsuren: Agreed. @glmfe , could you please clarify what's the reason for removing it from the running step as well? Is this a temporary change?

glmfe: Yes, we need to keep all of them before merging. I only removed them to get a quick result from the CI, since esptool is failing when running on 'latest' and I couldn’t test it locally

glmfe: reverted 😄

PR #922, opened 2025-10-27
fix(modem): CEREG parsing fails due to incorrect state index (IDFGH-16681) (IDFGH-16682)

### LINKED ISSUE #923: Bug(esp_modem): get_network_registration_state fails to parse state value from AT+CEREG? response (IDFGH-16683)
## Description
The function get_network_registration_state fails to correctly parse the network registration status from the AT+CEREG? command response (e.g., +CEREG: 2,1).

## Steps to Reproduce
Call get_network_registration_state.

The underlying AT command returns a valid response like +CEREG: 2,1.

The function attempts to use std::from_chars to parse the state.

### Expected Behavior
The function should correctly extract the integer state (e.g., 1) and return command_result::OK.

### Actual Behavior
The function fails because it attempts to parse the state value starting at the preceding comma delimiter (e.g., attempting to parse ,1 instead of 1), leading to an std::errc::invalid_argument error from std::from_chars. The function incorrectly returns command_result::FAIL.

## Proposed Solution
The parsing index needs to be adjusted by one (+ 1) to skip the comma delimiter, ensuring std::from_chars starts at the first digit of the state value.

JellevanKraaij:
## Description
This PR fixes a bug in get_network_registration_state where parsing the AT command response +CEREG: <mode>,<stat>,... failed because std::from_chars was attempting to parse the number starting at the preceding comma.

The fix involves incrementing the start position by one (+ 1) before calling std::from_chars to correctly skip the leading comma delimiter.

## Related

Fixes #923 

## Testing

Verified that the function now correctly extracts the <stat> integer from typical responses like +CEREG: 2,1

### DIFF
--- a/components/esp_modem/src/esp_modem_command_library.cpp
+++ b/components/esp_modem/src/esp_modem_command_library.cpp
@@ -641,6 +641,8 @@ command_result get_network_registration_state(CommandableIf *t, int &state)
         return command_result::FAIL;
     }
 
+    state_pos_start += 1; // move past the comma
+
     if (out.find(pattern) == std::string::npos || (state_pos_end = out.find(',', state_pos_start)) == std::string::npos) {
         if (std::from_chars(out.data() + state_pos_start, out.data() + out.size(), state).ec == std::errc::invalid_argument) {
             return command_result::FAIL;


david-cermak:
> When will this be available in esp registry 

just released, should be available already

> Did not see it in main, sorry!

I am sorry, should have backported before `v1.4.1`.
I'll create `v2.0` soon

Issue #898, opened 2025-09-30
WEBSOCKET Double free in esp_websocket_client during error/teardown race (IDFGH-16555)


filzek:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### What component are you using? If you choose Other, provide details in More Information.

esp_websocket_client

### component version

1.5.0

### IDF version.

v5.4.1-18-g6f30fb434d

### More Information.

Hi @glmfe, the websocket double free/tls exception still happening frequently.

Description
When a WebSocket connection aborts unexpectedly, the client sometimes crashes with a TLSF assertion failure indicating a double free. This appears to happen because both the error path and the send/abort path attempt to close and free the same TLS transport, leading to heap corruption.

Observed Behavior

Connection is aborted (errno=113, mbedtls read error :-0x004C).

WEBSOCKET_EVENT_ERROR is fired.

Application code attempts to send or abort, which calls into esp_websocket_client_abort_connection().

The transport is closed twice (esp_transport_close appears twice in the backtrace).

TLS tries to free already-freed memory → assert failed: tlsf_free ... "block already marked as free".

Device reboots with heap corruption.

Backtrace (excerpt)

assert failed: tlsf_free tlsf.c:629 (!block_is_free(block) && "block already marked as free")
esp_tls_conn_destroy -> base_close -> esp_transport_close
ws_close -> esp_transport_close
esp_websocket_client_abort_connection
esp_websocket_client_send_text
EWC_process_queue_tosend
...

Expected Behavior

The WebSocket client should handle error/teardown idempotently.

A single close/free of the TLS transport should occur, even if application code calls abort/stop concurrently with internal error handling.

No heap corruption or double free should be possible.

Steps to Reproduce

Establish a TLS WebSocket connection.

Force the server to close the connection abruptly (RST).

Allow the client task to attempt a send just after the error event is triggered.

Observe crash with TLSF double free assertion.

Environment

ESP-IDF: special build (14.2_20240403 toolchain)

Component: esp_websocket_client + esp-tls + tcp_transport

Target: ESP32 (WROVER module)

Impact
This results in device resets in production when WebSocket connections drop under load or unstable networks.

Possible Root Cause
[...]

PR #924, opened 2025-10-28
fix(websocket): Fix websocket client race on abort and memory leak(IDFGH-16555)

gabsuren:
## Description

This PR fixes critical memory leaks and crashes in the ESP WebSocket client that occur during reconnection scenarios(CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK = y).

- **Double-free crashes**: Heap corruption during abort/reconnect scenarios  
- **Data loss**: First packet after reconnection not received
- **Error buffer accumulation**: 2KB memory leak on disconnect

**Changes Made:**
- Add state check in abort_connection to prevent double-close
- Fix memory leak: free errormsg_buffer on disconnect
- Reset connection state on reconnect to prevent stale data
- Implement lock ordering for separate TX lock mode
- Added sdkconfig.ci.tx_lock conf

## Related
https://github.com/espressif/esp-protocols/issues/898

## Checklist

- [✓] 🚨 This PR does not introduce breaking changes.
- [ ✓ ] All CI checks (GH Actions) pass.
- [ ✓] Documentation is updated as needed.
- [✓ ] Tests are updated or added as necessary.
- [ ✓] Code is well-commented, especially in complex areas.
- [ ✓] Git history is clean — commits are squashed to the minimum necessary.

---

> [!NOTE]
> Adds state-guarded abort, frees error buffer on disconnect, resets connection state on connect, and enforces safe TX lock ordering (SEPARATE_TX_LOCK) across send/ping/pong/close paths; includes example sdkconfig.
> 
> - **WebSocket client core**:
>   - Add `esp_websocket_client_abort_connection()` state checks to avoid double-close; dispatch disconnect and free `errormsg_buffer` in cleanup.
>   - Reset `payload_len/offset`, `last_fin`, `last_opcode` on connect; process pending handshake data via `esp_websocket_client_recv()`.
> - **Locking (CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK)**:
>   - Enforce lock ordering: release `lock`, acquire `tx_lock` with timeout, re-acquire `lock`, revalidate state before sending PING/PONG/CLOSE.
>   - On send errors, switch from `tx_lock` to `lock` before abort; call abort and return safely.
> - **I/O loop adjustments**:
>   - Guard poll/read error paths with `lock` before aborting; minor reordering to prevent races.
> - **Cleanup**:
>   - After destroying transport list, set `client->transport_list` and `client->transport` to `NULL`.
> - **Examples/Config**:
>   - Add `examples/target/sdkconfig.ci.tx_lock` enabling `CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK` with timeout.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23ca97d5eccff866e89cb880df3e8366bf7d12b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -244,6 +244,16 @@ static esp_err_t esp_websocket_client_dispatch_event(esp_websocket_client_handle
 static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_handle_t client, esp_websocket_error_type_t error_type)
 {
     ESP_WS_CLIENT_STATE_CHECK(TAG, client, return ESP_FAIL);
+
+    // Note: This function must be called with client->lock already held
```
euripedesrocha: If this is a pre-condition for the function, should we add a guard here? If a guard isn't possible this **must** be documented to user and not be a comment inside the function.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -256,6 +266,17 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
     }
     client->error_handle.error_type = error_type;
     esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DISCONNECTED, NULL, 0);
+
+    if (client->errormsg_buffer) {
+        ESP_LOGI(TAG, "Freeing error buffer (%d bytes) - Free heap: %" PRIu32 " bytes",
```
euripedesrocha: This should be Debug or verbose level.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -256,6 +266,17 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
...
+                 client->errormsg_size, esp_get_free_heap_size());
+        free(client->errormsg_buffer);
+        client->errormsg_buffer = NULL;
+        client->errormsg_size = 0;
+    } else {
+        ESP_LOGI(TAG, "Disconnect - Free heap: %" PRIu32 " bytes", esp_get_free_heap_size());
```
euripedesrocha: Same here.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -679,8 +702,18 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
             } else {
                 esp_websocket_client_error(client, "esp_transport_write() returned %d, errno=%d", ret, errno);
             }
+            ESP_LOGD(TAG, "Calling abort_connection due to send error");
+#ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
+            xSemaphoreGiveRecursive(client->tx_lock);
```
euripedesrocha: It is better to move this verification to abort connection function.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1030,15 +1062,47 @@ static esp_err_t esp_websocket_client_recv(esp_websocket_client_handle_t client)
         const char *data = (client->payload_len == 0) ? NULL : client->rx_buffer;
         ESP_LOGD(TAG, "Sending PONG with payload len=%d", client->payload_len);
 #ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
+        // CRITICAL: To avoid deadlock, we must follow lock ordering: tx_lock BEFORE client->lock
```
euripedesrocha: These comments are unnecessary.

components/esp_websocket_client/esp_websocket_client.c
gabsuren: @euripedesrocha I am not sure about it, as abort connection is used in 5 different places, if we move it inside abort_connection it adds complex detection logic "which lock am I holding?" Only 1 place (send error path) needs lock switching (tx_lock → client->lock)

components/esp_websocket_client/esp_websocket_client.c
gabsuren: Good point, updated the documentation

gabsuren:
https://github.com/espressif/esp-protocols/issues/898#issuecomment-3539579537

components/esp_websocket_client/esp_websocket_client.c
euripedesrocha: You are right in that sense but looking at it again it might be better to release the tx_lock after acquiring the main lock maybe at the end of this section after we complete the abort process to avoid task to be preempted.

gabsuren: Good point. Thank you! Updated accordingly

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -679,8 +712,18 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
             } else {
                 esp_websocket_client_error(client, "esp_transport_write() returned %d, errno=%d", ret, errno);
             }
+            ESP_LOGD(TAG, "Calling abort_connection due to send error");
+#ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
+            xSemaphoreTakeRecursive(client->lock, portMAX_DELAY);
+            xSemaphoreGiveRecursive(client->tx_lock);
```
david-cermak: Question: shouldn't we better unlock the tx-lock before locking the main one? smells of a deadlock to me, but just asking

gabsuren: @david-cermak Thank you for the good question! I was checking the scenarios. 

If we reverse the order, we'd have this race condition:

1. Thread A releases tx_lock
2. Thread B acquires tx_lock, checks is_connected() → returns TRUE (state not changed yet)
3. Thread B starts sending on transport
4. Thread A acquires client->lock, calls abort_connection(), closes transport
5. Thread B is now writing to a closed transport → crash/undefined behavior

The brief overlap (holding both locks) prevents this race by ensuring the state change 
happens atomically from other threads' perspective.

**Fix applied:** PING and CLOSE now release client->lock before acquiring tx_lock 
(matching the PONG pattern), which breaks the circular wait and prevents deadlock 
while maintaining race-free operation.

I also considered changing the connection state before abort_connection(), but that 
would have race conditions as well without proper locking.

david-cermak: okay, got it, thanks for checking and updating the ping/close case!

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -256,6 +276,18 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
     }
     client->error_handle.error_type = error_type;
     esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DISCONNECTED, NULL, 0);
+
+cleanup:
+    if (client->errormsg_buffer) {
+        ESP_LOGD(TAG, "Freeing error buffer (%d bytes) - Free heap: %" PRIu32 " bytes",
```
euripedesrocha:
```suggestion
        ESP_LOGD(TAG, "Freeing error buffer (%d bytes) - Free heap before: %" PRIu32 " bytes",
```

PR #925, opened 2025-10-29
fix(websocket): resolve JSON dependency issues for component and exam…

gabsuren:


components/esp_websocket_client/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 if(${IDF_TARGET} STREQUAL "linux")
 	idf_component_register(SRCS "esp_websocket_client.c"
                     INCLUDE_DIRS "include"
-                    REQUIRES esp-tls tcp_transport http_parser esp_event nvs_flash esp_stubs json
+                    REQUIRES esp-tls tcp_transport http_parser esp_event nvs_flash esp_stubs
```
david-cermak: Unrelated to this change, but maybe we should cleanup this list a little. `nvs_flash` and `esp_stubs` are required?

gabsuren: @david-cermak  Good point, removed. Thank you for review!

PR #930, opened 2025-11-03
fix(websocket): fix PING timing - enable periodic PING during active traffic (IDFGH-16701)

gabsuren:
> [!NOTE]
> Stops resetting `client->ping_tick_ms` on read poll timeout so periodic PINGs proceed on schedule during active traffic.
> 
> - **WebSocket client (`components/esp_websocket_client/esp_websocket_client.c`)**:
>   - Adjust PING scheduling in `WEBSOCKET_STATE_CONNECTED` by removing `client->ping_tick_ms = _tick_get_ms()` after read poll timeout, preventing unintended PING delays.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f424325d89d9d69a889bb1b2735ef70dd47ac84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1174,7 +1174,6 @@ static void esp_websocket_client_task(void *pv)
                 ESP_LOGV(TAG, "Read poll timeout: skipping esp_transport_read()...");
                 break;
             }
-            client->ping_tick_ms = _tick_get_ms();
             break;
         case WEBSOCKET_STATE_WAIT_TIMEOUT:
 


david-cermak (approved):
LGTM, thanks for the fix!

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1174,7 +1174,6 @@ static void esp_websocket_client_task(void *pv)
                 ESP_LOGV(TAG, "Read poll timeout: skipping esp_transport_read()...");
                 break;
             }
-            client->ping_tick_ms = _tick_get_ms();
```
david-cermak, line 1177: Question: Should we make it configurable? I mean it makes sense for some usecases -- saves unnecessary PING traffic when sock is readable?

gabsuren: @david-cermak I was considering it, but per [RFC 6455 Section 5.5.2-5.5.3](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2), PING is for "keepalive or verifying remote endpoint is still responsive" - so no restriction on idle-only according to RFC.

And the current implementation users already have 3 options:
Users already have 3 options:
> 1. Full monitoring (default): PING + timeout enforcement
> 2. Monitoring without disconnect: PING + disable_pingpong_discon=true
> 3. No monitoring: ping_interval_sec = 0

| Config | PING Sent? | Timeout Check? | Disconnect on No PONG? |
|--------|-----------|----------------|------------------------|
| `ping_interval_sec = 10` (default) | ✅ Every 10s | ✅ 120s | ✅ Yes |
| `disable_pingpong_discon = true` | ✅ Every 10s | ❌ No | ❌ No |
| `ping_interval_sec = 0` | ❌ Disabled | ❌ No | ❌ No |

**Reasons**:
1. ✅ Users already have 3 configuration levels (see table above)
2. ✅ Keepalive traffic is minimal (~6 bytes every 10s = 0.6 bytes/s)
3. ✅ RFC 6455 recommends periodic keepalive for connection health


What Do You Think?

gabsuren: However, if you think the "skip PING during traffic" behavior has value, I can add explicit `skip_ping_on_traffic` configuration option

david-cermak: Maybe we can just enclose it within an undefined macro. So if we users want to preserve this feature, they just `add_compile_defs()` in their makefile.

On the other hand I also see that it adds complexity. I'll leave the final decision to you, agree with both approaches -- feel free to resolve, consider the PR approved from my end.

gabsuren: @david-cermak thank you for the review! I will push it as it is now. If users report issues with PING traffic during active connections, then will add it in a seperate PR

PR #943, opened 2025-11-14
fix(mdns): mdns_browse_delete works as intended (IDFGH-16796)

benedekkupper:
## Description

Not much to say, the `mdns_browse_delete()` didn't have a proper API nor a correct implementation, it was creating an object for itself that it deleted later on, the whole method was mostly copy-paste from `mdns_browse_new()`.

## Checklist


benedekkupper:
I see now that mdns_console.c is using the current API of this method, so this PR isn't valid. However, what's still missing is a way to correctly free the `mdns_browse_t` objects that these two calls are creating. As I understand it, these are created on the heap, and there's no way to free the object returned by `mdns_browse_new()`.

PR #945, opened 2025-11-18
Add mqtt dependency to component's projects

david-cermak:
* Use esp-mqtt as optional dependency for
  * `console_cmd_mqtt`
  * `esp_mqtt_cxx`
* Workaround `ethernet-init` issues for `console_cmd_ifconfig`
* Use current IDF releases for console CI build targets

### Future PR:

- [ ] Bump `console_cmd_mqtt`
- [ ] Bump `console_cmd_ifconfig`
- [ ] Bump `esp_mqtt_cxx`

### DIFF
--- a/components/console_cmd_mqtt/idf_component.yml
+++ b/components/console_cmd_mqtt/idf_component.yml
@@ -9,3 +9,7 @@ dependencies:
     version: '>=1.1.0'
     override_path: '../console_simple_init'
     public: true
+  espressif/mqtt:
+    rules:
+    - if: idf_version >=6.0
+      version: ^1.0.0


components/console_cmd_mqtt/idf_component.yml
```diff
@@ -9,3 +9,7 @@ dependencies:
     version: '>=1.1.0'
     override_path: '../console_simple_init'
     public: true
+  espressif/mqtt:
+    rules:
+    - if: idf_version >=6.0
```
euripedesrocha, line 14: You can use the component from idf > 5.3 (if I'm not mistaken) No need to update this though.

david-cermak: Thanks, I'll keep it as is (for simplicity, just to have only one mqtt in build)

PR #958, opened 2025-11-26
fix(mdns): Fix to keep TXT/SRV in answers to queries

david-cermak:
* unlike discoveries, where they need to stay in additional section
* Quick fix of regression b7b8c5dbd7433497d806d40c53d0cfcea8dfaebb
---

The hotfix makes _mdns_create_answer_from_service() aware of whether a query is service discovery or an instance-level Q/A without broad architectural changes.
We detect instance questions via question->host and introduce is_instance_question / is_any_instance_question. For service-browse queries (QTYPE=PTR or QTYPE=ANY where there is no instance label, e.g. _service._tcp.local), behavior remains: the PTR goes in the Answer section, and SRV/TXT plus A/AAAA go into the Additional section, matching RFC 6763 §12.1 and Thread expectations.
For instance-level ANY queries (Instance._service._tcp.local, QTYPE=ANY), we now treat them like SRV/TXT Q/A: SRV and TXT for that instance are placed in the Answer section, with A/AAAA in Additional and no unrelated PTR answer created. Direct SRV/TXT queries are unchanged. This should make Bonjour Conformance Test happier while preserving the discovery behavior that Thread certification requires.

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1791,7 +1791,21 @@ static bool _mdns_create_answer_from_service(mdns_tx_packet_t *packet, mdns_serv
 {
     mdns_host_item_t *host = mdns_get_host_item(service->hostname);
     bool is_delegated = (host != &_mdns_self_host);
-    if (question->type == MDNS_TYPE_PTR || question->type == MDNS_TYPE_ANY) {
+    bool is_instance_question = !_str_null_or_empty(question->host);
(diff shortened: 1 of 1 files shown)


david-cermak:
Fixed in 0f6235f13e6dd6bf1bd7ccd1fd4064324c5ac352 (on master only, not `v1.9` as we're still on `v1.x` -- no breaking changes as of now)

PR #959, opened 2025-11-27
[mdns]: v1.10: Fix to keep TXT/SRV in answers to queries

david-cermak:
unlike discoveries, where they need to stay in additional section
Quick fix of regression b7b8c5dbd7433497d806d40c53d0cfcea8dfaebb

### DIFF
--- a/components/mdns/mdns_send.c
+++ b/components/mdns/mdns_send.c
@@ -257,9 +257,20 @@ static bool create_answer_from_service(mdns_tx_packet_t *packet, mdns_service_t 
 {
     mdns_host_item_t *host = get_host_item(service->hostname);
     bool is_delegated = (host != mdns_priv_get_self_host());
-    if (question->type == MDNS_TYPE_PTR || question->type == MDNS_TYPE_ANY) {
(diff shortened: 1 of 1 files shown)


components/mdns/mdns_send.c
```diff
@@ -271,6 +273,13 @@ static bool create_answer_from_service(mdns_tx_packet_t *packet, mdns_service_t
                                          send_flush, false)) {
             return false;
         }
+    } else if (is_any_instance_question) {
```
tanyanquan: Maybe we can put this block first, followed by 
`else if (question->type == MDNS_TYPE_PTR || question->type == MDNS_TYPE_ANY)` 
for clarity

zwx1995esp: How about using the original expression `(question->type == MDNS_TYPE_ANY) && is_instance_question` , just to follow consistency with other if conditions, and adding a comment in this if branch?

zwx1995esp (approved):
LGTM, 👍

components/mdns/mdns_send.c
david-cermak: Good point, have swapped these branches and removed the unnecessary `is_any_instance_question`.
thanks!

PR #973, opened 2025-12-15
fix(mosq): Fix build with the new picolibc

david-cermak:
* Uses `LIBC_NEWLIB` by default
* Handles `char` indices

Fixes: https://github.com/espressif/esp-protocols/actions/runs/20132443163/job/57777031475


components/mosquitto/examples/broker/sdkconfig.defaults
```diff
@@ -0,0 +1 @@
+CONFIG_LIBC_NEWLIB=y
```
euripedesrocha, line 1: Is mosquitto compatible only with Newlib? If so we need to add this to documentation as warning to users.

david-cermak: This is a temporary fix for now. Created IDF MR!44230 to fix it properly, but this needs to get deployed and yet propagate to the installed tools.
I'll create a GH issue to remove after fixed upstream.

david-cermak: created https://github.com/espressif/esp-protocols/issues/976

euripedesrocha (approved):

PR #974, opened 2025-12-15
[asio]: Fix picolib missing pthread_sigmask() declaration

david-cermak:
Another picolib related
* potentially missing `pthread_sigmask()` declarations
* fixes https://github.com/espressif/esp-protocols/actions/runs/20132443182/job/57777032103#step:4:1484

### DIFF
--- a/components/asio/port/include/asio_stub.hpp
+++ b/components/asio/port/include/asio_stub.hpp
@@ -1,0 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+#include <signal.h>
+
+extern "C" int pthread_sigmask(int, const sigset_t *, sigset_t *);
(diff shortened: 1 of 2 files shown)


components/asio/port/include/asio_stub.hpp
```diff
@@ -0,0 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+#include <signal.h>
+
+extern "C" int pthread_sigmask(int, const sigset_t *, sigset_t *);
```
euripedesrocha, line 9: No need to define it?

david-cermak: It's just a declaration issue, I'd actually defined in all (supported) IDF versions

david-cermak: This is being fixed in !44246 (but the same story as https://github.com/espressif/esp-protocols/pull/973)

david-cermak: created https://github.com/espressif/esp-protocols/issues/976

PR #975, opened 2025-12-15
fix(websocket): prevent connection abort on write timeout (IDFGH-16792)

gabsuren:
> [!NOTE]
> Prevents unintended disconnects when a send operation times out without an underlying socket error.
> 
> - In `esp_websocket_client_send_with_exact_opcode`, set `errno = 0` before `esp_transport_ws_send_raw` and, if it returns `wlen == 0` on the first chunk with finite `timeout` and `errno == 0`, return early (no abort) instead of treating it as a fatal error.
> - Keeps connection intact and avoids calling `esp_websocket_client_abort_connection` on benign write timeouts; no API changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 091324b23086a591df6c1eb6777d34b435431ccf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -700,9 +700,15 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
         }
         memcpy(client->tx_buffer, data + widx, need_write);
         // send with ws specific way and specific opcode
+        errno = 0;
         wlen = esp_transport_ws_send_raw(client->transport, opcode, (char *)client->tx_buffer, need_write,
                                          (timeout == portMAX_DELAY) ? -1 : timeout * portTICK_PERIOD_MS);
         if (wlen < 0 || (wlen == 0 && need_write != 0)) {
+            if (wlen == 0 && widx == 0 && timeout != portMAX_DELAY && errno == 0) {
+                ret = widx;
+                esp_websocket_free_buf(client, true);
+                goto unlock_and_return;
+            }
             ret = wlen;
             esp_websocket_free_buf(client, true);
 


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -700,9 +700,15 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
         }
         memcpy(client->tx_buffer, data + widx, need_write);
         // send with ws specific way and specific opcode
+        errno = 0;
         wlen = esp_transport_ws_send_raw(client->transport, opcode, (char *)client->tx_buffer, need_write,
                                          (timeout == portMAX_DELAY) ? -1 : timeout * portTICK_PERIOD_MS);
         if (wlen < 0 || (wlen == 0 && need_write != 0)) {
+            if (wlen == 0 && widx == 0 && timeout != portMAX_DELAY && errno == 0) {
```
euripedesrocha, line 707: Could you double-check errno value when we time out on write? LWIP layer sets it to EINPROGRESS(119).

Another thing that I'm missing is the return of widx here, we are in between write but testing for zero. We're changing the return for partial write but why we don't have any update on the call site for the function? Does it impact user code after this change?

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -700,9 +700,15 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
...
+                ret = widx;
+                esp_websocket_free_buf(client, true);
+                goto unlock_and_return;
```
Bryght-Labs-Richard, line 710: This seems like it might leave a partial WebSocket frame in-flight during timeout. If so, how do we plan to resynchronize the binary stream before any other callers attempt to transmit through the client?

I bring this up because as soon as the lock is released, another thread may attempt to send a different packet, but its header may overlap the in-flight payload, causing framing corruption.

gabsuren: @bryghtlabs-richard You are absolutely right. 
`esp_transport_ws_send_raw` sends the header and payload in separate writes, so the operation isn't atomic. A timeout could occur after the header is sent but before the payload, leaving the stream in a corrupted state where the server interprets the next message's header as the previous payload.
Since the API doesn't distinguish between "nothing sent" and "partial write," aborting the connection is the only way to ensure protocol integrity. 
I'll close this PR and investigate a safer fix at the transport layer.

Thanks for noticing!

PR #979, opened 2025-12-29
bump(websocket): 1.6.0 -> 1.6.1

gabsuren:
1.6.1
Bug Fixes
- Fix race conditions, memory leak, and data loss (23ca97d5)
    - Add state check in abort_connection to prevent double-close
    - Fix memory leak: free errormsg_buffer on disconnect
    - Reset connection state on reconnect to prevent stale data
    - Implement lock ordering for separate TX lock mode
    - Read buffered data immediately after connection to prevent data loss
    - Added sdkconfig.ci.tx_lock config

## Checklist

- [ *] 🚨 This PR does not introduce breaking changes.
- [ *] All CI checks (GH Actions) pass.
- [ *] Documentation is updated as needed.
- [ *] Tests are updated or added as necessary.
- [ *] Code is well-commented, especially in complex areas.
- [ *] Git history is clean — commits are squashed to the minimum necessary.

### DIFF
--- a/components/esp_websocket_client/CHANGELOG.md
+++ b/components/esp_websocket_client/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [1.6.1](https://github.com/espressif/esp-protocols/commits/websocket-v1.6.1)
+
+### Bug Fixes
+
+- Fix race conditions, memory leak, and data loss ([23ca97d5](https://github.com/espressif/esp-protocols/commit/23ca97d5))
+    - Add state check in abort_connection to prevent double-close
+    - Fix memory leak: free errormsg_buffer on disconnect
+    - Reset connection state on reconnect to prevent stale data
+    - Implement lock ordering for separate TX lock mode
+    - Read buffered data immediately after connection to prevent data loss
+    - Added sdkconfig.ci.tx_lock config
+
 ## [1.6.0](https://github.com/espressif/esp-protocols/commits/websocket-v1.6.0)
 
 ### Features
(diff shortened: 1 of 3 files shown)


components/esp_websocket_client/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+### Bug Fixes
+
+- Fix race conditions, memory leak, and data loss ([23ca97d5](https://github.com/espressif/esp-protocols/commit/23ca97d5))
+    - Add state check in abort_connection to prevent double-close
+    - Fix memory leak: free errormsg_buffer on disconnect
+    - Reset connection state on reconnect to prevent stale data
+    - Implement lock ordering for separate TX lock mode
+    - Read buffered data immediately after connection to prevent data loss
+    - Added sdkconfig.ci.tx_lock config
```
david-cermak, line 13: Nitpick: Not sure why the formatting is off here? Do we have an issue with the generator script?

david-cermak (approved):
LGTM, just minor formatting issue -- feel free to ignore, but would like to fix it somehow.
Also, I recall that each line did contain a reference to the original commit?

components/esp_websocket_client/CHANGELOG.md
gabsuren: @david-cermak thanks for pointing this out. 
Actually the changes were part of a single large commit https://github.com/espressif/esp-protocols/commit/23ca97d5eccff866e89cb880df3e8366bf7d12b7 (I probably should have split it into separate commits for each change). 
As a result, the generated changelog looks like this:
```
1.6.1
    Bug Fixes
    - Fix race conditions, memory leaks, and data loss (23ca97d5)
        - Add state check in abort_connection to prevent double-close
        - Fix memory leak: free errormsg_buffer on disconnect
        - Reset connection state on reconnect to prevent stale data
        - Implement lock ordering for separate TX lock mode
        - Read buffered data immediately after connection to prevent data loss
        - Add sdkconfig.ci.tx_lock config

```

Another option to keep them in one line separated  by `-` and appending the commit SHA at the end. 
What do you think would be the preferred approach here?

PR #980, opened 2025-12-29
esp_websocket_client: Add Kconfig options for PSRAM allocation (IDFGH-17011)

shardt68:
## Description

This PR introduces two new Kconfig options to the `esp_websocket_client` component to allow memory allocation in external RAM (PSRAM):

1. **`ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM`**: Enables allocation of the WebSocket task stack in PSRAM using `xTaskCreateStaticPinnedToCore`.
2. **`ESP_WS_CLIENT_ALLOC_IN_EXT_RAM`**: Enables allocation of the `esp_websocket_client` structure and its internal configuration storage in PSRAM.

### Motivation
In applications requiring multiple concurrent secure connections (e.g., TLS/HTTPS), internal RAM is a critical bottleneck. Each TLS session typically requires 16-20KB of internal RAM for buffers. By allowing the WebSocket client's task stack and internal structures to be moved to PSRAM, significant internal memory is freed for these critical network operations, improving overall system stability and preventing `ESP_ERR_NO_MEM` failures in memory-constrained scenarios.

### Implementation Details
- Uses `heap_caps_calloc` with `MALLOC_CAP_SPIRAM` for the client structure and config.
- Implements `xTaskCreateStaticPinnedToCore` for the task stack when PSRAM is selected.
- **Safety**: Includes a fallback mechanism that reverts to internal RAM allocation if PSRAM allocation fails at runtime.
- **Compatibility**: The default behavior remains unchanged (internal RAM allocation).

## Related
- Related to general ESP-IDF best practices for PSRAM-enabled devices (ESP32-S3, ESP32-WROVER) to mitigate internal RAM fragmentation and exhaustion during heavy network I/O.

## Testing
- **Hardware**: Tested on ESP32-S3 (8MB Octal PSRAM).
- **Scenario**: Verified stable WebSocket connection while simultaneously performing multiple concurrent HTTPS/TLS requests to external APIs.
- **Memory Verification**: Confirmed via `heap_caps_get_free_size` that internal RAM usage decreased by the expected amount (~5KB for structure + stack size).
- **Stability**: Verified correct task startup and proper resource cleanup during client destruction.
- **Fallback Test**: Verified that the client still initializes correctly on a device with PSRAM disabled/unavailable by falling back to internal RAM.

---

> [!NOTE]
> **Medium Risk**
> Touches core client lifecycle (init, start, task self-delete, destroy) and FreeRTOS external-stack rules; misconfiguration or PSRAM pressure can fail init/start with no fallback, though defaults preserve prior behavior.
> 
> **Overview**
> Adds two **optional** Kconfig switches (default off) so `esp_websocket_client` can move selected allocations off internal RAM when SPIRAM and related FreeRTOS caps are enabled.
> 
> **`ESP_WS_CLIENT_ALLOC_IN_EXT_RAM`** — `esp_websocket_client_init()` allocates the client object and `websocket_config_storage_t` with `heap_caps_calloc(..., MALLOC_CAP_SPIRAM)`. Teardown uses `heap_caps_free` via `ESP_WS_CLIENT_OBJ_FREE`. Failed PSRAM alloc returns `NULL`; Kconfig documents **no** internal-RAM fallback.
> 
> **`ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM`** — `esp_websocket_client_start()` creates the worker with `xTaskCreatePinnedToCoreWithCaps(..., MALLOC_CAP_SPIRAM)`; the task exits with `vTaskDeleteWithCaps(NULL)` instead of `vTaskDelete`. Start fails with `ESP_FAIL` if task creation fails.
> 
> Also tightens task-start error handling (`res != pdPASS || task_handle == NULL`) and pulls in `esp_heap_caps.h` / `freertos/idf_additions.h` for the caps APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64b47ec979f17867709ca18771404287f99424b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -6,6 +6,7 @@
 
 #include <stdio.h>
 
+#include "esp_heap_caps.h"
 #include "esp_websocket_client.h"
 #include "esp_transport.h"
 #include "esp_transport_tcp.h"
@@ -13,6 +14,7 @@
 /* using uri parser */
 #include "http_parser.h"
 #include "freertos/task.h"
+#include "freertos/idf_additions.h"
 #include "freertos/semphr.h"
 #include "freertos/queue.h"
 #include "freertos/event_groups.h"
@@ -24,6 +26,12 @@
 #include <arpa/inet.h>
 
 static const char *TAG = "websocket_client";
+
+#if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
+#define ESP_WS_CLIENT_OBJ_FREE(ptr) heap_caps_free(ptr)
+#else
+#define ESP_WS_CLIENT_OBJ_FREE(ptr) free(ptr)
+#endif
 
 #define WEBSOCKET_TCP_DEFAULT_PORT      (80)
 #define WEBSOCKET_SSL_DEFAULT_PORT      (443)
@@ -488,7 +496,7 @@ static esp_err_t esp_websocket_client_destroy_config(esp_websocket_client_handle
     free(cfg->user_agent);
     free(cfg->headers);
     memset(cfg, 0, sizeof(websocket_config_storage_t));
-    free(client->config);
+    ESP_WS_CLIENT_OBJ_FREE(client->config);
     client->config = NULL;
     return ESP_OK;
 }
@@ -535,7 +543,7 @@ static void destroy_and_free_resources(esp_websocket_client_handle_t client)
         vEventGroupDelete(client->status_bits);
         client->status_bits = NULL;
     }
-    free(client);
+    ESP_WS_CLIENT_OBJ_FREE(client);
     client = NULL;
 }
 
@@ -554,7 +562,6 @@ static esp_err_t stop_wait_task(esp_websocket_client_handle_t client)
     client->state = WEBSOCKET_STATE_UNKNOW;
     return ESP_OK;
 }
-
 #if WS_TRANSPORT_HEADER_CALLBACK_SUPPORT
 static void websocket_header_hook(void * client, const char * line, int line_len)
 {
@@ -785,7 +792,11 @@ unlock_and_return:
 
 esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_client_config_t *config)
 {
+#if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
+    esp_websocket_client_handle_t client = heap_caps_calloc(1, sizeof(struct esp_websocket_client), MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT);
+#else
     esp_websocket_client_handle_t client = calloc(1, sizeof(struct esp_websocket_client));
+#endif
     ESP_WS_CLIENT_MEM_CHECK(TAG, client, return NULL);
 
     esp_event_loop_args_t event_args = {
@@ -795,7 +806,7 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
 
     if (esp_event_loop_create(&event_args, &client->event_handle) != ESP_OK) {
         ESP_LOGE(TAG, "Error create event handler for websocket client");
-        free(client);
+        ESP_WS_CLIENT_OBJ_FREE(client);
         return NULL;
     }
 
@@ -820,7 +831,11 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
     ESP_WS_CLIENT_MEM_CHECK(TAG, client->tx_lock, goto _websocket_init_fail);
 #endif
 
+#if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
+    client->config = heap_caps_calloc(1, sizeof(websocket_config_storage_t), MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT);
+#else
     client->config = calloc(1, sizeof(websocket_config_storage_t));
+#endif
     ESP_WS_CLIENT_MEM_CHECK(TAG, client->config, goto _websocket_init_fail);
 
     if (config->transport == WEBSOCKET_TRANSPORT_OVER_TCP) {
@@ -1450,7 +1465,13 @@ static void esp_websocket_client_task(void *pv)
     } else {
         xEventGroupSetBits(client->status_bits, STOPPED_BIT);
     }
+#if CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM
+    /* Task stack/TCB were allocated with memory caps via xTaskCreatePinnedToCoreWithCaps(),
+     * so it must be deleted with vTaskDeleteWithCaps(), including on self-deletion. */
+    vTaskDeleteWithCaps(NULL);
+#else
     vTaskDelete(NULL);
+#endif
 }
 
 esp_err_t esp_websocket_client_start(esp_websocket_client_handle_t client)
@@ -1458,6 +1479,7 @@ esp_err_t esp_websocket_client_start(esp_websocket_client_handle_t client)
     if (client == NULL) {
         return ESP_ERR_INVALID_ARG;
     }
+
     if (client->state >= WEBSOCKET_STATE_INIT) {
         ESP_LOGE(TAG, "The client has started");
         return ESP_FAIL;
@@ -1472,8 +1494,19 @@ esp_err_t esp_websocket_client_start(esp_websocket_client_handle_t client)
     }
 
     xEventGroupClearBits(client->status_bits, STOPPED_BIT | CLOSE_FRAME_SENT_BIT | REQUESTED_STOP_BIT | WAKEUP_BIT);
-    if (xTaskCreatePinnedToCore(esp_websocket_client_task, client->config->task_name ? client->config->task_name : "websocket_task",
-                                client->config->task_stack, client, client->config->task_prio, &client->task_handle, client->config->task_core_id) != pdTRUE) {
+
+    BaseType_t res;
+#if CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM
+    res = xTaskCreatePinnedToCoreWithCaps(esp_websocket_client_task, client->config->task_name ? client->config->task_name : "websocket_task",
+                                          client->config->task_stack, client, client->config->task_prio, &client->task_handle, client->config->task_core_id,
+                                          MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT);
+#else
+    res = xTaskCreatePinnedToCore(esp_websocket_client_task, client->config->task_name ? client->config->task_name : "websocket_task",
+                                  client->config->task_stack, client, client->config->task_prio, &client->task_handle, client->config->task_core_id);
+#endif
+
+    if (res != pdPASS || client->task_handle == NULL) {
+        client->task_handle = NULL;
         ESP_LOGE(TAG, "Error create websocket task");
         xEventGroupSetBits(client->status_bits, STOPPED_BIT);
         return ESP_FAIL;
--- a/components/esp_websocket_client/Kconfig
+++ b/components/esp_websocket_client/Kconfig
@@ -20,4 +20,32 @@ menu "ESP WebSocket client"
         default 2000
         help
             Timeout for acquiring the TX lock when using separate TX lock.
+
+    config ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
+        bool "Allocate esp_websocket_client structure and config storage in PSRAM"
+        default n
+        depends on SPIRAM
+        depends on (SPIRAM_USE_CAPS_ALLOC || SPIRAM_USE_MALLOC)
+        help
+            Enable this option to allocate the esp_websocket_client structure and its
+            internal configuration storage in external RAM (PSRAM). This is useful for
+            freeing up internal RAM in memory-constrained applications.
+            If PSRAM allocation fails, esp_websocket_client_init() returns NULL.
+            There is no fallback to internal RAM.
+
+    config ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM
+        bool "Allocate websocket task stack in PSRAM"
+        default n
+        depends on SPIRAM
+        depends on (SPIRAM_USE_CAPS_ALLOC || SPIRAM_USE_MALLOC)
+        depends on FREERTOS_SUPPORT_STATIC_ALLOCATION
+        depends on FREERTOS_TASK_CREATE_ALLOW_EXT_MEM
+        help
+            Enable this option to allocate the WebSocket task stack in external RAM (PSRAM)
+            using xTaskCreatePinnedToCoreWithCaps()/vTaskDeleteWithCaps(). Only the task stack
+            is placed in PSRAM; the TCB (Task Control Block) remains in internal RAM as
+            required by the API.
+            This option requires FreeRTOS to allow creating tasks with stacks in external memory.
+            If PSRAM stack allocation fails, esp_websocket_client_start() returns ESP_FAIL.
+            There is no fallback to internal RAM.
 endmenu


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -511,13 +529,30 @@ static esp_err_t stop_wait_task(esp_websocket_client_handle_t client)
...
+    xSemaphoreTakeRecursive(client->lock, portMAX_DELAY);
     client->run = false;
     xEventGroupSetBits(client->status_bits, REQUESTED_STOP_BIT);
+    xSemaphoreGiveRecursive(client->lock);
+
     xEventGroupWaitBits(client->status_bits, STOPPED_BIT, false, true, portMAX_DELAY);
+    
+    xSemaphoreTakeRecursive(client->lock, portMAX_DELAY);
+    if (client->task_handle) {
+        while (client->task_handle && eTaskGetState(client->task_handle) != eSuspended) {
```
gabsuren: Wait for task to suspend might block indefinitely if the task is blocked elsewhere.  If the client is stuck in a network call or blocked on a mutex, stop() will hang the caller forever

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1377,35 +1483,123 @@ static void esp_websocket_client_task(void *pv)
...
+
+        if (!already_destroying) {
+            if (xTaskCreate(esp_websocket_client_destroy_task, "ws_destroy", 4096, client, client->config->task_prio, NULL) != pdPASS) {
+                ESP_LOGE(TAG, "Failed to create destroy task, memory will leak");
+                xEventGroupClearBits(client->status_bits, DESTRUCTION_IN_PROGRESS_BIT);
+            }
+        }
     }
-    vTaskDelete(NULL);
+    vTaskSuspend(NULL);
```
gabsuren: If the task crashes, blocks, or enters an infinite loop before it reaches line 1502, it will never become suspended right?

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1377,35 +1483,123 @@ static void esp_websocket_client_task(void *pv)
...
+        xSemaphoreTakeRecursive(client->lock, portMAX_DELAY);
+        if (xEventGroupGetBits(client->status_bits) & DESTRUCTION_IN_PROGRESS_BIT) {
+            already_destroying = true;
+        } else {
+            xEventGroupSetBits(client->status_bits, DESTRUCTION_IN_PROGRESS_BIT);
+        }
+        xSemaphoreGiveRecursive(client->lock);
+
+        if (!already_destroying) {
+            if (xTaskCreate(esp_websocket_client_destroy_task, "ws_destroy", 4096, client, client->config->task_prio, NULL) != pdPASS) {
```
gabsuren: If the system is Out Of Memory (OOM), the helper task is not created. The client task suspends itself, but no one is left to free its allocated memory.

gabsuren (changes requested):
Left some notes. Please take a look. 
The task cleanup logic is duplicated across 4 different locations. This increases maintenance burden and the risk that future changes will introduce bugs if not applied consistently to all 4 spots.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -156,6 +156,8 @@ struct esp_websocket_client {
     int                         payload_offset;
     esp_transport_keep_alive_t  keep_alive_cfg;
     struct ifreq                *if_name;
+    StackType_t                 *task_stack_buffer;
+    StaticTask_t                *task_buffer;
```
gabsuren: I think it's also should be configurable
#if CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM  // <--- SHOULD BE ADDED
    StackType_t                 *task_stack_buffer;
    StaticTask_t                *task_buffer;
#endif

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -747,7 +757,11 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
 
 esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_client_config_t *config)
 {
+#if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
+    esp_websocket_client_handle_t client = heap_caps_calloc(1, sizeof(struct esp_websocket_client), MALLOC_CAP_SPIRAM);
```
gabsuren: Best practice is to use `MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT` , while ESP32 PSRAM is byte-addressable, explicit flags are safer for portability across different ESP chips where external RAM behavior might vary.

gabsuren: If heap_caps_calloc returns NULL, the function exits with an error there is no falling back to internal RAM ? Is this implementation strict only for PSRAM. if so you should remove the fallback logic from the task stack allocation (start function)

gabsuren (changes requested):

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1432,8 +1454,47 @@ esp_err_t esp_websocket_client_start(esp_websocket_client_handle_t client)
...
+            );
+            if (client->task_handle == NULL) {
+                res = pdFAIL;
+            }
+        } else {
+            ESP_LOGE(TAG, "Failed to allocate PSRAM stack for WebSocket task");
+            res = pdFAIL;
+        }
+    } else {
+        res = xTaskCreatePinnedToCore(esp_websocket_client_task, client->config->task_name ? client->config->task_name : "websocket_task",
```
gabsuren: If the PR's intent is strict fail-fast (which I think it should be) then the else { xTaskCreatePinnedToCore } branch contradicts that intent. It silently allocates the task stack in internal RAM whenever task_stack <= 0, even though the user opted into PSRAM allocation.

gabsuren (changes requested):
Hi @shardt68, thanks for iterating on this, the strict-allocation direction is good. A few things to address before merge


1. Commits

Please rebase onto current master and squash the 7 commits into a single conventional commit, for example:

feat(websocket): add Kconfig options for PSRAM allocation

2. Strict allocation is not fully strict yet

In esp_websocket_client_start(), the else { xTaskCreatePinnedToCore(...) } branch silently falls back to internal RAM when task_stack <= 0.

Since the user explicitly enabled PSRAM task-stack allocation, I think the behavior should be:

PSRAM stack allocation succeeds, or start fails with a clear error.

Please remove the fallback branch so the option consistently means “allocate the task stack in PSRAM, or fail”.

3. Kconfig documentation / dependencies

Please document and enforce the expected dependencies:

- depends on SPIRAM
- depends on FREERTOS_SUPPORT_STATIC_ALLOCATION for the task-stack option
- mention that only the task stack moves to PSRAM; the TCB stays in internal RAM
- mention the allocation-failure behavior: return NULL / ESP_FAIL, no fallback to internal RAM


Looks good otherwise, once these are in, happy to take another pass. Thanks!

components/esp_websocket_client/esp_websocket_client.c
shardt68: @gabsuren made the requested changes, hope it fits, Regards!

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -777,7 +790,11 @@ static int esp_websocket_client_send_with_exact_opcode(esp_websocket_client_hand
 
 esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_client_config_t *config)
 {
+    #if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
```
gabsuren: Minor spacing inconsistency

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -812,7 +829,11 @@ esp_websocket_client_handle_t esp_websocket_client_init(const esp_websocket_clie
     ESP_WS_CLIENT_MEM_CHECK(TAG, client->tx_lock, goto _websocket_init_fail);
 #endif
 
+    #if CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM
```
gabsuren: same here (spacings)

components/esp_websocket_client/esp_websocket_client.c
gabsuren, line 13: Please add explicit `#include "esp_heap_caps.h"` because this file directly uses heap_caps_calloc, heap_caps_free

gabsuren:
@shardt68 thank you for the updates. Please fix the pre-commit check and address the minor notes I left. Other than that, I think the PR is mostly ready for merging.

There is one remaining major issue with `CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM` enabled: the direct dangerous case is `destroy_on_exit`. When `selected_for_destroying` is true, the WebSocket task can call [destroy_and_free_resources](https://github.com/shardt68/esp-protocols/blob/2c3347f39b5532a6ddb85cbddc4c1efd69aa0382/components/esp_websocket_client/esp_websocket_client.c#L1441)() from inside itself, which frees its own static stack/TCB before calling [vTaskDelete(NULL).](https://github.com/shardt68/esp-protocols/blob/2c3347f39b5532a6ddb85cbddc4c1efd69aa0382/components/esp_websocket_client/esp_websocket_client.c#L1445) 

This was not an issue before because the task stack/TCB were owned by FreeRTOS. With the new external-stack path, they are client-owned buffers, so freeing them from the running task can lead to use-after-free or memory corruption.

I think this should be fixed before enabling `CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM`, or the limitation should at least be `documented` and `destroy_on_exit` should be prevented/unsupported in this mode untill the fix.

shardt68:
I addressed the remaining Bugbot findings.

PSRAM allocations freed with free
Fixed by using matching deallocation for client/config objects:
heap_caps_free when CONFIG_ESP_WS_CLIENT_ALLOC_IN_EXT_RAM is enabled
free otherwise
Applied consistently in init error path and teardown paths.
Start failure leaks PSRAM task buffers
Updated start() error path in PSRAM stack mode to free task_stack_buffer/task_buffer immediately and clear pointers, instead of keeping them until destroy.
Additionally, the previous high-severity race around PSRAM stack free timing was already handled in commit c6a31b2 (yield before freeing stack/TCB buffers).

Please take another look.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -527,7 +538,24 @@ static void destroy_and_free_resources(esp_websocket_client_handle_t client)
...
     }
-    free(client);
+#if CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM
+    /* Yield one scheduler tick before freeing the PSRAM stack/TCB buffers.
+     * The WebSocket task sets STOPPED_BIT and then calls vTaskDelete(NULL).
+     * On multi-core targets the destroyer can reach this point before the
+     * WebSocket task has finished executing vTaskDelete(NULL), causing a
+     * use-after-free of the PSRAM stack.  One tick is sufficient for the
+     * scheduler to remove the task from all ready lists. */
+    vTaskDelay(1);
```
gabsuren: vTaskDelay(1) is not a reliable barrier for the delete-on-another-core case

shardt68:
Addressed the remaining unresolved Kconfig point:\n\n- Added missing dependency  to  in .\n- Updated help text to explicitly mention that FreeRTOS must allow external-memory task stacks.\n\nCommit: 6a178672eb184d8bba51cd94711ab1fb116e497e\n\nPlease re-check this thread; it should be covered now.

shardt68:
Addressed the remaining unresolved Kconfig point:

- Added missing dependency `FREERTOS_TASK_CREATE_ALLOW_EXT_MEM` to `ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM` in `components/esp_websocket_client/Kconfig`.
- Updated help text to explicitly mention that FreeRTOS must allow external-memory task stacks.

Commit: 6a178672eb184d8bba51cd94711ab1fb116e497e

Please re-check this thread; it should be covered now.

shardt68:
All review threads have been resolved:

- ✅ Fixed Kconfig dependency for `FREERTOS_TASK_CREATE_ALLOW_EXT_MEM` (commit e68e3615)
- ✅ All previous feedback addressed and integrated
- ✅ PR focused strictly on PSRAM allocation options as requested

Ready for review and merge. Thanks!

gabsuren:
@shardt68 Thanks for all the work on this PR. I also missed this ESP-IDF API earlier, so apologies for bringing it up this late.

Have you considered using `xTaskCreatePinnedToCoreWithCaps(..., MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT)` with `vTaskDeleteWithCaps(NULL)` instead of managing the stack and TCB buffers manually?

ESP-IDF would then own the buffers, with the stack in PSRAM and the TCB in internal RAM. This could keep the task lifecycle close to the normal internal-RAM path and avoid most of the extra cleanup and deletion complexity.

If there is no restriction on requiring ESP-IDF 5.3 or newer, this may be a simpler approach. `ESP_WS_CLIENT_ALLOC_IN_EXT_RAM` would remain unaffected.

Regards,
Suren

shardt68:
## Update: adopted `xTaskCreatePinnedToCoreWithCaps()` / `vTaskDeleteWithCaps()`

Thanks @gabsuren for the review. I've reworked the PSRAM task-stack handling to use the ESP-IDF FreeRTOS caps APIs instead of manual buffer management. The branch was squashed and force-pushed.

### Changes
- **Refactored task creation/deletion** to `xTaskCreatePinnedToCoreWithCaps(..., MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT)` / `vTaskDeleteWithCaps()` (via `freertos/idf_additions.h`), replacing the manual `heap_caps_calloc()` of the stack + `StaticTask_t` TCB and `xTaskCreateStaticPinnedToCore()`.
- **Removed** the manual bookkeeping fields (`task_stack_buffer`, `task_buffer`) from `struct esp_websocket_client` and the corresponding `heap_caps_free()` cleanup in `destroy_and_free_resources()`.
- **Removed the `vTaskSuspend(NULL)` workaround** — self-deletion via `vTaskDeleteWithCaps(NULL)` safely frees the caller's own PSRAM stack (it spawns a temporary internal cleanup task), so the task exit path is now unified for both the PSRAM and the default case.
- **`esp_websocket_client_destroy_on_exit()` is now fully supported** with the PSRAM option enabled — the previous `ESP_ERR_NOT_SUPPORTED` restriction was removed.
- Task-creation failure handling simplified to a single generic path.
- Updated the Kconfig help text of `ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM` to describe the new API and to note that only the task **stack** is placed in PSRAM while the **TCB stays in internal RAM** (as required by the API).

### Findings & decisions
- **`ESP_WS_CLIENT_ALLOC_IN_EXT_RAM`** (client-object allocation in PSRAM) is a separate feature and is **unaffected** by this change.
- **Version floor — kept `idf: version: \">=5.0\"`, no `ESP_IDF_VERSION_VAL` guard added.** Rationale:
  - The new APIs (`xTaskCreatePinnedToCoreWithCaps` / `vTaskDeleteWithCaps`) have existed since **IDF v5.1**.
  - The Kconfig option `depends on FREERTOS_TASK_CREATE_ALLOW_EXT_MEM`, which was only introduced in **IDF v5.4+**. On older IDF that symbol is undefined, so the option evaluates to `n` and every `#if CONFIG_ESP_WS_CLIENT_TASK_STACK_IN_EXT_RAM` block compiles out.
  - Net effect: the component still builds on `>=5.0` with the feature disabled, and wherever the feature *can* be enabled (>=5.4) the APIs are guaranteed to be present. The existing `depends on` therefore already provides exact version gating, so no explicit version guard or manifest bump is needed.

gabsuren:
@shardt68 thank you for your contribution! The changes are now on master.

PR #983, opened 2026-01-05
fix(mbedtls_cxx): Add support for mbedtls psa api

david-cermak:
> [!NOTE]
> Adds mbedTLS v4 (PSA) support with conditional compilation while preserving v3 behavior, plus DTLS timer and cookie updates.
> 
> - Add version detection in `mbedtls_wrap.hpp` and split code paths for v4 (PSA) vs v3 (entropy/CTR_DRBG + timing)
> - Initialize PSA via `psa_crypto_init()` in `Tls::init()` for v4; retain RNG seeding/config for v3 and free on deinit/destructor
> - Implement DTLS timer callbacks using `esp_timer` for v4; keep `mbedtls_timing_*` for v3
> - Update DTLS cookie setup to PSA-compatible form on v4; unchanged RNG-backed setup on v3
> - Require `esp_timer` in `CMakeLists.txt`
> - README: document v3/v4 support and PSA crypto lifecycle notes
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 884d8fe6f54ba54808e62e4c368e2659aa9b9965. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/mbedtls_cxx/mbedtls_wrap.cpp
+++ b/components/mbedtls_cxx/mbedtls_wrap.cpp
@@ -1,23 +1,80 @@
 /*
- * SPDX-FileCopyrightText: 2023-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2023-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-#include <mbedtls/timing.h>
-#include "mbedtls/ctr_drbg.h"
 #include "mbedtls/ssl.h"
 #include "mbedtls_wrap.hpp"
+
+#if MBEDTLS_CXX_MBEDTLS_MAJOR >= 4
+#include "esp_timer.h"
+#include "psa/crypto.h"
+
+namespace {
+
+void timer_set_delay(void *data, uint32_t int_ms, uint32_t fin_ms)
+{
+    auto *ctx = static_cast<idf::mbedtls_cxx::dtls_timer_context *>(data);
+    if (fin_ms == 0) {
+        ctx->int_ms = 0;
+        ctx->fin_ms = 0;
+        ctx->start_us = 0;
+        return;
+    }
+    ctx->int_ms = int_ms;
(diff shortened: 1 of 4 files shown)


components/mbedtls_cxx/mbedtls_wrap.cpp
```diff
@@ -1,31 +1,87 @@
...
 #include "mbedtls/ssl.h"
 #include "mbedtls_wrap.hpp"
 
 using namespace idf::mbedtls_cxx;
 
+#if MBEDTLS_CXX_MBEDTLS_MAJOR >= 4
+#include "esp_timer.h"
+#include "psa/crypto.h"
+
+static void mbedtls_cxx_timer_set_delay(void *data, uint32_t int_ms, uint32_t fin_ms)
```
euripedesrocha: Nitpick: This would be cleaner if added before the `using namespace` and under an anonymous namespace.

components/mbedtls_cxx/mbedtls_wrap.cpp
```diff
@@ -76,6 +136,13 @@ bool Tls::deinit()
     ::mbedtls_pk_free(&pk_key_);
     ::mbedtls_x509_crt_free(&public_cert_);
     ::mbedtls_x509_crt_free(&ca_cert_);
+#if MBEDTLS_CXX_MBEDTLS_MAJOR < 4
```
euripedesrocha, line 143: No need to clean up in the new API?

david-cermak: Very good point!
but there's no cleanup for these two particular function.

there is a general PSA-crypto cleanup, but that's common for the entire application, so it's not safe to call it here (neither with ref-counting)
I've added a note to the readme with the explanation

components/mbedtls_cxx/mbedtls_wrap.cpp
david-cermak: fixed, thanks!

PR #985, opened 2026-01-08
fix(asio): Add support for mbedtls-v4 API (IDF v6.0+ support)

david-cermak:


components/asio/idf_component.yml
```diff
@@ -9,3 +9,7 @@ dependencies:
     version: ">=5.0"
   espressif/sock_utils:
     version: "^0.1"
+  david-cermak/mbedtls_v3_shim:
```
euripedesrocha: Should we provide it in espressif namespace?

david-cermak: I'm still unsure about that...

I'd keep it in my personal namespace and rework these fixes according to mbedtl-cxx (with #ifdefs as all libs do it this way)

david-cermak: removed and migrated the sources

PR #997, opened 2026-01-27
fix(websocket): Fix the bug related to multi-threaded access to trans… (IDFGH-17155)

udoudou:
The `close` operation on the `client->transport` resource is performed under the protection of `client->lock`. `esp_websocket_client_send_with_exact_opcode` checks the state without holding the lock when determining whether the transport can be accessed. This can lead to some unexpected behaviors:

1. If `tx_lock` is not enabled, calling `esp_transport_ws_send_raw` will be abnormally blocked if the transport is already closed.

2. If `tx_lock` is enabled, `esp_transport_close` and `esp_transport_ws_send_raw` may be called concurrently. Since `close` involves releasing some resources, this may lead to unpredictable behavior.

## Description
To fix this issue, I adjusted the implementation of resource mutual exclusion protection.
1、tx_lock has been removed
The original purpose of introducing `tx_lock` was to allow for parallel processing of receiving and sending, improving transmission efficiency. However, this doesn't necessarily require introducing a new `tx_lock`. This is because the receiving direction is accessed only by the `esp_websocket_client_task`. The sending direction may be accessed concurrently by several user tasks and the `esp_websocket_client_task`. Therefore, the `esp_websocket_client_task` does not need lock protection when receiving. Lock protection is only applied when the `esp_websocket_client_task` needs to send or when a user task sends, to check if the transport is available.

2、Change `esp_websocket_client_abort_connection` to asynchronous close transport.
Modify `esp_websocket_client_abort_connection` to only modify the state, preventing user tasks from directly executing close transport and conflicting with `esp_websocket_client_task`'s receive transport.

The APIs involved in resource access contention mainly include esp_websocket_client_send_bin, esp_websocket_client_send_bin_partial, esp_websocket_client_send_text, esp_websocket_client_send_text_partial, esp_websocket_client_send_cont_msg, esp_websocket_client_send_fin, esp_websocket_client_send_with_opcode, esp_websocket_client_close, and esp_websocket_client_close_with_code. Ultimately, all of these are involved in esp_websocket_client_send_with_exact_opcode. The main issue is contention involving transport sending, errormsg_buffer, and state when the state is WEBSOCKET_STATE_CONNECTED. After the modification, esp_websocket_client_send_with_exact_opcode accesses and modifies these resources under lock protection. In esp_websocket_client_task, when the state is WEBSOCKET_STATE_CONNECTED, any transport sending, changes to a connectionless state, or access to errormsg_buffer are all performed under lock protection. The close of the transport is uniformly performed by esp_websocket_client_task in a non-WEBSOCKET_STATE_CONNECTED state, ensuring that user task sending will not occur concurrently with the close of the transport, and esp_websocket_client_task does not require lock protection when receiving.

## Testing

---

## Checklist

---

> [!NOTE]
> Refactors websocket client concurrency and connection lifecycle handling.
> 
> - Removes `tx_lock` Kconfig/options and all associated separate TX locking; uses a single `client->lock` to guard all transport sends/state/error buffer access
> - Introduces `WEBSOCKET_STATE_WAIT_ABORT_CONNECT` and changes `esp_websocket_client_abort_connection()` to be asynchronous (sets state, dispatches disconnect); actual `esp_transport_close()` now performed by the client task
> - Ensures PING/PONG/CLOSE and send paths acquire `client->lock` and re-check `state` before accessing transport; adds locked aborts on read/write failures
> - Adjusts main task state machine: splits abort handling, reconnect/wait logic, and server-initiated close flow; removes previous WAIT_TIMEOUT handling tied to immediate close
> - Cleans up Kconfig (drops tx lock options) and updates SPDX year
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3bbe73ab983263237b8966542f65bc9fdde4801d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -38,10 +38,6 @@ static const char *TAG = "websocket_client";
 #define WEBSOCKET_KEEP_ALIVE_IDLE       (5)
 #define WEBSOCKET_KEEP_ALIVE_INTERVAL   (5)
 #define WEBSOCKET_KEEP_ALIVE_COUNT      (3)
-
-#ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
-#define WEBSOCKET_TX_LOCK_TIMEOUT_MS    (CONFIG_ESP_WS_CLIENT_TX_LOCK_TIMEOUT_MS)
-#endif
 
 #define ESP_WS_CLIENT_MEM_CHECK(TAG, a, action) if (!(a)) {                                         \
         ESP_LOGE(TAG,"%s(%d): %s", __FUNCTION__, __LINE__, "Memory exhausted");                     \
@@ -121,6 +117,7 @@ typedef enum {
     WEBSOCKET_STATE_INIT,
     WEBSOCKET_STATE_CONNECTED,
     WEBSOCKET_STATE_WAIT_TIMEOUT,
+    WEBSOCKET_STATE_WAIT_ABORT_CONNECT,
     WEBSOCKET_STATE_CLOSING,
 } websocket_client_state_t;
 
@@ -142,9 +139,6 @@ struct esp_websocket_client {
     bool                        selected_for_destroying;
     EventGroupHandle_t          status_bits;
     SemaphoreHandle_t           lock;
-#ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
-    SemaphoreHandle_t           tx_lock;
-#endif
     size_t                      errormsg_size;
     char                        *errormsg_buffer;
     char                        *rx_buffer;
@@ -258,22 +252,12 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
     ESP_WS_CLIENT_STATE_CHECK(TAG, client, return ESP_FAIL);
 
 
-    if (client->state == WEBSOCKET_STATE_CLOSING || client->state == WEBSOCKET_STATE_UNKNOW ||
-            client->state == WEBSOCKET_STATE_WAIT_TIMEOUT) {
+    if (client->state != WEBSOCKET_STATE_CONNECTED && client->state != WEBSOCKET_STATE_INIT) {
         ESP_LOGW(TAG, "Connection already closing/closed, skipping abort");
         goto cleanup;
     }
 
-    esp_transport_close(client->transport);
-
-    if (!client->config->auto_reconnect) {
-        client->run = false;
-        client->state = WEBSOCKET_STATE_UNKNOW;
-    } else {
-        client->reconnect_tick_ms = _tick_get_ms();
-        ESP_LOGI(TAG, "Reconnect after %d ms", client->wait_timeout_ms);
-        client->state = WEBSOCKET_STATE_WAIT_TIMEOUT;
-    }
+    client->state = WEBSOCKET_STATE_WAIT_ABORT_CONNECT;
     client->error_handle.error_type = error_type;
     esp_websocket_client_dispatch_event(client, WEBSOCKET_EVENT_DISCONNECTED, NULL, 0);
 
@@ -489,9 +473,6 @@ static void destroy_and_free_resources(esp_websocket_client_handle_t client)
         client->transport = NULL;
     }
     vSemaphoreDelete(client->lock);
-#ifdef CONFIG_ESP_WS_CLIENT_SEPARATE_TX_LOCK
-    vSemaphoreDelete(client->tx_lock);
-#endif
     free(client->tx_buffer);
     free(client->rx_buffer);
     free(client->errormsg_buffer);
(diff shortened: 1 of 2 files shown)


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1332,43 +1216,78 @@ static void esp_websocket_client_task(void *pv)
                 esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
                 xSemaphoreGiveRecursive(client->lock);
             } else if (read_select > 0) {
-                xSemaphoreTakeRecursive(client->lock, lock_timeout);
                 if (esp_websocket_client_recv(client) == ESP_FAIL) {
                     ESP_LOGE(TAG, "Error receive data");
-                    esp_websocket_client_abort_connection(client, WEBSOCKET_ERROR_TYPE_TCP_TRANSPORT);
                 }
-                xSemaphoreGiveRecursive(client->lock);
```
gabsuren, line 1340: Won't removing the lock be unsafe for wss:// (SSL/TLS) connections?
The underlying SSL library maintains internal state that cannot be safely accessed by mbedtls_ssl_read and mbedtls_ssl_write simultaneously without external synchronization.
By releasing `client->lock` during esp_websocket_client_recv(), we allow a race condition here.

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -258,22 +252,12 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
...
-
-    if (!client->config->auto_reconnect) {
-        client->run = false;
-        client->state = WEBSOCKET_STATE_UNKNOW;
-    } else {
-        client->reconnect_tick_ms = _tick_get_ms();
-        ESP_LOGI(TAG, "Reconnect after %d ms", client->wait_timeout_ms);
-        client->state = WEBSOCKET_STATE_WAIT_TIMEOUT;
-    }
+    client->state = WEBSOCKET_STATE_WAIT_ABORT_CONNECT;
```
gabsuren, line 260: Could you please clarify the behavior regarding delayed cleanup?
Since `esp_websocket_client_abort_connection` now sets the state to `WEBSOCKET_STATE_WAIT_ABORT_CONNECT` instead of closing the transport immediately, it seems the actual closure is deferred to the main task loop, right?

If the main task is currently blocked inside esp_transport_read() (which runs without the lock), it will not process this state change until the read returns or times out (up to network_timeout_ms, default 10s).

This means related resources will remain open for that duration even after the user explicitly calls esp_websocket_client_stop() or _destroy() . This could prevent immediate resource reclamation or fast reconnection attempts if the user expects synchronous cleanup.

gabsuren (changes requested):

components/esp_websocket_client/esp_websocket_client.c
udoudou: The read interface is only called within the esp_websocket_client_task. Furthermore, the shutdown of the transport is now performed within the esp_websocket_client_task. Therefore, no additional lock protection is required.

components/esp_websocket_client/esp_websocket_client.c
udoudou: In fact, the original implementation also couldn't immediately reclaim or quickly reconnect because `esp_transport_read` was protected by a lock. Even if a user thread called the interface to directly close the transport, it still needed to wait for `esp_transport_read` to return and release the lock before it could acquire the lock and perform the close operation. Therefore, I think this has little impact.

Furthermore, this design reduces the optimization to requiring only one lock. Management and maintenance should be easier, without worrying about multiple locks deadlocking each other.

udoudou: To optimize the closing speed, my current idea is to obtain the socket file descriptor (fd) corresponding to the transport and close that fd directly. This would allow write/read operations from other tasks to return immediately. The socket fd layer itself is designed to allow this, but after abstraction by esp_transport, the race condition risk still needs careful evaluation. Currently, I suspect a potential risk: after a user thread closes socket fd A, a third task might create a new socket fd and return the same fd A. The esp_websocket_client_task would then directly or indirectly perform operations on the cached fd A again. This optimization is not currently considered for this aspect.

components/esp_websocket_client/esp_websocket_client.c
gabsuren: I agree that `recv` is only called by the main task, so we don't have multiple threads trying to read at the same time.

However, my concern is about concurrent **Read** and **Write**.

With this change, the main task calls `esp_transport_read()` without holding the lock. While it is blocked there waiting for data, a user thread can call `esp_websocket_client_send()`, acquire the lock (since the main task doesn't have it), and call `esp_transport_write()`.

This means `esp_transport_read` and `esp_transport_write` will execute simultaneously on the same transport handle.

_For TCP (ws://):_ This is likely fine as the kernel handles concurrent socket access.
_For SSL (wss://):_ This is generally unsafe. The underlying mbedtls context is updated by both read and write operations. If esp_transport_ssl doesn't have its own internal mutex (which I believe it doesn't), accessing the same SSL context from two threads at once will cause undefined behavior or crashes.

udoudou: This point has not yet been considered; I will check this area and make further improvements and modifications.

components/esp_websocket_client/esp_websocket_client.c
gabsuren: I understand that the original implementation also has latency because `stop()` had to wait for the lock.

However, the difference is that `stop()` was synchronous. When it returned, the user was guaranteed that the resources (FD) were released.

With the new asynchronous cleanup (deferred to the task loop), stop() returns immediately, but the socket remains open. If the user code does:

```
esp_websocket_client_stop(client);

// ... destroy ...

`esp_websocket_client_init(&new_config);` // RESTART
```

The `new init/start` might fail because the old socket is still held by the blocked reader task.

Regarding the FD close optimization: I agree that closing the FD directly from another thread is very risky due to the reuse race condition you mentioned. It is safer to avoid that unless esp_transport provides a safe cancellation API.

PR #1008, opened 2026-02-11
[mdns]: Minor corrections to mdns socket layer

david-cermak:
# mDNS socket layer cleanup (mdns_networking_socket.c)

This note documents follow-up fixes applied to the BSD-socket based mDNS networking backend (`components/mdns/mdns_networking_socket.c`).

## 1. Deterministic RX task shutdown (remove fixed delay)

**Problem:** `mdns_priv_if_deinit()` previously stopped the RX task with a fixed `vTaskDelay(500ms)`, which is not a deterministic synchronization mechanism (task may still be running; rapid re-init can race).

**Correction:**
- Track the RX task using `s_sock_recv_task_handle`.
- Add a binary semaphore `s_sock_recv_task_exit_sem`.
- On last-interface deinit:
  - set `s_run_sock_recv_task = false`
  - block on `s_sock_recv_task_exit_sem` (bounded wait)
  - clear the task handle
- In the RX task: `xSemaphoreGive(s_sock_recv_task_exit_sem)` immediately before `vTaskDelete(NULL)`.

## 2. Improve select()/recvfrom() unblocking during socket close

**Problem:** Closing a socket from another task does not reliably and immediately unblock a thread blocked in `select()`/`recvfrom()` on all stacks/ports.

**Correction:**
- `delete_socket()` now performs a best-effort `shutdown(sock, SHUT_RDWR)` prior to `close(sock)` to encourage prompt unblocking.

## 3. recvfrom() error handling (avoid repeated error/log-spin)

**Problem:** A `recvfrom()` error only broke out of the inner per-interface loop, leaving the RX task alive and potentially re-hitting the same invalid socket each select cycle.

**Correction:**
- On `recvfrom()` failure:
  - log at `WARN`
  - close the socket (`delete_socket(sock)`)
  - set `s_interfaces[tcpip_if].sock = -1`
  - clear `s_interfaces[tcpip_if].proto = 0`
  - continue with other interfaces
- `select()` now ignores `EINTR` by retrying the loop.

## 4. Multicast group join failure must not mark protocol as ready

**Problem:** `create_pcb()` set the per-interface protocol bit even when `join_mdns_multicast_group()` failed, resulting in an interface being treated as ready for that protocol while multicast RX might be non-functional.

**Correction:**
- If `join_mdns_multicast_group()` fails, `create_pcb()` returns `false` and does **not** set the protocol bit.
- If the socket was newly created for this interface and no other protocol is active, it is closed to avoid leaking a socket with no valid membership.

## 5. Normalize socket close paths

**Problem:** Socket cleanup was split between direct `close()` calls and `delete_socket()`, risking divergence if `delete_socket()` gains additional behavior.

**Correction:**
- `create_socket()` error path now calls `delete_socket(sock)` instead of `close(sock)`.

### DIFF
--- a/components/mdns/mdns_networking_socket.c
+++ b/components/mdns/mdns_networking_socket.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -45,6 +45,8 @@ static interfaces_t s_interfaces[MDNS_MAX_INTERFACES];
 
 static const char *TAG = "mdns_networking";
 static bool s_run_sock_recv_task = false;
+static TaskHandle_t s_sock_recv_task_handle = NULL;
+static SemaphoreHandle_t s_sock_recv_task_exit_sem = NULL;
 static int create_socket(esp_netif_t *netif);
 static int join_mdns_multicast_group(int sock, esp_netif_t *netif, mdns_ip_protocol_t ip_protocol);
 
@@ -91,6 +93,12 @@ static void __attribute__((constructor)) ctor_networking_socket(void)
 
 static void delete_socket(int sock)
 {
+    if (sock < 0) {
+        return;
+    }
+
+    // Best-effort shutdown helps to unblock select()/recvfrom() on some stacks.
+    shutdown(sock, SHUT_RDWR);
     close(sock);
 }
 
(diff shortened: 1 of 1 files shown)


components/mdns/mdns_networking_socket.c
```diff
@@ -304,77 +320,100 @@ void sock_recv_task(void *arg)
...
+                memcpy(buf, recvbuf, len);
+                packet_pbuf->next = NULL;
+                packet_pbuf->payload = buf;
+                packet_pbuf->tot_len = len;
+                packet_pbuf->len = len;
+                packet->tcpip_if = tcpip_if;
+                packet->pb = packet_pbuf;
+                packet->src_port = ntohs(port);
+                memcpy(&packet->src, &addr, sizeof(esp_ip_addr_t));
+                // TODO(IDF-3651): Add the correct dest addr -- for mdns to decide multicast/unicast
```
euripedesrocha, line 369: Is it necessary to keep the TODO here? Task is still open and we are tracking it there.

euripedesrocha (approved):

components/mdns/mdns_networking_socket.c
david-cermak: There's no change wrtt tracker is this PR (it's just showing white-spaces as I added early `continue`)

Image

david-cermak: I'd leave it for now and will remove before tagging.

PR #1018, opened 2026-02-27
fix(websocket): Fix stale CLOSE_FRAME_SENT_BIT blocking PINGs after reconnect

gabsuren:
Related https://github.com/espressif/esp-protocols/pull/999

Only set CLOSE_FRAME_SENT_BIT when send_close() succeeds; clear it on every
INIT->CONNECTED transition so a reconnected connection always sends PINGs normally.
Also document the auto_reconnect/close_reconnect dependency in config struct.


Bryght-Labs-Richard (approved):
Looks good to me. Should I close #999 since this includes both commits?

david-cermak (approved):
LGTM!

PR #1026, opened 2026-03-12
fix(esp_modem): Output buffer not cleared when AT command returns empty response (IDFGH-17370)

JellevanKraaij:
## Description

When `esp_modem_at()` or `esp_modem_at_raw()` returns an empty response,
the output buffer `p_out` was not cleared. This caused callers to read
uninitialized or stale data when checking `p_out` after a successful AT
command that returns no payload (e.g. `AT` returns just `OK` with no
additional data).

This fix explicitly sets `p_out[0] = '\0'` when the response is empty,
ensuring callers can safely check the buffer without needing to
pre-initialize it themselves.

## Testing

Tested on SIM7080G with esp-idf v5.5.3 using `esp_modem_at(dce, "AT", buf, 1000)`.
Previously `buf` contained garbage when the command succeeded, now it is
correctly cleared to an empty string.

---

> [!NOTE]
> **Low Risk**
> Low risk behavioral fix in the C API: it changes `esp_modem_at()`/`esp_modem_at_raw()` to always write to `p_out` (including empty responses), which may affect callers that relied on the buffer remaining unchanged.
> 
> **Overview**
> Fixes `esp_modem_at()` and `esp_modem_at_raw()` to **always** copy the modem response into `p_out` when provided, even when the response string is empty.
> 
> This ensures callers don’t observe stale/garbage data in `p_out` after successful AT commands that return no payload (e.g., `OK` only).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 17d82dea3b2d24ee4e4f54421f3ed9b0873d3e63. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_modem/src/esp_modem_c_api.cpp
+++ b/components/esp_modem/src/esp_modem_c_api.cpp
@@ -233,12 +233,12 @@ extern "C" esp_err_t esp_modem_get_signal_quality(esp_modem_dce_t *dce_wrap, int
 
 extern "C" esp_err_t esp_modem_get_imsi(esp_modem_dce_t *dce_wrap, char *p_imsi)
 {
-    if (dce_wrap == nullptr || dce_wrap->dce == nullptr) {
+    if (dce_wrap == nullptr || dce_wrap->dce == nullptr || p_imsi == nullptr) {
         return ESP_ERR_INVALID_ARG;
     }
     std::string imsi;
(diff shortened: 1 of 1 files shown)


david-cermak (approved):
still LGTM! Thanks for additional fixes.

PR #1035, opened 2026-03-20
fix(websocket): use local test server instead of public echo.websocke…

gabsuren:


gabsuren:
@euripedesrocha thanks for the review!

PR #1042, opened 2026-04-02
[modem]: Add support for DTE-Tx hooks

david-cermak:
Add before/after transmit hooks to DTE so users can toggle a GPIO (DTR / sleep pin) around every UART write; including internal PPP frames, LCP echo-replies, and other traffic that applications cannot wrap today. This enables UART power-save modes (Quectel QSCLK, u-blox UPSV, SIMCom CSCLK, etc.) without silent data loss.

* **C++ API**: `DTE::set_transmit_hooks(before_tx, after_tx)` hooks fire in all four DTE write paths (write(uint8_t*,size_t), write(DTE_Command), send(...), and inside command(...))

* **C API* : `esp_modem_set_transmit_hooks(dce, before_tx, after_tx, user_ctx)` wraps the C++ hooks with an opaque context pointer
* **Public terminal factory**: `create_uart_terminal()` moved from private to public header, enabling C++ users to build custom DTE subclasses without private-include workarounds

* Fully additive, no breaking changes. When no hooks are registered, overhead is a single null-check per write.

### Notes
* In CMUX mode, hooks may be called concurrently from PPP and AT-command threads. Users needing DTR gating under CMUX should use an atomic refcount.
* The hooks wrap only the actual terminal->write() call, not the full command-wait cycle, so DTR stays asserted only while bytes are in flight.

---

> [!NOTE]
> **Medium Risk**
> Touches core DTE transmit paths and adds callback execution around every write, which could impact timing/concurrency (especially under CMUX) if hooks misbehave, though behavior is additive when hooks are unset.
> 
> **Overview**
> Adds **before/after transmit hooks** to `DTE` (`set_transmit_hooks`) and wraps all DTE transmit paths (`command()`, `write()`, `write(DTE_Command)`, `send()`) so callers can toggle GPIOs around every terminal write, including PPP/data traffic.
> 
> Exposes `create_uart_terminal()` as a public C++ API for building custom DTE subclasses, and adds a C API entrypoint `esp_modem_set_transmit_hooks()` (with opaque `user_ctx`) plus host tests validating hook firing, ordering, and clearing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ec901e36d30b1a3ecb00d53c7c03f71c6f00c69. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
+++ b/components/esp_modem/include/cxx_include/esp_modem_dte.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -7,9 +7,13 @@
 #pragma once
 
 #include <memory>
+#if defined(__cpp_lib_atomic_shared_ptr)
+#include <atomic>
+#endif
 #include <utility>
 #include <cstddef>
 #include <cstdint>
+#include <functional>
 #include "cxx_include/esp_modem_primitives.hpp"
 #include "cxx_include/esp_modem_terminal.hpp"
 #include "cxx_include/esp_modem_types.hpp"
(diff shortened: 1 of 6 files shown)


components/esp_modem/include/cxx_include/esp_modem_dte.hpp
```diff
@@ -242,6 +271,11 @@ class DTE : public CommandableIf {
     modem_mode mode;                                        /*!< DTE operation mode */
     std::function<bool(uint8_t *data, size_t len)> on_data; /*!< on data callback for current terminal */
     std::function<void(terminal_error err)> user_error_cb;  /*!< user callback on error event from attached terminals */
+#if defined(__cpp_lib_atomic_shared_ptr)
+    std::atomic<std::shared_ptr<const TransmitHooks>> transmit_hooks_ {}; /*!< Immutable hook pair */
+#else
+    std::shared_ptr<const TransmitHooks> transmit_hooks_;                /*!< Immutable hook pair; use std::atomic_{load,store} */
+#endif
```
david-cermak, line 278: @euripedesrocha Here's my workaround to get atomic load/store with `shared_prt<>` work across all IDF versions. (`atomic_load/store_explicit` vs. `shared_ptr<>.load` is either deprecated or not present)

Perhaps you have a better idea/suggestion how to address this?

PR #1058, opened 2026-05-22
fix(mdns): queue browse send and trigger per-family browse on GOT_IP events

gytxxsy:
### Problem

Two issues were observed in browse trigger timing:

1. Browse send could run before the queued PCB enable action completed.
2. Browse resend was coupled to `IP_EVENT_GOT_IP6`, so IPv4 updates did not trigger an immediate IPv4 browse resend.

### Fix

- Add a queued internal action: `ACTION_BROWSE_SEND_BY_IP_PROTOCOL`.
- Post this action from IP events instead of calling browse send directly in the event handler.
- Trigger browse resend by matching address family:
  - `IP_EVENT_STA_GOT_IP` / `IP_EVENT_ETH_GOT_IP` -> send browse on IPv4
  - `IP_EVENT_GOT_IP6` -> send browse on IPv6

### Result

Browse send now executes in-order with other mDNS service actions, avoiding the enable/send race.
It also resends queries on the corresponding protocol as soon as that protocol's IP event arrives.

---

## Checklist

- [ ] ~Documentation is updated as needed.~
- [ ] ~Tests are updated or added as necessary.~
- [ ] ~Code is well-commented, especially in complex areas.~

### DIFF
--- a/components/mdns/mdns_netif.c
+++ b/components/mdns/mdns_netif.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2025-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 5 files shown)


components/mdns/mdns_netif.c
zwx1995esp, line 283: Question: why we announce pcb with protocol V6 in the IPv4 address event? @gytxxsy PTAL.

components/mdns/mdns_netif.c
```diff
@@ -272,7 +292,7 @@ static void handle_system_event_for_preset(void *arg, esp_event_base_t event_bas
                     }
                     post_enable_pcb(mdns_if, MDNS_IP_PROTOCOL_V6);
                     post_announce_pcb(mdns_if, MDNS_IP_PROTOCOL_V4);
-                    mdns_priv_browse_send_all(mdns_if);
+                    post_browse_send_all_action(mdns_if);
```
zwx1995esp: I think we need to separate the V4 and V6 browse sending logic, here in the v6 event, we only send the v6 browse, and also send the v4 browse in the IPv4 event.

components/mdns/mdns_netif.c
gytxxsy: Because it is a cross-family refresh.

When IPv4 changes, mDNS may still want to send an announce on the already-running IPv6 channel so peers listening on IPv6 can refresh cached records quickly.
If IPv6 is not ready yet, that announce is effectively a no-op (it will not send).

zwx1995esp:
LGTM. Hi @david-cermak Could you please take a look at this fix? Thanks!

david-cermak (commented):
LGTM in general, but pls fix the CI -- I believe that it's this change introduced failure

gytxxsy:
> LGTM in general, but pls fix the CI -- I believe that it's this change introduced failure

Hi @david-cermak, I have updated the code to resolve the CI issue.

david-cermak (approved):
LGTM, thanks for the quick update!

Issue #1059, opened 2026-05-24
esp_websocket_client v1.6.x idf_component.yml has incorrect IDF compat — fails on v5.5 tag (refile of #947 with v5.5-tag-specific framing)


ff4415:
### Context

Refiles the symptom from #947 (closed Won't Do as "PlatformIO snapshot lag") with framing that the previous closure didn't address: the same compile failure reproduces against **Espressif's own `v5.5` git tag**, not via PlatformIO.

### Verification

I just queried the IDF tree directly via GitHub's API:

```
$ gh api "/repos/espressif/esp-idf/contents/components/tcp_transport/include/esp_transport_ws.h?ref=v5.5" \
    --jq ".content" | base64 -d | grep -nE "redir|get_redir_uri"
(NOT present at v5.5 tag)

$ gh api "/repos/espressif/esp-idf/contents/components/tcp_transport/include/esp_transport_ws.h?ref=v5.5.4" \
    --jq ".content" | base64 -d | grep -nE "redir|get_redir_uri"
167: * @brief               Returns websocket redir host for the last connection attempt
175:char* esp_transport_ws_get_redir_uri(esp_transport_handle_t t);
```

So:
- **`v5.5` tag** (= the initial `v5.5.0` release commit `8c750b08`, "change(version): Update version to 5.5.0", Jul 18 2025): **does not declare** `esp_transport_ws_get_redir_uri`.
- **`v5.5.4` tag**: declares the function at line 175.

@gabsuren's reply on #947 stated that commit `97e6f4ab` ("feat(tcp_transport): Add websocket HTTP redirect", Mar 25 2025) was "part of the official release/v5.5 branch since that date" and inferred the v5.5.0 tag would carry it. The verification above shows that's not the case — either the cherry-pick to `release/v5.5` happened between the `v5.5` and `v5.5.4` tags, or it was on a different branch. Either way: **the redirect function is absent from the `v5.5` tag exactly as cut by Espressif**.

This is independent of PlatformIO. `idf.py install` against `git clone --branch v5.5 https://github.com/espressif/esp-idf.git` reproduces the symptom from #947 — using only Espressif's official tooling and tag.

### Reproduction (Espressif-only toolchain)

```bash
git clone --branch v5.5 --depth 1 https://github.com/espressif/esp-idf.git
cd esp-idf && ./install.sh esp32s3
. ./export.sh

# In a project depending on esp_websocket_client ^1.6.0:
idf.py build
[...]

Issue #947, opened 2025-11-19
implicit declaration of function 'esp_transport_ws_get_redir_uri'; did you mean 'esp_transport_ws_get_read_opcode' (IDFGH-16826)


MiXiaoMing:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### What component are you using? If you choose Other, provide details in More Information.

esp_websocket_client

### component version

version: 1.6.0

### IDF version.

5.5.0

### More Information.

<img width="1160" height="295" alt="Image" src="https://github.com/user-attachments/assets/670b5de5-2d5b-4b2d-b79c-c0980b81b1ad" />

gabsuren:
@MiXiaoMing
I’ve checked [ESP-IDF v5.5.0](https://github.com/espressif/esp-idf/tree/release/v5.5), and the function `esp_transport_ws_get_redir_uri()` is declared here:

https://github.com/espressif/esp-idf/blob/release/v5.5/components/tcp_transport/include/esp_transport_ws.h#L175

It looks like the issue might be related to a version mismatch between your local ESP-IDF and the one expected by the project. 
Please double-check that your local ESP-IDF version exactly matches the one used remotely (v5.5.0 in this case).
Let me know if the problem persists or if you have any other questions!

Best regards,
Suren

andrezheng-git:
I'm encountering the same issue. The esp_transport_ws.h in my IDF (v5.5.0) lacks the redirect_uri function, yet the newly released WebSocket 1.6.0 uses this function. I'm certain that the local WebSocket client 1.5.0 is compatible with my IDF 5.5.0, but the problem arose with the release of 1.6.0.
commit 8c750b088c7cd857d079c0eeb495da199b359461 (HEAD, tag: v5.5)
Author: Wei Yuhan <weiyuhan@espressif.com>
Date:   Fri Jul 18 20:47:27 2025 +0800

    change(version): Update version to 5.5.0

Could subsequent commits have fixed this issue? I've applied some patches to my local IDF and I 不敢 update to the latest commit. Please ensure that WebSocket 1.6.0 is compatible with all commits in the v5.5.0 branch.

esp_transport_ws.h is as follow:
/*
[...]

PR #1062, opened 2026-05-25
fix(websocket): correct IDF version guard for HTTP redirect support

gabsuren:
Related https://github.com/espressif/esp-protocols/issues/1059 and https://github.com/espressif/esp-protocols/issues/947

Problem
`esp_websocket_client v1.6.x` introduced HTTP redirect support gated behind:
https://github.com/espressif/esp-protocols/commit/ce1560ac

```
#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
#define WS_TRANSPORT_REDIRECT_HEADER_SUPPORT    1
#endif
```
However, _esp_transport_ws_get_redir_uri_ — the only function used inside this block — was introduced via commit [_97e6f4ab_](https://github.com/espressif/esp-idf/commit/97e6f4ab0fd2d4c7bef3998fbb08e0d6de987028) which missed the v5.5.0 tag. It first appears in v5.5.1 of ESP-IDF.

Any user building against the v5.5.0 git tag gets a hard compile error:

`error: implicit declaration of function 'esp_transport_ws_get_redir_uri'`

Root Cause
The v5.5.0 tag was cut on Jul 18, 2025. Commit 97e6f4ab was cherry-picked to release/v5.5 after that date and first landed in v5.5.1. Verified directly from git:

What this does
The macro WS_TRANSPORT_REDIRECT_HEADER_SUPPORT gates an entire block of code that calls esp_transport_ws_get_redir_uri. By raising the threshold from 5.5.0 to 5.5.1:

**On IDF v5.5.0**  the condition evaluates false, the macro is never defined, the redirect block is entirely skipped by the preprocessor, and the missing function is never referenced → no compile error

On **IDF v5.5.1+**  the condition evaluates true, the macro is defined, the redirect block compiles as before → full redirect feature works

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -67,8 +67,8 @@ static const char *TAG = "websocket_client";
 #define WS_HTTP_BASIC_AUTH  "Basic "
 #define WS_HTTP_REDIRECT(code) ((code >= 300) && (code < 400))
 
-#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
-// Features supported in 5.5.0
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 1)
+// esp_transport_ws_get_redir_uri was added after the v5.5.0 tag; first appears in v5.5.1
 #define WS_TRANSPORT_REDIRECT_HEADER_SUPPORT    1
 #endif
 


glmfe (approved):

components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -67,8 +67,8 @@ static const char *TAG = "websocket_client";
 #define WS_HTTP_BASIC_AUTH  "Basic "
 #define WS_HTTP_REDIRECT(code) ((code >= 300) && (code < 400))
 
-#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
-// Features supported in 5.5.0
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 1)
+// esp_transport_ws_get_redir_uri was added after the v5.5.0 tag; first appears in v5.5.1
 #define WS_TRANSPORT_REDIRECT_HEADER_SUPPORT    1
```
david-cermak, line 72: Wouldn't it be better if we define this in the ws_transport header file?

david-cermak (approved):
LGTM, now. Just a note, how we can look at it in  future.

components/esp_websocket_client/esp_websocket_client.c
gabsuren: @david-cermak Good point. I agree this would be cleaner in the transport header. For this fix, I’d prefer to keep the change minimal to avoid expanding the backport scope, but I’ll keep this in mind for future cleanup.

PR #1066, opened 2026-05-27
fix(mdns): Fix mdns-browse to stage A/AAAA

david-cermak:
also updates browse PTR model to support TTL=0/removal

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -118,6 +118,33 @@ typedef struct mdns_result_s {
 } mdns_result_t;
 
 typedef void (*mdns_query_notify_t)(mdns_search_once_t *search);
+
+/**
+ * @brief Browse result change notifier
+ *
+ * Called once per browse result that changed in a given response packet (not
+ * once per packet). The @p result argument points at the changed entry, but it
+ * remains a live node in the internal browse cache until the callback returns.
+ *
+ * @warning @p result->next links other cached instances for this browse, not
+ *          necessarily other results that changed in the same packet. For
+ *          ordinary add/update notifications, use only @p result; do not walk
+ *          @c next, because unchanged instances may appear there.
+ *
+ * @warning Batch PTR TTL=0 ("goodbye") responses (for example when a peer
+ *          exits and Bonjour sends many removals in one packet) may invoke the
+ *          notifier once while several instances in the cache are updated.
+ *          Until this is improved, applications may walk @c next and treat
+ *          entries with @c ttl == 0 as removals. Do not assume every node on
+ *          @c next changed in the current packet.
+ *
+ * @warning If handling is deferred outside this callback, copy @p result first
+ *          (including strings and addresses). Goodbye entries may be freed when
+ *          the callback returns.
+ *
+ * @param result  The browse result that changed. See the warnings above for
+ *                use of @c result->next.
+ */
 typedef void (*mdns_browse_notify_t)(mdns_result_t *result);
 
 /**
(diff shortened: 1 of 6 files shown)


david-cermak:
@gytxxsy @zwx1995esp PTAL, as well. Making some minor changes to mdns-browse

components/mdns/mdns_browser.c
```diff
@@ -199,9 +207,156 @@ static void browse_add(mdns_browse_t *browse)
...
+    if (!r) {
+        HOOK_MALLOC_FAILED;
+        return;
+    }
+    memset(r, 0, sizeof(mdns_result_t));
+    r->instance_name = mdns_mem_strdup(instance);
+    r->service_type = mdns_mem_strdup(service);
+    r->proto = mdns_mem_strdup(proto);
+    if (!r->instance_name || !r->service_type || !r->proto) {
+        mdns_mem_free(r->instance_name);
```
zwx1995esp, line 352: I think we need a `HOOK_MALLOC_FAILED` here.

zwx1995esp (approved):
Only one NIT from my side, others LGTM.

components/mdns/include/mdns.h
```diff
@@ -118,6 +118,33 @@ typedef struct mdns_result_s {
...
 typedef void (*mdns_query_notify_t)(mdns_search_once_t *search);
+
+/**
+ * @brief Browse result change notifier
+ *
+ * Called once per browse result that changed in a given response packet (not
+ * once per packet). The @p result argument points at the changed entry, but it
+ * remains a live node in the internal browse cache until the callback returns.
+ *
+ * @warning @p result->next links other cached instances for this browse, not
```
euripedesrocha, line 129: Can this be prevented in a next version? I assume that removing this would be a breaking change now.

components/mdns/tests/host_test/pytest_mdns.py
```diff
@@ -37,6 +63,81 @@ def terminate(self):
...
+    ]
+    if srv_first:
+        cmd.append('--srv-first')
+    if goodbye_after >= 0:
+        cmd.extend(['--goodbye-after', str(goodbye_after)])
+    return cmd
+
+
+@pytest.fixture
+def bonjour_responder():
```
euripedesrocha: For this style of fixture I've preferring to use a `contextmanager` creating a scope inside the test. It would avoid the code duplication here as, at first sight the fixture differs only on parameter on proc, probably the fixture could also take a parameter.

components/mdns/tests/host_test/pytest_mdns.py
```diff
@@ -37,6 +63,81 @@ def terminate(self):
...
+    proc = subprocess.Popen(
+        _bonjour_responder_cmd(),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+    )
+    time.sleep(0.5)
+    assert proc.poll() is None, 'Bonjour-order responder failed to start (need eth0 and port 5353?)'
```
euripedesrocha: Why `poll()` must be None?

components/mdns/mdns_browser.c
```diff
@@ -199,9 +207,156 @@ static void browse_add(mdns_browse_t *browse)
...
 }
 
+static esp_err_t add_browse_result(mdns_browse_sync_t *sync_browse, mdns_result_t *r);
+
 /**
  * @brief  Called from packet parser to find matching running search
  */
+mdns_browse_t *mdns_priv_browse_find_ptr(mdns_name_t *name)
+{
+    mdns_browse_t *b = s_browse;
```
euripedesrocha, line 220: Is it thread safe to use `s_browse`? And should we make `b` `const`?

components/mdns/mdns_browser.c
```diff
@@ -199,9 +207,156 @@ static void browse_add(mdns_browse_t *browse)
...
+ *       is already allocated for a *different* browse, this function returns
+ *       the existing object unchanged — callers that compare
+ *       out_sync_browse->browse against the current browse will silently
+ *       skip the update.  This is acceptable because mDNS answers for
+ *       multiple browsed service types in a single packet are uncommon.
+ */
+mdns_browse_sync_t *mdns_priv_browse_ensure_sync(mdns_browse_t *browse, mdns_browse_sync_t *sync)
+{
+    if (!browse) {
+        return sync;
```
euripedesrocha, line 248: Do we have any guidance for the user in the case we return sync? User can validate the pointer but what are the actions that follow?

components/mdns/mdns_receive.c
```diff
@@ -598,10 +598,19 @@ static void mdns_parse_packet(mdns_rx_packet_t *packet)
     bool do_not_reply = false;
     mdns_search_once_t *search_result = NULL;
     mdns_browse_t *browse_result = NULL;
+    /*
```
euripedesrocha, line 601: I'm missing some documentation change about this. How much users are affected by this behavior?

components/mdns/include/mdns.h
david-cermak: Yes, it will be addressed later, as mentioned in the issue

components/mdns/tests/host_test/pytest_mdns.py
david-cermak: this is actually pytest original (not the esp-pytest territory) , but some code reuse could be done. Have reworked.

components/mdns/tests/host_test/pytest_mdns.py
david-cermak: added a note (means the bonjour service completed)

components/mdns/mdns_browser.c
david-cermak: Yes, added a note. couldn't be const because the function returns the active browse node

components/mdns/mdns_browser.c
david-cermak: added a note

components/mdns/mdns_receive.c
david-cermak: added a doxy note to the api docs

components/mdns/mdns_browser.c
david-cermak: fixed, thanks!

euripedesrocha (approved):

PR #1067, opened 2026-05-27
[mdns] Support for fuzzing browses

david-cermak:
Related to https://github.com/espressif/esp-protocols/pull/1066

### DIFF
--- a/components/mdns/mdns_browser.c
+++ b/components/mdns/mdns_browser.c
@@ -412,6 +412,14 @@ static void sync_browse_result_link_free(mdns_browse_sync_t *browse_sync)
     mdns_mem_free(browse_sync);
 }
 
+void mdns_priv_browse_sync_free(mdns_browse_sync_t *browse_sync)
+{
+    if (!browse_sync) {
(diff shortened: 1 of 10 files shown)


components/mdns/mdns_browser.c
```diff
@@ -412,6 +412,14 @@ static void sync_browse_result_link_free(mdns_browse_sync_t *browse_sync)
     mdns_mem_free(browse_sync);
 }
 
+void mdns_priv_browse_sync_free(mdns_browse_sync_t *browse_sync)
+{
+    if (!browse_sync) {
```
euripedesrocha, line 417: Should we assert or warn user?

euripedesrocha (approved):

components/mdns/mdns_browser.c
david-cermak: this is actually a private API (only used by the receiver for now to clear packet on error).
I wouldn't add an assert nor warning -- to be inline with the standard `free(NULL)` silent pass

PR #1061, opened 2026-05-25
feat(lws): add Kconfig toggles + bump submodule

vikramdattu:
## Summary

Two changes to `components/libwebsockets`:

1. **Three Kconfig toggles** so consumers can tune libwebsockets without forking the component — `LWS_WITH_EXTENSIONS`, `LWS_WITH_FILE_OPS`, `LWS_WITH_THREADPOOL`.
2. **Submodule bump** from `a74362ff` → `d659d4f2` (93 commits forward on the warmcat 4.3.x mainline).

## Motivation

The AWS KVS WebRTC SDK port we maintain in espressif-port carries a local `components/libwebsockets/` whose only delta vs the one here is a CMakeLists that force-sets three knobs (extensions on, file_ops off, threadpool on) for the WebRTC-signaling workload. With this PR those workloads can flip Kconfig instead of forking — and we can drop the local copy in espressif-port entirely and consume `espressif/libwebsockets` from the registry.

The submodule bump is independent: we already pin a newer SHA on the same warmcat 4.3.x mainline (no patches, no fork), and it picks up upstream fixes including the Linux-host build path that was landed via an earlier Espressif contribution.

## Backwards compatibility

All three Kconfig defaults preserve the previously-baked behaviour:

| Knob | Default | What changes when flipped |
|---|---|---|
| `LWS_WITH_EXTENSIONS` | **n** | extensions on — needed for permessage-deflate (RFC 7692). Caller also needs zlib at link time. |
| `LWS_WITH_FILE_OPS` | **y** | drop the libwebsockets server-side VFS layer; saves flash on clients that never serve static files. |
| `LWS_WITH_THREADPOOL` | **n** | enable lws's managed worker thread pool (pthread-backed); used by callers that off-load parallel TLS handshake work. |

Existing consumers — `examples/client/`, downstream registry users on `espressif/libwebsockets@4.3.3~1` — see zero behaviour change.

### Note on existing `port/lws_port.c` wraps

This PR doesn't touch the existing `__wrap_mbedtls_ssl_handshake_step` / `__wrap_lws_adopt_descriptor_vhost` linker hooks. Consumers that migrate from a fork inherit them as-is. The handshake-step wrap converts mbedTLS's async `WANT_READ/WANT_WRITE` returns into a synchronous busy-loop, which is fine when lws's mbedTLS BIO is blocking (the inner call blocks instead of returning WANT_*) — that's the case for ESP-IDF's TLS port. Flagging here so reviewers / future-migrators know.

[...]

PR #1070, opened 2026-06-04
feat(lws): bump submodule for mbedTLS 4 support; allow IDF v6.0

vikramdattu:
## Summary

- Bump **libwebsockets** submodule from \`a74362ff\` → the **\`v5.0.0\` release tag** (\`2491a1b10\`) — warmcat's first release carrying the mbedTLS 4 migration. (Earlier revisions of this PR pinned interim \`main\` SHAs out of necessity; that's no longer needed.)
- Drop the **\`idf <6.0\`** cap from \`idf_component.yml\` — IDF v6.0 ships mbedTLS 4 by default and libwebsockets v5.0.0 supports it. Component version can become a clean **\`5.0.0\`** at bump time (tracking upstream, like today's \`4.3.3~x\`).

## Verification

- \`examples/client\` builds clean on **IDF release/v6.0** (target esp32) — exit 0, full link.
- Verified against the v5.0.0 tag on IDF v6.0.2 (esp32, mbedTLS 4.1.0) — builds clean, full link.

## Notes

- Sequenced after PR #1061 (libwebsockets Kconfig + 4.3-stable bump). #1061 stays on 4.3-stable; this PR moves to main.

---

> [!NOTE]
> **Medium Risk**
> Touches TLS/mbedTLS integration, errno/TLS linking, and networking port shims across IDF versions; scope is mostly build/example/CI with gated v6-only paths.
> 
> **Overview**
> **Enables the libwebsockets component and CI on ESP-IDF v6.0** (mbedTLS 4 / picolibc) while keeping v5.3–v5.5 in the matrix.
> 
> The **CMake port** is the bulk of the change: it forces mbedTLS v4 detection when IDF ships mbedTLS 4, relaxes lws `-Werror` and PSA redefinition noise, fixes **errno** via `LWIP_ERRNO_INCLUDE` and **esp_libc** include ordering, adds **tf-psa-crypto** and v6-only component include paths, and declares **`esp_driver_gpio`** as a public dependency. **`port/lws_port.c`** adds an **`if_nametoindex`** fallback for older IDF lwIP.
> 
> The **client example** applies skip-CN TLS for mutual auth, reassembles fragmented binary receives for tests, resets the task WDT around blocking **`lws_service`**, pulls **cJSON** from the registry on v6, and uses a **large partition** default. **Pytest** and **CI ignore warnings** are updated for new logs and SSL API behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1d4b54389c9affbae5421a30d9b52c72fcdc79d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/libwebsockets/idf_component.yml
+++ b/components/libwebsockets/idf_component.yml
@@ -2,4 +2,4 @@ version: "4.3.3~3"
 url: https://github.com/espressif/esp-protocols/tree/master/components/libwebsockets
 description: The component provides a simple ESP-IDF port of libwebsockets client.
 dependencies:
-  idf: '>=5.3,<6.0'
+  idf: '>=5.3'
--- a/.github/workflows/lws_build.yml
+++ b/.github/workflows/lws_build.yml
@@ -13,7 +13,7 @@ jobs:
     name: Libwebsockets build
     strategy:
       matrix:
-        idf_ver: ["release-v5.3", "release-v5.4", "release-v5.5"]
+        idf_ver: ["release-v5.3", "release-v5.4", "release-v5.5", "release-v6.0"]
         test: [ { app: example, path: "examples/client" }]
     runs-on: ubuntu-22.04
     container: espressif/idf:${{ matrix.idf_ver }}
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        idf_ver: ["release-v5.3", "release-v5.4", "release-v5.5"]
+        idf_ver: ["release-v5.3", "release-v5.4", "release-v5.5", "release-v6.0"]
         idf_target: ["esp32"]
         test: [ { app: example, path: "examples/client" }]
     runs-on:
--- a/ci/ignore_build_warnings.txt
+++ b/ci/ignore_build_warnings.txt
@@ -7,3 +7,4 @@ Warning: Deprecated: Command 'extract_public_key' is deprecated. Use 'extract-pu
 warning: unknown kconfig symbol 'EXAMPLE_ETH_PHY_IP101'
 WARNING: The following Kconfig variables were used in "if" clauses, but not
 warning: unknown kconfig symbol 'LIBC_NEWLIB'
+warning: 'MBEDTLS_PSA_BUILTIN_[A-Z0-9_]+' redefined
--- a/components/libwebsockets/CMakeLists.txt
+++ b/components/libwebsockets/CMakeLists.txt
@@ -65,10 +65,27 @@ if(_mbedtls_major MATCHES "[ \t]+([0-9]+)" AND CMAKE_MATCH_1 GREATER_EQUAL 4)
     set(LWS_HAVE_MBEDTLS_V4 1 CACHE INTERNAL "mbedtls 4 detected on IDF")
 endif()
 
+# Stop lws adding its own -Werror (the IDF build already sets -Werror globally).
+set(DISABLE_WERROR ON CACHE BOOL "lws: don't treat warnings as errors" FORCE)
+
 add_subdirectory(libwebsockets)
 
-# pollfd.c at this lws pin has an unused 'vpt' on FreeRTOS; drop after the next bump.
-target_compile_options(websockets PRIVATE -Wno-unused-variable)
+# The IDF build compiles everything with a global -Werror.  IDF v6's mbedtls
+# headers emit a harmless MBEDTLS_PSA_BUILTIN_ALG_HMAC redefinition warning
+# (esp_config.h vs tf-psa-crypto) that lws's sources pull in via libwebsockets.h
+# -> mbedtls/ssl.h; -Wno-error keeps that (and the unused 'vpt' in this pin's
+# pollfd.c) from failing the build.  See esp-idf esp_config.h redefinition bug.
+target_compile_options(websockets PRIVATE -Wno-unused-variable -Wno-error)
+
+# IDF v6's libc (picolibc) makes errno thread-local. With lws's LWIP_PROVIDE_ERRNO
+# now gated off on v6, lwip's errno.h forwards to the system errno - but lws's
+# hardcoded include list (lib/CMakeLists.txt) puts .../lwip/src/include/lwip on the
+# path, so a bare <errno.h> resolves to lwip's own errno.h (LWIP_ERRNO_STDINCLUDE
+# would re-include itself). Point lwip at <sys/errno.h> instead, which is not
+# shadowed and pulls picolibc's TLS errno + E* macros via esp_libc's #include_next.
+# Otherwise lws objects fail with "errno undeclared" (compile) or "TLS definition
+# ... mismatches non-TLS reference" (link).
+target_compile_definitions(websockets PRIVATE "LWIP_ERRNO_INCLUDE=<sys/errno.h>")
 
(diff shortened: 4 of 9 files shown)


vikramdattu:
Hi @david-cermak — heads up on the clang-tidy failure: it's not the
code in this PR, it's an upstream `libwebsockets` bug.

At the SHA we pin (`7726361c`) — and at every SHA after `a68609413
"qpack"` (May 20) — the tree declares a gitlink
`minimal-examples-lowlevel/api-tests/api-test-qpack/qifs` but ships no
`.gitmodules` file. So `git submodule update --init --recursive`
fails. I checked: `.gitmodules` was never in the repo; Andy Green
added the gitlink without registering it. Bisecting to an earlier
lws SHA can't escape it without losing the mbedtls4 commit.

I've filed the upstream fix: warmcat/libwebsockets#3597.

Two paths:

1. Wait for the warmcat maintainer to review and merge #3597 — could
   be a few days.
2. Interim workaround in this PR's CI: replace the recursive
   submodule init with explicit per-component non-recursive inits
   (scoped to lws, doesn't touch asio/mosquitto).

Which do you prefer?

david-cermak:
> * Wait for the warmcat maintainer to review and merge #3597 — could
>   be a few days.
> * Interim workaround in this PR's CI: replace the recursive
>   submodule init with explicit per-component non-recursive inits
>   (scoped to lws, doesn't touch asio/mosquitto).

Both options are fine with me. If the upstream fix gets merged within days, I would wait.

components/libwebsockets/idf_component.yml
```diff
@@ -1,5 +1,5 @@
-version: "4.3.3~1"
+version: "4.3.3~2"
```
vikramdattu: @david-cermak the version and the changelog above will pick the commit title automatically?

david-cermak: @vikramdattu  see https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md#release-process

the commit (when created by the bump command) will update the version files and changelog (based on commit delta from the previous tag)

david-cermak: > when created by the bump command

but that doesn't support these `~postfix` numbers (yet), so you'd need to edit the commit.

Here's an example https://github.com/espressif/esp-protocols/pull/1071

PS: If you like, we can merge #1071 to release `v4.3.3~2` for now, and after the upstream maintainers fix the above, we can merge this one to release `v4.3.3~3`

vikramdattu:
Rebase needed once https://github.com/warmcat/libwebsockets/pull/3600 and https://github.com/warmcat/libwebsockets/pull/3601 and https://github.com/warmcat/libwebsockets/pull/3602 are merged

vikramdattu:
@david-cermak heads-up for when you set the component version on this PR — it'll be a **minor bump + a `~` snapshot suffix**, not a routine `~N` revision.

The lws submodule is pinned to a warmcat **`main` snapshot** (`da5d6f5f9`, `git describe` = `v4.5.0-590-gda5d6f5f9`), not a release tag — because the **mbedTLS-4 / IDF-v6 support isn't in any tagged v4.5.x release yet**; it only exists on `main`. So the component's lws version moves from `4.3.3` up to the ~`4.5.x` line (minor jump), and the `~1` denotes "esp-protocols snapshot of that main state." Same rationale is noted inline in `idf_component.yml`.

(We pinned an older/settled main commit deliberately — warmcat force-pushes `main` and has orphaned our pins before; a 1–2-day-old commit is likelier to survive. Once a tagged lws release carries mbedTLS-4, we can switch to the tag.)

vikramdattu:
Update: warmcat released **v5.0.0** — the first release tag carrying the mbedTLS 4 migration. Pushed `13f0652` repinning the submodule from the interim main snapshot (`da5d6f5f9`) to the tag (`2491a1b10`).

This resolves the two open threads above:
- No more SHA-pinning against a force-pushed `main` — the earlier "stable snapshot" dance is gone for good.
- @david-cermak re. the component version: my earlier note about a **minor bump + `~` snapshot suffix** no longer applies. With the submodule on a proper release tag, the component version can simply become **`5.0.0`** at bump time (matching upstream, same convention as the current `4.3.3~x` tracking 4.3.x).

Also dropped the stale "unused `vpt`" part of the `-Wno-error` comment — the pollfd.c guard fix (warmcat/libwebsockets#3596) is included in v5.0.0. The `-Wno-error` itself is still needed for IDF v6's `MBEDTLS_PSA_BUILTIN_*` redefinition noise.

Re-verified: `examples/client` builds clean on **IDF v6.0.2** (esp32, mbedTLS 4.1.0) against the tag.

david-cermak:
> Update: warmcat released v5.0.0

@vikramdattu Is this ready for review, then? I think we can at least merge it to master once CI's green

vikramdattu:
Root-caused the on-target failures and pushed the fix (dd69483).

**What failed:** all four `Target test` jobs (v5.3–v6.0) died in the **mutual_auth** case with:

```
CLIENT_CONNECTION_ERROR:   ! The certificate Common Name (CN) does not match with the expected CN
```

**Root cause — a latent example bug, not a v5.0.0 TLS regression:** `sdkconfig.ci.mutual_auth` sets `WS_OVER_TLS_SKIP_COMMON_NAME_CHECK=y`, but the example only applied `LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK` when `SERVER_AUTH` was also selected — for mutual auth the flag was silently dropped. The CI server is reached by IP with a fixed-CN test certificate, so CN verification can never succeed there.

Two things hid this until today:
1. The pre-v5.0.0 ESP TLS wrapper never performed CN verification at all. v5.0.0 removed the wrapper (`mbedtls: remove wrapper and use directly`) and does real mbedTLS certificate verification — correct behavior, stricter than before.
2. These on-target tests had **never actually run on this PR**: they were gated behind the old `release-v5.3` build failure, which the v5.0.0 pin fixed. First time running → first time failing.

The fix makes the example honour the Kconfig option for both auth modes. With the flag set, lws v5.0.0 skips `mbedtls_ssl_set_hostname()` and masks `BADCERT_CN_MISMATCH` post-handshake, which is also the right shape for mbedTLS ≥3.6.4's hostname-verification hardening (never-set hostname + `VERIFY_REQUIRED` now hard-fails there; lws uses `VERIFY_OPTIONAL` + post-check, so both mbedTLS 3.6.x and 4.x paths are fine).

One more observation from the (passing) plain-TLS phase worth a follow-up: the TLS handshake takes ~6 s on esp32 and trips the task watchdog on `main` (`E task_wdt: … main (CPU 0)`) before recovering. Doesn't fail the test, but the handshake-long blocking of the service loop deserves a look — possibly related to what warmcat/libwebsockets#3330 was poking at.

Verified locally: mutual_auth config builds clean for esp32 on IDF v6.0.2.

vikramdattu:
> >  Update: warmcat released v5.0.0
>
> @vikramdattu Is this ready for review, then? I think we can at least merge it to master once CI's green

@david-cermak this can be reviewed and merged now. Please note that libwebsockets/ pin is now v5.0.0, so release version you may want to use will be in sync with that.

david-cermak (approved):

PR #1072, opened 2026-06-09
Fix wrong "truncated AAAA" error

littleant:
## Description

AAAA messages are always `MDNS_ANSWER_AAAA_SIZE` ([16 bytes](https://datatracker.ietf.org/doc/html/rfc3596#section-2.2)) long.
The [`esp_ip6_addr_t`](https://github.com/espressif/esp-idf/blob/bc6616854ea6303c538b97de66b7aaf290f8f1b0/components/esp_netif/include/esp_netif_ip_addr.h#L102) contains an additional `zone`-value, which makes the struct bigger than 16 bytes. This causes the condition to always fail, when it shouldn't.

## Testing

I experience the error, when I enable `MDNS_ENABLE_DEBUG_PRINTS` in my [basic_thread_border_router](https://github.com/espressif/esp-thread-br/tree/main/examples/basic_thread_border_router) project. With my fix the error disappears and displays the AAAA responses.

### DIFF
--- a/components/mdns/mdns_debug.c
+++ b/components/mdns/mdns_debug.c
@@ -293,12 +293,12 @@ void static dbg_packet(const uint8_t *data, size_t len)
                 }
                 dbg_printf("\n");
             } else if (type == MDNS_TYPE_AAAA) {
-                if (data_len < sizeof(esp_ip6_addr_t)) {
+                if (data_len < MDNS_ANSWER_AAAA_SIZE) {
                     dbg_printf("ERROR: truncated AAAA\n");
                     continue;
                 }
-                esp_ip6_addr_t ip6;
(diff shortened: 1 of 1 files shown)


david-cermak (approved):
Thanks for the fix!

david-cermak:
@littleant Could you please fix the commit message -> https://github.com/espressif/esp-protocols/actions/runs/27187728252/job/80314632759?pr=1072

It's just enough to add a prefix `"fix(mdns): Fix....."`

For more details, please check the contribution guide: https://github.com/espressif/esp-protocols/blob/master/CONTRIBUTING.md

david-cermak:
The new commit message looks good, but there are two commits now, in this PR.
Please squash them into a single commit (+ use the new commit message) and force-push. Thanks!

david-cermak:
Thank you for the quick update for your contribution 👍

PR #1085, opened 2026-06-26
feat(websocket): interruptable wait timeout

gabsuren:
Based on https://github.com/espressif/esp-protocols/pull/607

Previously, calling `esp_websocket_client_set_reconnect_timeout()` while a reconnection delay was already in progress had no immediate effect this made dynamic backoff strategies unreliable.

---

> [!NOTE]
> **Medium Risk**
> Touches FreeRTOS event-group timing and the reconnect state machine; behavior change for apps using dynamic backoff, but scope is limited to auto-reconnect wait logic.
> 
> **Overview**
> **Reconnect backoff can now take effect while a delay is already running.** `esp_websocket_client_set_reconnect_timeout()` sets `WAKEUP_BIT` on the client event group so the websocket task leaves its wait early and re-evaluates the remaining delay against the updated `wait_timeout_ms`.
> 
> In **`WEBSOCKET_STATE_WAIT_TIMEOUT`**, the task no longer blocks for half the timeout on a coarse timer. It computes **remaining** time from `reconnect_tick_ms`, waits on `REQUESTED_STOP_BIT | WAKEUP_BIT` for that interval (minimum one tick), and the state-machine branch uses `>=` with an explicit `wait_timeout_ms <= 0` guard so reconnect timing stays consistent.
> 
> **`WAKEUP_BIT`** is cleared on client start and at the beginning of each wait loop iteration. Public API notes in `esp_websocket_client.h` now state that changing the reconnect timeout **updates an active** reconnection delay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 848c6a1931d5e6a7708eebc5b9a991c1359978f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -76,6 +76,7 @@ const static int STOPPED_BIT = BIT0;
 const static int CLOSE_FRAME_SENT_BIT = BIT1;   // Indicates that a close frame was sent by the client
 // and we are waiting for the server to continue with clean close
 const static int REQUESTED_STOP_BIT = BIT2;     // Indicates that a client stop has been requested
+const static int WAKEUP_BIT = BIT3;             // Wakes the reconnect wait early (e.g. on timeout change)
 
 ESP_EVENT_DEFINE_BASE(WEBSOCKET_EVENTS);
 
@@ -1315,8 +1316,9 @@ static void esp_websocket_client_task(void *pv)
             }
             break;
         case WEBSOCKET_STATE_WAIT_TIMEOUT:
-
-            if (_tick_get_ms() - client->reconnect_tick_ms > client->wait_timeout_ms) {
+            int wait_timeout_ms = client->wait_timeout_ms;
+            if (wait_timeout_ms <= 0 ||
+                    _tick_get_ms() - client->reconnect_tick_ms >= (uint64_t)wait_timeout_ms) {
                 client->state = WEBSOCKET_STATE_INIT;
                 client->reconnect_tick_ms = _tick_get_ms();
                 ESP_LOGD(TAG, "Reconnecting...");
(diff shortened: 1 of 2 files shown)


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -1385,8 +1386,13 @@ static void esp_websocket_client_task(void *pv)
...
             }
         } else if (WEBSOCKET_STATE_WAIT_TIMEOUT == client->state) {
-            // waiting for reconnection or a request to stop the client...
-            xEventGroupWaitBits(client->status_bits, REQUESTED_STOP_BIT, false, true, client->wait_timeout_ms / 2 / portTICK_PERIOD_MS);
+            xEventGroupClearBits(client->status_bits, WAKEUP_BIT);
+            uint64_t elapsed_ms = _tick_get_ms() - client->reconnect_tick_ms;
+            if (client->run && elapsed_ms < (uint64_t)client->wait_timeout_ms) {
+                uint32_t remaining_ms = (uint32_t)((uint64_t)client->wait_timeout_ms - elapsed_ms);
+                xEventGroupWaitBits(client->status_bits, REQUESTED_STOP_BIT | WAKEUP_BIT,
+                                    false, false, pdMS_TO_TICKS(remaining_ms) + 1);
```
david-cermak: why ` + 1` ?

david-cermak (commented):
LGTM

components/esp_websocket_client/esp_websocket_client.c
gabsuren: It was there to avoid a zero-tick wait when `remaining_ms` is shorter than one RTOS tick. I changed it to only force `1` tick when the converted timeout is `0`, instead of adding one extra tick to every wait.

david-cermak (approved):

gabsuren:
@david-cermak thank you for the review!

PR #1086, opened 2026-06-29
fix(mdns): fix deadlock caused by re-acquiring the service lock

gytxxsy:
## Description
This PR fixes a self-deadlock caused by nested acquisition of the mDNS service lock.

`service_task()` holds the service lock while processing `ACTION_BROWSE_SYNC`. In this path, `browse_sync()` may call `mdns_query_results_free()`, which attempts to acquire the same lock again.

This PR switches the internal call site from `mdns_query_results_free()` to mdns_priv_query_results_free(), avoiding nested acquisition of the service lock.

We also fixed an existing lock bug in `mdns_unregister_netif()` where the function took the service lock twice and never released it on the exit path.

---

## Checklist

- [ ] ~Documentation is updated as needed`.~
- [ ] ~Tests are updated or added as necessary.~
- [ ] ~Code is well-commented, especially in complex areas.~

---

> [!NOTE]
> **Medium Risk**
> Touches core mDNS locking and browse teardown on the service task; incorrect lock pairing could still deadlock or corrupt shared state, but the changes are small and directly fix known lock bugs.
> 
> **Overview**
> **Browse sync** now frees expired browse results with `mdns_priv_query_results_free()` instead of the public `mdns_query_results_free()` when `result->ttl == 0`. The public helper takes the mDNS service lock; `browse_sync()` runs on the service task while that lock is already held, so the old call could self-deadlock on a non-recursive mutex.
> 
> **`mdns_unregister_netif()`** fixes a lock bug: the exit path called `mdns_priv_service_lock()` a second time instead of `mdns_priv_service_unlock()`, which could leave the service mutex held after unregister.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88c68330000f0e3ea23014a2259a8579625ba332. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/mdns/mdns_netif.c
+++ b/components/mdns/mdns_netif.c
@@ -453,6 +453,6 @@ esp_err_t mdns_unregister_netif(esp_netif_t *esp_netif)
             break;
         }
     }
-    mdns_priv_service_lock();
+    mdns_priv_service_unlock();
     return err;
 }
--- a/components/mdns/mdns_browser.c
+++ b/components/mdns/mdns_browser.c
@@ -76,7 +76,7 @@ static void browse_sync(mdns_browse_sync_t *browse_sync)
             queueDetach(mdns_result_t, browse->result, result);
             // Just free current result
             result->next = NULL;
-            mdns_query_results_free(result);
+            mdns_priv_query_results_free(result);
         }
         sync_result = sync_result->next;
     }


gytxxsy:
@david-cermak 
Could you please take a look of this PR? Thanks.

components/mdns/mdns_netif.c
```diff
@@ -453,6 +453,6 @@ esp_err_t mdns_unregister_netif(esp_netif_t *esp_netif)
             break;
         }
     }
-    mdns_priv_service_lock();
+    mdns_priv_service_unlock();
```
david-cermak, line 456: Thanks for the fix 👍

components/mdns/mdns_service.c
```diff
@@ -31,8 +31,8 @@
 #endif
 #define MDNS_TASK_AFFINITY          CONFIG_MDNS_TASK_AFFINITY
 
-#define MDNS_SERVICE_LOCK()     xSemaphoreTake(s_service_semaphore, portMAX_DELAY)
-#define MDNS_SERVICE_UNLOCK()   xSemaphoreGive(s_service_semaphore)
+#define MDNS_SERVICE_LOCK()     xSemaphoreTakeRecursive(s_service_semaphore, portMAX_DELAY)
+#define MDNS_SERVICE_UNLOCK()   xSemaphoreGiveRecursive(s_service_semaphore)
```
david-cermak: Instead of this, I'd suggest fixing the root-cause of the deadlock

mdns_browser.c:79:

mdns_query_results_free(result); `->` mdns_priv_query_results_free(result);

david-cermak (changes requested):
Thanks for the fix, I'd still suggest to fix the original issue first.
We can then make the lock recursive, in an unrelated MR (or with some refactor)

components/mdns/mdns_service.c
gytxxsy: This issue has been discussed before, and the conclusion was that using a recursive lock would be a good approach. I wonder if we could simply use a recursive lock to resolve this issue？
https://github.com/espressif/esp-protocols/pull/745

david-cermak: Let's fix the regression first, then we can update the lock.

I'm not a big fan of using recursive lock here (as mentioned before, it's like a watchdog -- helpful but could hide certain issues). lets discuss in some future PR

gytxxsy: OK, I have used an internal function that does not acquire the lock to fix this deadlock issue.

david-cermak (approved):
Thank you!

david-cermak:
The issue caught by the fuzzer is pre-PR, will fix soon

zwx1995esp:
@david-cermak Thanks for merging this PR, may I ask a bump for mdns and we can use the latest mdns code via the managed components?

PR #1102, opened 2026-07-17
fix(modem): Make set_apn() update the PDP context

### LINKED ISSUE #1077: APN protocol forced to "IP" by esp_modem_set_mode().
### Answers checklist.

### General issue report

This is half bug, half feature request. They are two sides of the same issue, so I have kept them as one. The bug has the higher priority in my mind as we can work around the feature part.

###This is the bug portion.

It appears esp_modem_set_mode() always sends three commands:
```
D (73678) command_lib: generic_command command ATE0
D (73685) command_lib: Response: OK
D (73687) command_lib: generic_command command AT+CGDCONT=1,"IP","telstra.internet"
D (73719) command_lib: Response: OK
D (73720) command_lib: generic_command command ATD*99#
D (73757) command_lib: Response: CONNECT 150000000
```

The "IP" field is the problem, because we previously sent this command:
```
D (70503) command_lib: generic_command command AT+CGDCONT=1,"IPV6","telstra.internet"
D (70524) command_lib: Response: AT+CGDCONT=1,"IPV6","telstra.internet" OK
```
but the latter command overrides this. I don't think the AT+CGDCONT is necessary here, though it is a handy safeguard; as long as it sends the right information.

### This is the feature request portion.
We need to be able to select IPv4 single stack, IPv4/IPv6 dual stack, or IPv6 single stack.

We also need to be able to set the connection id in the context to either 1 (AT&T, others) or 3 (Verizon). We may need to use other cid values in future.

The ideal solution would be that we are able to specify the PDP context in full, and that context be used for the next network registration. We can do that now, by sending an AT+CGDCONT command, but it does not stick.

I did try this:

```
  // update the esp_modem master PDP structure
  esp_modem_PdpContext_t new_pdp;
  new_pdp.context_id     = op.cid;
  new_pdp.apn            = op.apn;
  new_pdp.protocol_type  = op.stack;
  esp_modem_set_pdp_context( dce, &new_pdp );
```
This compiled and ran ok, but set data mode still sent "IP".

I've proven the point by disabling the troublesome code, and it resolves the issue, but I would prefer to use unmodified code as much as possible to avoid future maintenance issues.

david-cermak:
Closes https://github.com/espressif/esp-protocols/issues/1077

* [fix(modem): Make set_apn() update the PDP context](https://github.com/espressif/esp-protocols/pull/1102/changes/b7087f86f34899e05c0d963779e69e0b69ca87f0)
 - The bugfix part

* [feat(modem): Support for custom APN/proto config](https://github.com/espressif/esp-protocols/pull/1102/changes/3684abae0a0d3036fcc7e74f7f535f8cd8d77dea)
 - The feature part

* [fix(modem): Fix test loopback tear-down race](https://github.com/espressif/esp-protocols/pull/1102/changes/6e98171bb4689b4bbfa34eea0866d797a7880426)
 - host test issue in mock-terminal

---

> [!NOTE]
> **Medium Risk**
> Changes affect cellular attach (`AT+CGDCONT`) and stored PDP state on every data-mode entry; `set_apn` behavior is intentionally different but fixes a real bug. Teardown fixes are test-only but touch async callback lifetime.
> 
> **Overview**
> **esp_modem 2.1.0** extends cellular attach configuration and fixes how the stored PDP context is updated at runtime.
> 
> **PDP / APN configuration:** `esp_modem_dce_config` and `ESP_MODEM_DCE_DEFAULT_CONFIG` now include **context id (CID)** and **protocol type** (`IP` / `IPV6` / `IPV4V6`), with `0` / `NULL` keeping defaults (1 / `IP`). `GenericModule` applies these when building the internal `PdpContext`, so `setup_data_mode()` emits the matching `AT+CGDCONT`. C API paths (`esp_modem_set_pdp_context`, `esp_modem_sqn_gm02s_connect`) use the same defaulting rules and **persist** the context on the module before sending AT.
> 
> **`set_apn` fix:** New `GenericModule::set_apn()` updates only the APN on the existing PDP context. **`esp_modem_set_apn()`** now calls that instead of replacing the whole context (which previously dropped protocol type and CID). Full PDP changes remain on **`esp_modem_set_pdp_context()`**.
> 
> **Tests / teardown:** Host **`LoopbackTerm`** joins pending `std::async` work in `stop()`/`finish_async()`, uses `cb_lock` and a `stopping` flag to avoid use-after-free when clearing read callbacks, and records **`get_last_command()`** for new PDP host tests. Docs describe runtime vs create-time PDP configuration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef7991881bcb0486eebb1fd3d248f1853640acdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_modem/generate/src/esp_modem_modules.cpp
+++ b/components/esp_modem/generate/src/esp_modem_modules.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2021-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -13,7 +13,15 @@
 namespace esp_modem {
 
 GenericModule::GenericModule(std::shared_ptr<DTE> dte, const dce_config *config) :
-    dte(std::move(dte)), pdp(std::make_unique<PdpContext>(config->apn)) {}
+    dte(std::move(dte)), pdp(std::make_unique<PdpContext>(config->apn ? config->apn : ""))
+{
+    if (config->context_id != 0) {
+        pdp->context_id = config->context_id;
+    }
+    if (config->protocol_type != nullptr) {
+        pdp->protocol_type = config->protocol_type;
+    }
+}
 
 
 #include "esp_modem_command_declare_helper.inc"
--- a/components/esp_modem/.cz.yaml
+++ b/components/esp_modem/.cz.yaml
@@ -3,6 +3,6 @@ commitizen:
   bump_message: 'bump(modem): $current_version -> $new_version'
   pre_bump_hooks: python ../../ci/changelog.py esp_modem
   tag_format: modem-v$version
-  version: 2.0.2
+  version: 2.1.0
   version_files:
   - idf_component.yml
--- a/components/esp_modem/CHANGELOG.md
+++ b/components/esp_modem/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## [2.1.0](https://github.com/espressif/esp-protocols/commits/modem-v2.1.0)
+
+### Features
+
+- Support for custom APN/proto config ([3684abae](https://github.com/espressif/esp-protocols/commit/3684abae))
+
+### Bug Fixes
+
+- Fix test loopback tear-down race ([6e98171b](https://github.com/espressif/esp-protocols/commit/6e98171b))
+- Make set_apn() update the PDP context ([b7087f86](https://github.com/espressif/esp-protocols/commit/b7087f86))
+- Fix terminal-callback use-after-free and harden task teardown ([01004182](https://github.com/espressif/esp-protocols/commit/01004182))
+- Update CI API check to ignore C comments ([ccc3d4a8](https://github.com/espressif/esp-protocols/commit/ccc3d4a8))
+- Add URC-handler host test ([93db64db](https://github.com/espressif/esp-protocols/commit/93db64db))
+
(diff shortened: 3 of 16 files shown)


components/esp_modem/generate/src/esp_modem_modules.cpp
```diff
@@ -13,7 +13,15 @@
 namespace esp_modem {
 
 GenericModule::GenericModule(std::shared_ptr<DTE> dte, const dce_config *config) :
-    dte(std::move(dte)), pdp(std::make_unique<PdpContext>(config->apn)) {}
+    dte(std::move(dte)), pdp(std::make_unique<PdpContext>(config->apn ? config->apn : ""))
```
euripedesrocha, line 16: We are dereferencing config without sure of valitidty, is it safe?

euripedesrocha (approved):

components/esp_modem/generate/src/esp_modem_modules.cpp
david-cermak: we cannot do much at this point, this is a ctor (can add an assert in the body but it would be tricky since I'd need to move the constructing pdp into the body too, to run it after the assert)

I'd leave it as is for now, since this is an pre-PR issue.
I'll address it in a separate PR, adding asserts (`ESP_MODEM_THROW_IF_FALSE()`-style in the constructing functions and utils, so it runs before the constructor in the typical modem workflow)

david-cermak: created https://github.com/espressif/esp-protocols/pull/1119 to address this and some other pre-conditions on modem creation

PR #1112, opened 2026-07-26
fix(modem): UartTerminal::on_read usage/set data race

markxoe:
## Description

While testing esp_modem (v2.0.2) with a SIM7080 and ESP32-S3 I reproducibly got several crashes while calling `esp_modem_set_mode`. The stack traces showed either double-free or segfault. Further debugging with `CONFIG_HEAP_POISONING_COMPREHENSIVE` narrowed the problem down to this [`on_read()`](https://github.com/espressif/esp-protocols/blob/62cc0482b2ea077a0e1a25e306c2bd4e64fea258/components/esp_modem/src/esp_modem_uart.cpp#L125) callback being invoked even after it had already been set to `nullptr`.
The problem seems to originate from a data race where [`set_read_cb`](https://github.com/espressif/esp-protocols/blob/62cc0482b2ea077a0e1a25e306c2bd4e64fea258/components/esp_modem/src/esp_modem_uart.cpp#L65) is called between the check before usage of `on_read` and the call of `on_read`.

This PR fixes this issue by locking a mutex when `on_read` is either set or used.

## Related

A quick search found #1014 to be related: this bug is marked as `UART-003` but not fixed in the PR. I found no further mention of this bug.

## Testing

Only manual testing pre- and post-fixing were done due to the bug being complex to test. The bug has not yet occurred after the fix and thorough manual testing.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches concurrency on the UART RX hot path; incorrect locking could deadlock or stall modem I/O, but the change is narrowly scoped and matches existing esp_modem Lock primitives.
> 
> **Overview**
> Fixes crashes during `esp_modem_set_mode` caused by a race: the UART task could invoke `on_read` after another thread cleared it via `set_read_cb`.
> 
> **UartTerminal** now holds a recursive `Lock` (`on_read_lock`) declared before the UART task so it outlives the task. `set_read_cb` takes the lock while swapping the callback; the UART `task` loop takes the same lock around the null-check and invocation of `on_read` on both the `UART_DATA` path and the polling timeout path. Recursion is required because `on_read` may call `set_read_cb` re-entrantly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cebfab0eb16f00b8c5f860cdc055d6caef726fb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_modem/src/esp_modem_uart.cpp
+++ b/components/esp_modem/src/esp_modem_uart.cpp
@@ -13,6 +13,7 @@
 #include "esp_modem_config.h"
 #include "exception_stub.hpp"
 #include "cxx_include/esp_modem_dte.hpp"
+#include "cxx_include/esp_modem_primitives.hpp"
 #include "uart_resource.hpp"
 
 static const char *TAG = "uart_terminal";
@@ -64,6 +65,7 @@ public:
 
     void set_read_cb(std::function<bool(uint8_t *data, size_t len)> f) override
     {
+        Scoped<Lock> lock(on_read_lock);
         on_read = std::move(f);
     }
 
@@ -95,6 +97,10 @@ private:
     QueueHandle_t event_queue;
     uart_resource uart;
     SignalGroup signal;
+    /// Mutex to protect on_read callback from being updated while being called in \c task.
+    /// Must be recursive as the on_read callback may call set_read_cb to change the callback while being called.
+    /// Declared before task_handle so it outlives the UART task.
+    Lock on_read_lock;
     uart_task task_handle;
 };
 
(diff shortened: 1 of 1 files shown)


david-cermak:
Hi @markxoe 

Thank you for the contribution with your clean and minimal patch to modem sources!
Yes, this bug was found and never addressed. We already had a contribution though, fixing this (and some other related bugs `UART-004`, `UART-005`, CMUX teardown, etc...) in https://github.com/espressif/esp-protocols/pull/1083.
Hasn't been merged yet (due to some formal issues + the fact that its merge-base's on `release-v1.4`), but i'd still prefer the 1083 patch version -- tried to quickly cherry-pick onto master in https://github.com/espressif/esp-protocols/pull/1113.

markxoe:
Hi @david-cermak that looks like the better solution, thanks for merging it so quickly!

PR #1110, opened 2026-07-24
fix(mdns): Create static task only when SPIRAM-task-alloc is ON

david-cermak:
Partially addresses https://github.com/espressif/esp-protocols/issues/1100

---

> [!NOTE]
> **Medium Risk**
> Touches core mDNS task lifecycle and compile-time removal of browse paths; misconfiguration could break apps that call browse APIs with browse disabled.
> 
> **Overview**
> Adds **`CONFIG_MDNS_ENABLE_BROWSE`** (default on) so continuous browse (`mdns_browse_new` / related RX, netif, console, and `mdns_browser.c`) can be stripped for ~3–4 KB flash when only advertise or one-shot queries are needed. Public browse APIs and browse CLI commands are wrapped in the same guard; CI builds `sdkconfig.build_only.no_browse` via new **`sdkconfig.build_only.*`** steps in the mdns workflow.
> 
> Separately, the mDNS service task now uses **`xTaskCreatePinnedToCore`** when the stack stays on internal RAM and **`CONFIG_FREERTOS_TASK_CREATE_ALLOW_EXT_MEM`** is off; **`xTaskCreateStaticPinnedToCore`** with caps-allocated stack is kept for SPIRAM task creation or when external task memory is allowed. Stack free on deinit follows that split.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05e7e4f27fc185596e37d40e36688d4a1b76c78d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Issue #1100, opened 2026-07-15
mdns component high static resource usage


projectgus:
### Answers checklist.

- [x] I have read the documentation for esp-protocols [components](https://github.com/espressif/esp-protocols#readme) and the issue is not addressed there.
- [x] I have updated my esp-protocols branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Hi @david-cermak, hope you're doing well!

I don't know if this is an actionable request, but MicroPython has been using the mdns component v1.1.0 for a long time (oops!) Today I went to update it to the latest v1.11.3 to pick up a bug fix.

Our binary size report indicated this update would add 7452 bytes of flash size and just over 2KB more static RAM usage! The RAM usage in particular is a big concern for us on smaller chips.

We're using the same set of mDNS features before and after the update, so I was wondering if there might please be any config options that could be added to this component to disable some resource-heavy features and reclaim some flash size and static RAM?

(Context from MicroPython side: https://github.com/micropython/micropython/pull/19468 - we've temporarily compromised by updating only to v1.3.2 which adds 5000 bytes of binary size but no extra static RAM.)

david-cermak:
Hi Angus,
Nice to hear from you 👋 
Thanks for the report, checking the sizes with IDF tools ( `libmdns.a: esp_idf_size --archives`) I'm getting:

| tag    | .text | .rodata | flash | .data | .bss | DRAM |
|--------|------:|--------:|------:|------:|-----:|-----:|
| 1.1.0  | 26664 |      84 | 26796 |    48 | 1910 | 1958 |
| 1.3.2  | 30899 |      84 | 31031 |    48 | 1910 | 1958 |
| 1.9.0  | 30159 |     182 | 30389 |    48 | 2254 | 2302 |
| 1.10.0 | 31047 |     912 | 32007 |    48 | 2354 | 2402 |
| 1.11.0 | 31494 |     901 | 32443 |    48 | 2354 | 2402 |
| 1.11.3 | 33246 |    1047 | 34341 |    48 | 2354 | 2402 |

Which is in line with your report of flash size increase **+7452**, but don't see the +2K of static ram (just bss **~+456**)
Isn't it possible that the `data` delta in micropython's elf also include some **flash rodata** (not only `.dram0.data`)? (there was a big refactor in `v1.11` splitting the monolithic mdns.c into multiple modules, explaining the increase with debug/error log strings)

[...]

PR #1121, opened 2026-08-07
feat(mdns): Use internal cache for browser update

Siflorite:
## Description

I'm currently implementing OpenThread DNS-SD interfaces for ESP Thread libraries. Those interfaces require a PTR browser and continuous resolvers for PTR, SRV, and ADDR records. Current `mdns` component cannot provide full functions, thus a series of PRs will be proposed to implement necessary contents.

As for browser, the existing browser in `mdns` component provides continuous browser through `mdns_browse_*`, but it lacks these functions to meet OpenThread interface requirements:

- Public interfaces `mdns_browse_new()` and `mdns_browse_delete()` do not support browsers for subtypes.
- The notifier type `mdns_browse_notify_t` only returns a `mdns_result_t`, where `ttl` is the minimum TTL of all records parsed from packet, not specific PTR TTL.

This PR introduces `mdns_cache` to store cached results that match running browses (and future continuous resolvers). If cache is updated, matching browses will generate temporary results from cache and notify. When a browse is deleted, the matching cache entry will be removed.

### Cache structure
Cache is made up of two structures:
```c
/**
 * @brief   mDNS cache service structure, contains PTR (with subtype list), SRV and TXT records
 */
typedef struct mdns_service_cache_s {
    char *instance_name;
    char *service;
    char *proto;
    // PTR
    bool ptr_present;   /*!< true if PTR record is present */
    uint32_t ptr_ttl;
    // Subtype list
    mdns_cache_subtype_t *subtype_list;
    // SRV
    bool srv_present;   /*!< true if SRV record is present */
    uint16_t priority;
    uint16_t weight;
    uint16_t port;
    uint32_t srv_ttl;
    // TXT
    bool txt_present;   /*!< true if TXT record is present */
    mdns_txt_linked_item_t *txt_list;
    uint32_t txt_ttl;
    // Dirty flag
    bool dirty;         /*!< true if the service cache is modified but not yet notified */
    struct mdns_service_cache_s *next;
} mdns_service_cache_t;

/**
 * @brief   mDNS cache entry structure, contains hostname, IP address list and service cache list
 */
typedef struct mdns_cache_entry_s {
    char *hostname;
    esp_netif_t *esp_netif;
    mdns_ip_protocol_t ip_protocol;

    mdns_cache_addr_t *addr_list;
    mdns_service_cache_t *service_cache_list;
    struct mdns_cache_entry_s *next;
} mdns_cache_entry_t;
```

`mdns_cache.c` holds a `static mdns_cache_entry_t * s_cache` as the head of cache entry linked list. Cache entry is identified with hostname, netif, and IP protocol. Every cache entry holds a list of addresses binded to hostname and a list of services. Every service cache entry is identified with instance name, service name, and protocol. A service entry contains all content for PTR, SRV, and TXT record, as well as flags to mark the presence of each record and whether the cache is dirty.

### New routines

Browse registration routine is almost unchanged.

For notify events, the routine is as follow:
- `mdns_parse_packet()` works the same until `if (type == MDNS_TYPE_PTR)`
- If a running browse match the parsed PTR/SRV/TXT/A/AAAA record, call `mdns_priv_cache_update_*()` to update cache and mark the matching cache dirty.
- After all data have been processed, call `mdns_priv_cache_process_dirty()`, this function will call `mdns_priv_browse_update_from_service_cache()` for every dirty service cache.
- Find the matching browse to the dirty cache, generate a temporary result and invoke notifier. The result will be freed after the notifier returns.

When a browse is removed, the routine is as follow:
- If no more browse with identical service name and protocol exists, remove all cache that match the service name and protocol of the browse.
- If other browses with identical service name and protocol exist (only differ in subtype), and the browse to remove has subtype, just remove the subtype from the subtype list of the service cache.

The cache is updated in such a way:

For regular events (TTL>0):
- PTR record: If no matching service cache is found, create a new cache entry with hostname=NULL, and insert the new service under this entry. If subtype is not NULL, just add subtype to subtype list, otherwise update `ptr_present` and ttl.
- SRV record: If no matching service cache is found, create a new cache entry with hostname. If previously a new PTR record with the same service has created a hostname=NULL entry, move the service to this entry. Then update SRV record data.
- TXT record: Create new cache entry or service cache if not found, then update TXT record data.
- A/AAAA record: Create new cache entry if not found, insert address into address list of the cacthe entry or update TTL, then mark all service cache under this entry as dirty.

When a TTL=0 record is sent in, PTR will notify a goodbye event, then remove subtype from subtype list if subtype is not NULL, otherwise set ptr_present = false. SRV and TXT will mark present = false and clear record. A/AAAA will remove the address from address list of the cache entry and mark all its services as dirty. If all records under a service cache are absent, it will be removed; if an entry has no services, the entry will be removed.

All update functions return a `mdns_cache_update_result_t`, `MDNS_CACHE_ADDED` and `MDNS_CACHE_REMOVED` refer to whether a cache entry or a service cache is created or freed. `MDNS_CACHE_UPDATED` and `MDNS_CACHE_NO_CHANGE` refer to whether cache content is changed. `MDNS_CACHE_ERROR` means error occurred while handling cache.

### Changed behaviors

These behaviors are different from previous versions and need careful consideration:

- The `mdns_result_t *result` under `mdns_browse_t` is now removed, as the result of browse is only generated when notifies. As a result, the `next` pointer of a result obtained from notifier will always be NULL.
- Previously `mdns_parse_packet` assume that a packet only contains data of one service, thus only one browse will be notified. Now browses are notified based on dirty cache, therefore multiple browses can be notified, including batch TTL=0 events previously mentioned in `mdns_notifier_t`. Users should no longer traverse the results obtained from browses.
- The `ttl` in `mdns_result_t` used to represent the minimum TTL of all records, which may cause ambiguity. Now `ttl` only presents PTR TTL, it is recommended to use one-shot queries, or continuous resolvers which may be proposed in the future, to obtain TTL of SRV/TXT/A/AAAA records.
- A new member `char *subtype` has been appended to `mdns_result_t`, to distinguish subtype information for browse notifications. This may not cause ABI breaking change as `subtype` is appended to the end and offsets of other members remain unchanged. But all binaries using `sizeof(mdns_result_t)` need to be re-compiled.

## Related

Previously PR #1028 has proposed browse and query with subtypes, but it has not been pushed forward since March 2026. This PR uses some of the code in #1028 for subtype browse, the query part is not used. #1028 or a new PR should be pushed to implement one-shot query with subtypes, and support subtype for `mdns_lookup_selfhosted_service()` and `mdns_lookup_delegated_service()`.

## Testing

This PR has only been tested under mdns browse with one Thread BR and one Thread router, where browse and cache are created and removed in expected routine. A new PR should be made to implement unit tests for mdns_cache and mdns_browse.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Refactors browse data flow and callback lifetime semantics—apps that walked `result->next`, kept notifier pointers, or assumed old TTL/shared-hostname behavior need updates. Core mDNS task locking and memory paths change but stay localized to browse-enabled builds.
> 
> **Overview**
> **Moves continuous mDNS browse off per-browse result lists** and onto a shared internal cache (`mdns_cache.c`) keyed by netif, IP version, hostname, and service instance. Incoming PTR/SRV/TXT/A/AAAA answers update the cache during `mdns_parse_packet()`; **`mdns_priv_cache_process_sync()`** at the end of each packet drives browse notifications.
> 
> **Browse notifier contract changes:** callbacks receive a **temporary** `mdns_result_t` built from the cache (freed when the callback returns), **`result->next` is always NULL**, and registering a browse replays matching cached services. **`ACTION_BROWSE_SYNC`** and the old staged-IP / per-packet single-browse sync path are removed.
> 
> When the last browse for a `_service._proto` is deleted, matching cache entries for that type are dropped if nothing else subscribes. Public docs in `mdns.h` are updated to match; unit test build adds `mdns_cache.c`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18d5421d1d5382ed2166d69e4a8f9a9183252451. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/mdns/mdns_browser.c
+++ b/components/mdns/mdns_browser.c
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <string.h>
-#include "sdkconfig.h"
 #include "mdns_private.h"
 #include "mdns_browser.h"
 #include "mdns_mem_caps.h"
@@ -15,6 +14,13 @@
 #include "mdns_netif.h"
 #include "mdns_service.h"
 #include "esp_log.h"
+#include "mdns_cache.h"
+
+#define MDNS_CACHE_RECORD_BROWSE_MASK \
+    ((mdns_cache_record_mask_t)(MDNS_CACHE_RECORD_PTR | \
+                                MDNS_CACHE_RECORD_SRV | \
+                                MDNS_CACHE_RECORD_TXT | \
+                                MDNS_CACHE_RECORD_ADDR))
 
 static const char *TAG = "mdns_browser";
 
@@ -48,82 +54,13 @@ static esp_err_t send_browse_action(mdns_action_type_t type, mdns_browse_t *brow
  */
 static void browse_item_free(mdns_browse_t *browse)
 {
-    mdns_mem_free(browse->service);
-    mdns_mem_free(browse->proto);
-    if (browse->result) {
-        mdns_priv_query_results_free(browse->result);
-    }
-    mdns_mem_free(browse);
-}
-
-/**
- * @brief Check that a browse is still linked in @c s_browse
- *
- * Sync batches borrow @c browse_sync->browse. ACTION_BROWSE_END may detach and
- * free that browse while a later ACTION_BROWSE_SYNC is still queued, so the
- * sync handler must not touch the pointer without checking.
- */
-static bool browse_is_in_list(const mdns_browse_t *browse)
-{
-    for (const mdns_browse_t *b = s_browse; b != NULL; b = b->next) {
-        if (b == browse) {
-            return true;
-        }
-    }
-    return false;
-}
-
-/**
- * @brief Check that a result node is still linked in the browse cache
- *
- * Sync entries hold a borrowed pointer to a node in @c browse->result. One node
- * can be referenced by several queued sync batches, and the first batch to see
- * it with @c ttl==0 detaches and frees it, so a later batch must not use its
- * pointer without checking.
- */
-static bool result_is_cached(const mdns_browse_t *browse, const mdns_result_t *result)
-{
-    for (const mdns_result_t *r = browse->result; r != NULL; r = r->next) {
-        if (r == result) {
-            return true;
-        }
-    }
-    return false;
-}
-
-/**
- * @brief Deliver browse updates to the user notifier
- *
- * Invokes the notifier once per changed result accumulated for the current
- * packet. The passed @c result pointer is a live node in @c browse->result;
- * @c result->next is not cleared before the callback (only after TTL=0 removal).
- * See @ref mdns_browse_notify_t for how callers should use @c next.
- */
-static void browse_sync(mdns_browse_sync_t *browse_sync)
-{
-    mdns_browse_t *browse = browse_sync->browse;
-    // END may have already detached+freed this browse, or delete marked it off
-    if (!browse_is_in_list(browse) || browse->state != BROWSE_RUNNING) {
+    if (!browse) {
         return;
     }
-    mdns_browse_result_sync_t *sync_result = browse_sync->sync_result;
-    while (sync_result) {
-        mdns_result_t *result = sync_result->result;
-        if (!result_is_cached(browse, result)) {
-            // Already removed and freed by an earlier sync batch
-            sync_result = sync_result->next;
-            continue;
-        }
-        DBG_BROWSE_RESULTS(result, browse_sync->browse);
-        browse->notifier(result);
-        if (result->ttl == 0) {
-            queueDetach(mdns_result_t, browse->result, result);
-            // Just free current result
-            result->next = NULL;
-            mdns_priv_query_results_free(result);
-        }
-        sync_result = sync_result->next;
-    }
+
+    mdns_mem_free(browse->service);
+    mdns_mem_free(browse->proto);
+    mdns_mem_free(browse);
 }
 
 /**
@@ -169,6 +106,29 @@ void mdns_priv_browse_free(void)
     }
 }
 
+static bool names_equal(const char *a, const char *b)
+{
+    return !mdns_utils_str_null_or_empty(a) && !mdns_utils_str_null_or_empty(b) && strcasecmp(a, b) == 0;
+}
+
+/**
+ * @brief Check if a browse matches a given service and proto.
+ */
+static bool browse_matches_identity(const mdns_browse_t *browse, const char *service, const char *proto)
+{
+    return browse && names_equal(browse->service, service) && names_equal(browse->proto, proto);
+}
+
+bool mdns_priv_browse_has_service(const char *service, const char *proto)
+{
+    for (const mdns_browse_t *it = s_browse; it; it = it->next) {
+        if (it->state != BROWSE_OFF && browse_matches_identity(it, service, proto)) {
+            return true;
+        }
+    }
+    return false;
+}
+
 /**
  * @brief  Remove and free the browse from browse linked list
  */
@@ -181,6 +141,7 @@ static void browse_finish(mdns_browse_t *browse)
     for (mdns_browse_t *it = s_browse; it; it = it->next) {
         if (it == browse) {
             queueDetach(mdns_browse_t, s_browse, it);
+            mdns_priv_cache_remove_service_cache_if_unused(it->service, it->proto);
             browse_item_free(it);
             return;
         }
@@ -209,7 +170,6 @@ static mdns_browse_t *browse_init(const char *service, const char *proto, mdns_b
             return NULL;
         }
     }
-
     if (!mdns_utils_str_null_or_empty(proto)) {
         browse->proto = mdns_mem_strndup(proto, MDNS_NAME_BUF_LEN - 1);
         if (!browse->proto) {
@@ -227,14 +187,17 @@ static mdns_browse_t *browse_init(const char *service, const char *proto, mdns_b
  */
 static void browse_start(mdns_browse_t *browse)
 {
+    if (!browse || browse->state != BROWSE_INIT) {
+        return;
+    }
+    browse->state = BROWSE_RUNNING;
+    mdns_priv_cache_notify_browse(browse);
     for (uint8_t interface_idx = 0; interface_idx < MDNS_MAX_INTERFACES; interface_idx++) {
         for (uint8_t protocol_idx = 0; protocol_idx < MDNS_IP_PROTOCOL_MAX; protocol_idx++) {
             browse_send(browse, (mdns_if_t) interface_idx, (mdns_ip_protocol_t) protocol_idx);
         }
     }
 }
-
-static esp_err_t add_browse_result(mdns_browse_sync_t *sync_browse, mdns_result_t *r);
 
 /**
  * @brief  Called from packet parser to find matching running search
@@ -259,139 +222,6 @@ mdns_browse_t *mdns_priv_browse_find_ptr(mdns_name_t *name)
     return NULL;
 }
 
-/**
- * @note Only one browse sync object is kept per parsed packet.  If @p sync
- *       is already allocated for a *different* browse, this function returns
- *       the existing object unchanged — callers that compare
- *       out_sync_browse->browse against the current browse will silently
- *       skip the update.  This is acceptable because mDNS answers for
- *       multiple browsed service types in a single packet are uncommon.
- *       The returned object must still be checked by the caller because NULL
- *       means allocation failed when @p browse was non-NULL.
- */
-mdns_browse_sync_t *mdns_priv_browse_ensure_sync(mdns_browse_t *browse, mdns_browse_sync_t *sync)
-{
-    if (!browse) {
-        return sync;
-    }
-    if (!sync) {
-        sync = (mdns_browse_sync_t *)mdns_mem_malloc(sizeof(mdns_browse_sync_t));
-        if (!sync) {
-            HOOK_MALLOC_FAILED;
-            return NULL;
-        }
-        sync->browse = browse;
-        sync->sync_result = NULL;
-    }
-    return sync;
-}
-
-void mdns_priv_browse_staged_ip_free(mdns_browse_staged_ip_t *staged)
-{
-    while (staged) {
-        mdns_browse_staged_ip_t *next = staged->next;
-        mdns_mem_free(staged);
-        staged = next;
-    }
-}
-
-esp_err_t mdns_priv_browse_stage_ip(mdns_browse_staged_ip_t **staged, const char *hostname, esp_ip_addr_t *ip,
-                                    mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl)
-{
-    mdns_browse_staged_ip_t *item = (mdns_browse_staged_ip_t *)mdns_mem_malloc(sizeof(mdns_browse_staged_ip_t));
-    if (!item) {
-        HOOK_MALLOC_FAILED;
-        return ESP_ERR_NO_MEM;
-    }
-    memset(item, 0, sizeof(mdns_browse_staged_ip_t));
-    strncpy(item->hostname, hostname, MDNS_NAME_BUF_LEN - 1);
-    item->hostname[MDNS_NAME_BUF_LEN - 1] = '\0';
-    item->ip = *ip;
-    item->tcpip_if = tcpip_if;
-    item->ip_protocol = ip_protocol;
-    item->ttl = ttl;
-    item->next = *staged;
-    *staged = item;
-    return ESP_OK;
-}
-
-/**
- * @brief Apply packet-staged A/AAAA records after SRV hostnames are known
- *
- * @note Each staged address is applied via mdns_priv_browse_result_add_ip(), which
- *       attaches the address only to the first browse result with a matching
- *       hostname on the same interface and IP protocol. Additional instances that
- *       share the same target host do not receive a copy automatically.
- */
-void mdns_priv_browse_apply_staged_ips(mdns_browse_t *browse, mdns_browse_staged_ip_t *staged,
-                                       mdns_browse_sync_t *out_sync_browse)
-{
-    if (!browse || !staged || !out_sync_browse || out_sync_browse->browse != browse) {
-        return;
-    }
-    while (staged) {
-        mdns_priv_browse_result_add_ip(browse, staged->hostname, &staged->ip, staged->tcpip_if,
-                                       staged->ip_protocol, staged->ttl, out_sync_browse);
-        staged = staged->next;
-    }
-}
-
-void mdns_priv_browse_result_add_ptr(mdns_browse_t *browse, const char *instance, const char *service, const char *proto,
-                                     mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl,
-                                     mdns_browse_sync_t *out_sync_browse)
-{
-    if (!browse || !out_sync_browse || out_sync_browse->browse != browse
-            || mdns_utils_str_null_or_empty(instance) || mdns_utils_str_null_or_empty(service)
-            || mdns_utils_str_null_or_empty(proto)) {
-        return;
-    }
-    mdns_result_t *r = browse->result;
-    while (r) {
-        if (r->esp_netif == mdns_priv_get_esp_netif(tcpip_if) && r->ip_protocol == ip_protocol &&
-                !mdns_utils_str_null_or_empty(r->instance_name) && !strcasecmp(instance, r->instance_name) &&
-                !mdns_utils_str_null_or_empty(r->service_type) && !strcasecmp(service, r->service_type) &&
-                !mdns_utils_str_null_or_empty(r->proto) && !strcasecmp(proto, r->proto)) {
-            if (r->ttl != ttl) {
-                uint32_t previous_ttl = r->ttl;
-                if (r->ttl == 0) {
-                    r->ttl = ttl;
-                } else {
-                    mdns_priv_query_update_result_ttl(r, ttl);
-                }
-                if (previous_ttl != r->ttl) {
-                    add_browse_result(out_sync_browse, r);
-                }
-            }
-            return;
-        }
-        r = r->next;
-    }
-
-    r = (mdns_result_t *)mdns_mem_malloc(sizeof(mdns_result_t));
-    if (!r) {
-        HOOK_MALLOC_FAILED;
-        return;
-    }
-    memset(r, 0, sizeof(mdns_result_t));
-    r->instance_name = mdns_mem_strdup(instance);
-    r->service_type = mdns_mem_strdup(service);
-    r->proto = mdns_mem_strdup(proto);
-    if (!r->instance_name || !r->service_type || !r->proto) {
-        HOOK_MALLOC_FAILED;
-        mdns_mem_free(r->instance_name);
-        mdns_mem_free(r->service_type);
-        mdns_mem_free(r->proto);
-        mdns_mem_free(r);
-        return;
-    }
-    r->esp_netif = mdns_priv_get_esp_netif(tcpip_if);
-    r->ip_protocol = ip_protocol;
-    r->ttl = ttl;
-    r->next = browse->result;
-    browse->result = r;
-    add_browse_result(out_sync_browse, r);
-}
-
 mdns_browse_t *mdns_priv_browse_find(mdns_name_t *name, uint16_t type, mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol)
 {
     mdns_browse_t *b = s_browse;
@@ -399,7 +229,7 @@ mdns_browse_t *mdns_priv_browse_find(mdns_name_t *name, uint16_t type, mdns_if_t
     if (type != MDNS_TYPE_SRV && type != MDNS_TYPE_A && type != MDNS_TYPE_AAAA && type != MDNS_TYPE_TXT) {
         return NULL;
     }
-    mdns_result_t *r = NULL;
+
     while (b) {
         if (b->state != BROWSE_RUNNING) {
             b = b->next;
@@ -414,38 +244,15 @@ mdns_browse_t *mdns_priv_browse_find(mdns_name_t *name, uint16_t type, mdns_if_t
             }
             return b;
         } else if (type == MDNS_TYPE_A || type == MDNS_TYPE_AAAA) {
-            r = b->result;
-            while (r) {
-                if (r->esp_netif == mdns_priv_get_esp_netif(tcpip_if) && r->ip_protocol == ip_protocol && !mdns_utils_str_null_or_empty(r->hostname) && !strcasecmp(name->host, r->hostname)) {
-                    return b;
-                }
-                r = r->next;
+            if (mdns_priv_cache_host_has_service(name->host, mdns_priv_get_esp_netif(tcpip_if),
+                                                 ip_protocol, b->service, b->proto)) {
+                return b;
             }
             b = b->next;
             continue;
         }
     }
-    return b;
-}
-
-static void sync_browse_result_link_free(mdns_browse_sync_t *browse_sync)
-{
-    mdns_browse_result_sync_t *current = browse_sync->sync_result;
-    mdns_browse_result_sync_t *need_free;
-    while (current) {
-        need_free = current;
-        current = current->next;
-        mdns_mem_free(need_free);
-    }
-    mdns_mem_free(browse_sync);
-}
-
-void mdns_priv_browse_sync_free(mdns_browse_sync_t *browse_sync)
-{
-    if (!browse_sync) {
-        return;
-    }
-    sync_browse_result_link_free(browse_sync);
+    return NULL;
 }
 
 void mdns_priv_browse_action(mdns_action_t *action, mdns_action_subtype_t type)
@@ -454,10 +261,6 @@ void mdns_priv_browse_action(mdns_action_t *action, mdns_action_subtype_t type)
         switch (action->type) {
         case ACTION_BROWSE_START:
             browse_start(action->data.browse_add.browse);
-            break;
-        case ACTION_BROWSE_SYNC:
-            browse_sync(action->data.browse_sync.browse_sync);
-            sync_browse_result_link_free(action->data.browse_sync.browse_sync);
             break;
         case ACTION_BROWSE_END:
             browse_finish(action->data.browse_add.browse);
@@ -474,9 +277,6 @@ void mdns_priv_browse_action(mdns_action_t *action, mdns_action_subtype_t type)
             // Browse actions do not own the browse item.
             // If a queued action is discarded during shutdown, the linked browse item is released by mdns_priv_browse_free().
             break;
-        case ACTION_BROWSE_SYNC:
-            sync_browse_result_link_free(action->data.browse_sync.browse_sync);
-            break;
         default:
             abort();
         }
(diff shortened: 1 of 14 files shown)


components/mdns/private_include/mdns_private.h
```diff
@@ -409,3 +411,73 @@ typedef struct {
...
+    uint16_t priority;
+    uint16_t weight;
+    uint16_t port;
+    uint32_t srv_ttl;
+    // TXT
+    bool txt_present;   /*!< true if TXT record is present */
+    mdns_txt_linked_item_t *txt_list;
+    uint32_t txt_ttl;
+    // Dirty flag
+    bool dirty;         /*!< true if the service cache is modified but not yet notified */
```
zwx1995esp: Use a bitmask for the sync-out flag, as we may need to support non-blocking resolvers for both TXT and SRV records independently.
```suggestion
    uint8_t sync_out;         /*!< true if the service cache is modified and synchronization is required. */
```

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,198 @@
...
+ */
+mdns_result_t *mdns_priv_service_cache_to_result(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service);
+
+/**
+ * @brief Process all dirty cache entries.
+ *
+ * @note This function is called by the MDNS task to process all dirty cache entries.
+ *       It will generate temporary results, notify corresponding browsers, and clear the dirty flag.
+ */
+void mdns_priv_cache_process_dirty(void);
```
zwx1995esp:
```suggestion
void mdns_priv_cache_process_sync(void);
```

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,198 @@
...
+ * @param entry       Pointer to the cache entry.
+ * @param service     Pointer to the service cache.
+ * @return The converted mdns_result_t.
+ *
+ * @note The ownership of the returned `mdns_result_t` is transferred to the caller.
+ */
+mdns_result_t *mdns_priv_service_cache_to_result(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service);
+
+/**
+ * @brief Process all dirty cache entries.
```
zwx1995esp: Process all pending synchronization cache entries.

components/mdns/private_include/mdns_private.h
Siflorite: Now uses two flags to mark records and consumers to synchronize, currently browser only cares `sync_consumers & MDNS_CACHE_CONSUMER_BROWSE`:
```C
typedef enum {
    MDNS_CACHE_CONSUMER_BROWSE     = (1U << 0),
    MDNS_CACHE_CONSUMER_RESOLVER   = (1U << 1)
} mdns_cache_consumer_type_t;

typedef enum {
    MDNS_CACHE_RECORD_PTR       = (1U << 0),
    MDNS_CACHE_RECORD_SUBTYPE   = (1U << 1),
    MDNS_CACHE_RECORD_SRV       = (1U << 2),
    MDNS_CACHE_RECORD_TXT       = (1U << 3),
    MDNS_CACHE_RECORD_ADDR      = (1U << 4)
} mdns_cache_record_type_t;

typedef struct mdns_service_cache_s {
    // ...
    // Dirty flags
    uint8_t sync_records;   /*!< bitmask of records to sync, see @ref mdns_cache_record_type_t */
    uint8_t sync_consumers; /*!< bitmask of consumers to sync, see @ref mdns_cache_consumer_type_t */
    struct mdns_service_cache_s *next;
} mdns_service_cache_t;
```

components/mdns/private_include/mdns_cache.h
Siflorite: Changed function name and annotation
```C
/**
 * @brief Process all pending synchronization cache entries.
 *
 * @note This function is called by the MDNS task to process all pending synchronization cache entries.
 *       It will generate temporary results, notify corresponding consumers, and clear the sync out flags.
 */
void mdns_priv_cache_process_sync(void);
```

Siflorite:
I have modified the commit a bit. Some names and annotations changed based on reviews of @zwx1995esp and some new types are added to support future resolvers.

components/mdns/private_include/mdns_browser.h
```diff
@@ -104,41 +78,44 @@ void mdns_priv_browse_send_by_ip_protocol(mdns_if_t mdns_if, mdns_ip_protocol_t
...
-                                    mdns_if_t tcpip_if, mdns_ip_protocol_t ip_protocol, uint32_t ttl, mdns_browse_sync_t *out_sync_browse);
+bool mdns_priv_browse_notify_from_service_cache(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service,
+                                                mdns_browse_t *browse);
+
 /**
- * @brief  Add a SRV record to the browse result
+ * @brief Notify the affected normal/subtype browse about a PTR goodbye.
  *
- * @note Called from the packet parser (mdns_receive.c)
+ * @param subtype NULL: normal PTR goodbye; non-NULL: subtype PTR goodbye.
```
zwx1995esp:
```suggestion
 * @param subtype The subtype for the PTR goodbye record, can be NULL if no subtype in the PTR.
```

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+    }
+
+    if (!service_entry) {
+        service_entry = cache_add_service(host_entry, instance, service, proto);
+        if (!service_entry) {
+            cache_remove_entry_if_empty(host_entry);
+            return MDNS_CACHE_ERROR;
+        }
+
+        owner_entry = host_entry;
```
zwx1995esp: Never used the variable `owner_entry` after here in this function, could we remove this line?

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+        new_entry = true;
+    } else if (owner_entry != host_entry) {
+        if (!cache_move_service(owner_entry, host_entry, service_entry)) {
+            return MDNS_CACHE_ERROR;
+        }
+        owner_entry = host_entry;
+        moved = true;
+    }
+
+    result = service_cache_srv_update(service_entry, priority, weight, port, ttl);
```
zwx1995esp: Since the local variable`new_entry` and `moved` only be used to determine what the result value is. How about we define a temp variable to record the result of this function and just mark the result in the if branch above. And if the `result` is `MDNS_CACHE_NO_CHANGE` we change the `result` using the temp variable. Something like:
```
if (!service_entry) {
    ...
    result = MDNS_CACHE_ADDED;
} else if (owner_entry != host_entry) {
    ...
    result = MDNS_CACHE_UPDATED;
}

mdns_cache_update_result_t srv_result = service_cache_srv_update(service_entry, priority, weight, port, ttl);

if (result == MDNS_CACHE_NO_CHANGE) {
    result = srv_result;
}
```
Then we can remove these two local variables.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+{
+    if (!a || !b || !names_equal(a->key, b->key)) {
+        return false;
+    }
+    if (a->value_len != b->value_len) {
+        return false;
+    }
+    if (a->value_len == 0) {
+        return true;
+    }
```
zwx1995esp: Can be removed, since the function `memcmp` can process the scenario when the value of the length is `0`.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+        return false;
+    }
+
+    for (const mdns_txt_linked_item_t *it = a; it; it = it->next) {
+        if (!txt_list_contains(b, it)) {
+            return false;
+        }
+    }
+
+    for (const mdns_txt_linked_item_t *it = b; it; it = it->next) {
```
zwx1995esp: Since the length of the link list a is the same as the length of the link list b, then no need to check this again.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+    while (txt_list) {
+        count++;
+        txt_list = txt_list->next;
+    }
+    return count;
+}
+
+static bool txt_list_equal(const mdns_txt_linked_item_t *a, const mdns_txt_linked_item_t *b)
+{
+    if (a == b) {
```
zwx1995esp, line 605: I think we need also check the whole link list even the header pointer of the two link lists are the same.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+#ifdef CONFIG_LWIP_IPV4
+    if (a->type == ESP_IPADDR_TYPE_V4) {
+        return a->u_addr.ip4.addr == b->u_addr.ip4.addr;
+    }
+#endif
+
+    return false;
+}
+
+static bool service_cache_is_empty(const mdns_service_cache_t *service)
```
zwx1995esp, line 68: The pointer `service` is not checked null in the function. If we expect the user to check this variable before calling this function, we need to add some brief here, or we should check this pointer in the function.

components/mdns/mdns_browser.c
```diff
@@ -487,369 +414,266 @@ static esp_err_t add_browse_result(mdns_browse_sync_t *sync_browse, mdns_result_
...
+ */
+static uint32_t get_ptr_ttl(const mdns_service_cache_t *service, const char *subtype)
+{
+    if (mdns_utils_str_null_or_empty(subtype)) {
+        return service->ptr_ttl;
+    }
+
+    for (const mdns_cache_subtype_t *it = service->subtype_list; it; it = it->next) {
+        if (!mdns_utils_str_null_or_empty(it->subtype) && !strcasecmp(it->subtype, subtype)) {
+            return it->ttl;
```
Siflorite: There is new modification since last commit: now the ttl that browse notifies provide PTR TTL, but if the user only browse subtype, `ptr_present` will be false and `ptr_ttl` will be 0 in cache. This function provides the TTL of specific subtype or normal PTR, and in `browse_build_and_sync_temp_result()` this TTL will be applied to `result->ttl`.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+        if (service_cache_is_empty(service_entry)) {
+            return cache_remove_service(owner_entry, service_entry) ? MDNS_CACHE_REMOVED : MDNS_CACHE_NO_CHANGE;
+        }
+
+        service_cache_mark_sync_out(service_entry, MDNS_CACHE_UPDATED, MDNS_CACHE_RECORD_SRV, MDNS_CACHE_CONSUMER_BROWSE);
+
+        return MDNS_CACHE_UPDATED;
+    }
+
+    host_entry = cache_get_or_add_entry(hostname, esp_netif, ip_protocol);
```
zwx1995esp: Please check if we have already find an entry before we re-search the cache.

```suggestion
    if (service_entry && names_equal(owner_entry->hostname, hostname)) {
        host_entry = owner_entry;
    } else {
        host_entry = cache_get_or_add_entry(hostname, esp_netif, ip_protocol);
            if (!host_entry) {
                return MDNS_CACHE_ERROR;
            }
    }
```

zwx1995esp (commented):
There are a few minor nits. Since this PR is fairly large, I haven’t finished reviewing it yet.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+            mdns_service_cache_t *cache = *service_ptr;
+            if (names_equal(cache->service, service) && names_equal(cache->proto, proto)) {
+                *service_ptr = cache->next;
+                service_entry_free(cache);
+                continue;
+            }
+            service_ptr = &(*service_ptr)->next;
+        }
+
+        if (!entry->service_cache_list) {
```
zwx1995esp, line 405: We will check if the address link is empty or not before we remove the entry in the function `static bool cache_remove_entry_if_empty(mdns_cache_entry_t *entry)` at the line 216 in this file, but here we never check the address list before we remove the cache entry.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+        // Remove addr
+        mdns_cache_addr_t **addr_ptr = &entry->addr_list;
+        while (*addr_ptr) {
+            if (addr_equal(&(*addr_ptr)->addr, addr)) {
+                mdns_cache_addr_t *removed_addr = *addr_ptr;
+                *addr_ptr = removed_addr->next;
+                mdns_mem_free(removed_addr);
+
+                for (mdns_service_cache_t *service = entry->service_cache_list; service; service = service->next) {
+                    service_cache_mark_sync_out(service, MDNS_CACHE_UPDATED, MDNS_CACHE_RECORD_ADDR, MDNS_CACHE_CONSUMER_RESOLVER);
```
zwx1995esp: Should we mark `MDNS_CACHE_CONSUMER_BROWSE` here? Since we did not support the resolver so far.

Siflorite:
Since this PR is quite large and need careful review, I'll take it apart into independent commits until it is ready to merge, so that it will be easier to observe what I've edited. Then I will squash them into a single commit. Also I'll make cache only complied if browse/resolver is activated, see #1100 and #1110.

components/mdns/private_include/mdns_browser.h
Siflorite: I have changed this line to:
```
 * @param The subtype for PTR goodbye record, can be NULL if PTR record has no subtype.
```
The call chain: `mdns_parse_packet()` in `mdns_receive.c` -> `mdns_priv_cache_update_ptr()` -> If record TTL=0, call `mdns_priv_browse_notify_ptr_goodbye_from_service_cache()`. Then this function finds the corresponding `mdns_browse_t` item and notify a TTL=0 result. The parameters for `mdns_priv_cache_update_ptr()` come from 
```c
content = mdns_utils_parse_fqdn(data, content, name, len);       // Previously Line 847
mdns_browse_t *browse_for_ptr = mdns_priv_browse_find_ptr(name); // Previously Line 933
```
So NULL in `subtype` param shall mean **no subtype in browse item** or **no subtype in PTR record**? I prefer the second one.

components/mdns/mdns_cache.c
Siflorite: That's a good idea.

components/mdns/mdns_cache.c
Siflorite: Removed the value assignment in L613 and L619 since `owner_entry` no longer used.

components/mdns/mdns_cache.c
Siflorite: Since all `mdns_txt_linked_item_t` data came from `mdns_parse_packet()`, I've ensured that if `value_len` is 0, `value` pointer will be NULL. So we should check that either both `value` pointers are NULL or both are non-NULL:
```c
static bool txt_item_equal(const mdns_txt_linked_item_t *a, const mdns_txt_linked_item_t *b)
{
    if (!a || !b || !names_equal(a->key, b->key)) {
        return false;
    }
    if (a->value_len != b->value_len) {
        return false;
    }
    if ((!a->value && b->value) || (a->value && !b->value)) {
        return false;
    }
    return memcmp(a->value, b->value, a->value_len) == 0;
}
```

components/mdns/mdns_cache.c
Siflorite: Deleted duplicated loop.

components/mdns/mdns_cache.c
Siflorite: I'll check service pointer for future maintenance.

components/mdns/mdns_cache.c
Siflorite: Yes, for the sake of future ADDR resolvers, if a user only creates ADDR resolvers, only `addr_list` will be updated and the `service_cache_list` will always be NULL.

Now changed to `if (!entry->service_cache_list && !entry->addr_list)`.

components/mdns/mdns_cache.c
Siflorite: Ah that's a mistake here, since future ADDR resolvers don't need to look into services for ADDR update.

components/mdns/mdns_cache.c
Siflorite: This fix diminishes a duplicated search, because netif and ip_protocol are checked in `mdns_priv_cache_find_service()` if a service entry is found. But I think this will be better:
```c
if (owner_entry && names_equal(owner_entry->hostname, hostname))
```
Because `owner_entry` will also be not NULL if `service_entry` is found, and this will make the code more readable.

components/mdns/mdns_cache.c
zwx1995esp: So, could we add a comment before this function to remind other developers to check the service pointer?

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1039 @@
...
+    mdns_mem_free(service_entry->instance_name);
+    mdns_mem_free(service_entry->service);
+    mdns_mem_free(service_entry->proto);
+    while (service_entry->subtype_list) {
+        mdns_cache_subtype_t *subtype = service_entry->subtype_list;
+        service_entry->subtype_list = subtype->next;
+        mdns_mem_free((char *)subtype->subtype);
+        mdns_mem_free(subtype);
+    }
+    while (service_entry->txt_list) {
```
zwx1995esp, line 203: Could we use `free_txt_linked_list`?

components/mdns/mdns_cache.c
Siflorite: Both linked lists will not be changed in this function, this pointer check will reduce one check in traverses.

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * @brief Find a cache entry by hostname, esp_netif, and ip_protocol.
+ *
+ * @param hostname      The hostname to find.
+ * @param esp_netif     Pointer to the esp_netif to find.
+ * @param ip_protocol   IP protocol to find.
+ *
+ * @return Pointer to the cache entry if found, NULL otherwise.
+ */
+mdns_cache_entry_t *mdns_priv_cache_find_entry(const char *hostname, const esp_netif_t *esp_netif,
```
zwx1995esp: I didn’t find any callers in `mdns_browse.c`; it appears to be used only in `mdns_cache.c`. Could we make this function static and keep it local to `mdns_cache.c`?

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,206 @@
...
+ * @param esp_netif     Pointer to the esp_netif to find.
+ * @param ip_protocol   IP protocol to find.
+ * @param instance      The instance name to find.
+ * @param service       The service name to find.
+ * @param proto         The protocol to find.
+ * @param owner_entry   Receives the pointer to the cache entry that owns the service cache.
+ *
+ * @return Pointer to the service cache entry if found, NULL otherwise.
+ */
+mdns_service_cache_t *mdns_priv_cache_find_service(const esp_netif_t *esp_netif, mdns_ip_protocol_t ip_protocol,
```
zwx1995esp: Same for this function.

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,206 @@
...
+bool mdns_priv_host_has_service(const char *hostname, const esp_netif_t *esp_netif, mdns_ip_protocol_t ip_protocol,
+                                const char *service, const char *proto);
+
+/**
+ * @brief Remove all service caches by service and proto.
+ *
+ * @param service       The service name to remove.
+ * @param proto         The protocol to remove.
+ */
+void mdns_priv_remove_service_caches(const char *service, const char *proto);
```
zwx1995esp: Also this function.

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @param hostname      The hostname to check.
+ * @param esp_netif     Pointer to the esp_netif to check.
+ * @param ip_protocol   IP protocol to check.
+ * @param service       The service name to check.
+ * @param proto         The protocol to check.
+ *
+ * @return true if the host has the service, false otherwise.
+ */
+bool mdns_priv_host_has_service(const char *hostname, const esp_netif_t *esp_netif, mdns_ip_protocol_t ip_protocol,
```
zwx1995esp:
```suggestion
bool mdns_priv_cache_host_has_service(const char *hostname, const esp_netif_t *esp_netif, mdns_ip_protocol_t ip_protocol,
```

components/mdns/private_include/mdns_cache.h
```diff
@@ -0,0 +1,206 @@
...
+void mdns_priv_remove_service_caches(const char *service, const char *proto);
+
+/**
+ * @brief Remove a subtype from a service cache `_service._proto`.
+ *
+ * @param service       The service name to remove the subtype from.
+ * @param proto         The protocol to remove the subtype from.
+ * @param subtype       The subtype to remove.
+ */
+void mdns_priv_service_cache_remove_subtype(const char *service, const char *proto, const char *subtype);
```
zwx1995esp:
```suggestion
void mdns_priv_cache_service_remove_subtype(const char *service, const char *proto, const char *subtype);
```

components/mdns/mdns_receive.c
```diff
@@ -315,6 +316,90 @@ handle_error :
     mdns_mem_free(txt);
 }
 
+static void free_txt_linked_list(mdns_txt_linked_item_t *txt)
```
zwx1995esp: We have already define a same function in the `mdns_cache.c`. Could we move this into a shared utility function in `mdns_utils.c`?

components/mdns/mdns_receive.c
```diff
@@ -315,6 +316,90 @@ handle_error :
...
+    while (txt) {
+        mdns_txt_linked_item_t *next = txt->next;
+        mdns_mem_free((char *)txt->key);
+        mdns_mem_free(txt->value);
+        mdns_mem_free(txt);
+        txt = next;
+    }
+}
+
+static bool result_txt_linked_list_create(const uint8_t *data, size_t len, mdns_txt_linked_item_t **out_txt)
```
zwx1995esp, line 354: There is already a `result_txt_create` function. What is the difference between the two? Could we combine them? Otherwise, any future logic changes would need to be made in both places, which makes the code harder to maintain.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1029 @@
...
+        new_addr->addr = addr->addr;
+        *tail = new_addr;
+        tail = &new_addr->next;
+    }
+
+    *out_addr_list = head;
+    return true;
+}
+
+mdns_result_t *mdns_priv_service_cache_to_result(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service)
```
zwx1995esp: We use ESP_GOTO_ON_FALSE, but never use the error code. How about declaring this function as `esp_err_t mdns_priv_service_cache_to_result(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service, mdns_result_t *out_result)`?

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1029 @@
...
+void mdns_priv_cache_clear(void)
+{
+    while (s_cache) {
+        mdns_cache_entry_t *entry = s_cache;
+        s_cache = s_cache->next;
+        cache_entry_free(entry);
+    }
+}
+
+static bool project_txt(const mdns_txt_linked_item_t *txt_list, mdns_txt_item_t **out_txt, uint8_t **out_value_len,
```
zwx1995esp, line 796: Here we have the third txt generate function.

components/mdns/mdns_browser.c
```diff
@@ -487,369 +414,266 @@ static esp_err_t add_browse_result(mdns_browse_sync_t *sync_browse, mdns_result_
...
-    add_browse_result(out_sync_browse, r);
+
+    return true;
 }
 
-/**
- * @brief  Browse sync result
- */
-esp_err_t mdns_priv_browse_sync(mdns_browse_sync_t *browse_sync)
+bool mdns_priv_browse_update_from_service_cache(const mdns_cache_entry_t *entry, const mdns_service_cache_t *service)
```
zwx1995esp: I use AI to print the call chain for this function:
```
mdns_browse_new()
    └─ mdns_browse_new_with_subtype()
         ├─ browse_init()
         ├─ send_browse_action(ACTION_BROWSE_START)
         ├─ Add the browse to s_browse and mark it as BROWSE_RUNNING
         └─ The mDNS service task dequeues ACTION_BROWSE_START
              └─ mdns_priv_browse_action(action, ACTION_RUN)
                   └─ browse_start(browse)
                        └─ mdns_priv_cache_notify_browse(browse)
                             └─ Iterate over all cache entries and service caches
                                  └─ mdns_priv_browse_notify_from_service_cache(
                                         entry, service, browse)
                                       ├─ Check whether service/proto/subtype match
                                       └─ browse_build_and_sync_temp_result(...)
                                            ├─ mdns_priv_service_cache_to_result()
                                            ├─ mdns_priv_browse_ensure_sync()
                                            ├─ add_browse_result()
                                            └─ mdns_priv_browse_sync()
                                                 └─ Queue ACTION_BROWSE_SYNC
                                                      └─ The mDNS service task executes the action
                                                           └─ browse_sync()
                                                                └─ browse->notifier(result)
```
And found that f there are 10 matching entries in the cache, 10 ACTION_BROWSE_SYNC operations will be queued. I suggest changing this so that each browse replay creates only one sync operation.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1029 @@
...
+    if (!a || !b || !names_equal(a->key, b->key)) {
+        return false;
+    }
+    if (a->value_len != b->value_len) {
+        return false;
+    }
+    if ((!a->value && b->value) || (a->value && !b->value)) {
+        return false;
+    }
+    return memcmp(a->value, b->value, a->value_len) == 0;
```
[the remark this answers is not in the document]
Siflorite: fixed

components/mdns/private_include/mdns_cache.h
Siflorite: OK, perhaps this function was used by browsers when in development.

components/mdns/private_include/mdns_cache.h
Siflorite: I think this was created before `mdns_priv_cache_remove_service_cache_if_unused()`, anyway this should be made static now.

components/mdns/mdns_receive.c
Siflorite: Sounds good.

components/mdns/mdns_cache.c
Siflorite: That's the same problem as what you've mentioned above. Cache stores `mdns_txt_linked_item_t` for convenience but `mdns_result_t` that notifier returns require
```c
    // TXT
    mdns_txt_item_t *txt;                   /*!< txt record */
    uint8_t *txt_value_len;                 /*!< array of txt value len of each record */
    size_t txt_count;                       /*!< number of txt items */
```
So I used `project_txt` to transform `mdns_txt_linked_item_t` into TXT arrays, value length arrays and TXT items count.

components/mdns/mdns_browser.c
Siflorite: That's a severe issue. The point is that this sync process was previously used to ensure only one browse is updated on receiving one rx packet. But it is useless now since cache system ensures all matching browses can be notified. So I think a solution is to call notifier immediately in `browse_build_and_sync_temp_result()`, there is no need to queue up notifications.

components/mdns/mdns_receive.c
Siflorite: That's a compromise for compatibility. Old `mdns_result_t` uses a `mdns_txt_item_t` array and a `uint8_t *txt_value_len` to hold TXT records with value length. The newly added `mdns_cache` uses `mdns_txt_linked_item_t` to hold TXT linked list.

In `mdns_receive.c`, `result_txt_create()` is used for creating TXT content for one-shot queries, which returns a `mdns_result_t *` on notification. But I need to create TXT linked list to update TXT record in cache, so I created `result_txt_linked_list_create()`.

These two functions can share the same content generation logic, only differ in allocation process.

Solution: Share a common `parse_txt_item_data()` core to parse single txt item.

components/mdns/mdns_cache.c
Siflorite: That's quite fair.

components/mdns/mdns_browser.c
Siflorite: I have pushed a new commit "feat(mdns): notify browse results directly from cache", where original `mdns_browse_sync*` related stuff in `mdns_browse` have been deleted, now notify process is no longer queued up, `browse_build_and_notify_temp_result()` will notify immediately. Please recheck this commit.

Siflorite:
The PR is now ready for review, and it requires #1110 and #1125 to be merged first, so as to compile and use cache when `CONFIG_MDNS_ENABLE_BROWSE=y`, and eliminate `mdns_browse_new*` bugs. Then I will rebase this PR and squash into one commit.
A future PR will propose new unit tests for cache and browse.

components/mdns/mdns_browser.c
```diff
@@ -144,13 +155,21 @@ static void browse_finish(mdns_browse_t *browse)
             b = b->next;
         }
     }
+
+    if (removed) {
+        if (!mdns_utils_str_null_or_empty(browse->subtype)) {
+            mdns_priv_cache_service_remove_subtype(browse->service, browse->proto, browse->subtype);
+        }
+        mdns_priv_cache_remove_service_cache_if_unused(browse->service, browse->proto);
+    }
```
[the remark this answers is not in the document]
Siflorite: fixed

components/mdns/mdns_receive.c
```diff
@@ -1121,12 +1187,8 @@ static void mdns_parse_packet(mdns_rx_packet_t *packet)
...
                 memcpy(ip6.u_addr.ip6.addr, data_ptr, MDNS_ANSWER_AAAA_SIZE);
                 if (packet_browse || browse_result) {
-                    mdns_browse_t *browse_ip = browse_result ? browse_result : packet_browse;
-                    out_sync_browse = mdns_priv_browse_ensure_sync(browse_ip, out_sync_browse);
-                    if (out_sync_browse && mdns_priv_browse_stage_ip(&staged_browse_ips, name->host, &ip6,
-                                                                     packet->tcpip_if, packet->ip_protocol, ttl) != ESP_OK) {
-                        goto clear_rx_packet;
-                    }
+                    (void)mdns_priv_cache_update_addr(mdns_priv_get_esp_netif(packet->tcpip_if), packet->ip_protocol,
+                                                      name->host, &ip6, ttl);
```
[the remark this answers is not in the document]
Siflorite: I need to look into this problem more deeply.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1047 @@
...
+
+        if (service_cache_is_empty(service_entry)) {
+            return cache_remove_service(owner_entry, service_entry) ? MDNS_CACHE_REMOVED : MDNS_CACHE_NO_CHANGE;
+        }
+
+        service_cache_mark_sync_out(service_entry, MDNS_CACHE_UPDATED,
+                                    mdns_utils_str_null_or_empty(subtype) ? MDNS_CACHE_RECORD_PTR : MDNS_CACHE_RECORD_SUBTYPE,
+                                    MDNS_CACHE_CONSUMER_BROWSE);
+
+        return MDNS_CACHE_UPDATED;
```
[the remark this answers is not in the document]
Siflorite: removed mark sync for services in PTR goodbye branch

components/mdns/mdns_receive.c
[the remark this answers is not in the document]
Siflorite: Now created a path like former staged_ips

components/mdns/mdns_browser.c
```diff
@@ -125,32 +102,75 @@ void mdns_priv_browse_free(void)
...
+
```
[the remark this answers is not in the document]
Siflorite: fixed

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1057 @@
...
+            cache_remove_entry_if_empty(owner_entry);
+            return MDNS_CACHE_ERROR;
+        }
+        new_service = true;
+    }
+
+    if (!mdns_utils_str_null_or_empty(subtype)) {
+        result = service_cache_add_subtype(service_entry, subtype, ttl);
+        if (result == MDNS_CACHE_ERROR) {
+            return MDNS_CACHE_ERROR;
```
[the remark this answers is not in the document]
Siflorite: now in `if (result == MDNS_CACHE_ERROR)`, service will be removed if `new_service` is true.

components/mdns/mdns_receive.c
```diff
@@ -1251,38 +1355,15 @@ static void mdns_parse_packet(mdns_rx_packet_t *packet)
...
-            mdns_mem_free(out_sync_browse);
-        }
-        out_sync_browse = NULL;
-    }
 
 clear_rx_packet:
-    mdns_priv_browse_staged_ip_free(staged_browse_ips);
-    staged_browse_ips = NULL;
+    mdns_priv_cache_process_sync();
+    rx_staged_ip_free(staged_ips);
```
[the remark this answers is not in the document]
Siflorite: `rx_staged_ips_apply` moved under tag `clear_rx_target`.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,1060 @@
...
+    mdns_service_cache_t **old_entry_cache = &old_entry->service_cache_list;
+
+    while (*old_entry_cache) {
+        if (*old_entry_cache == cache) {
+            *old_entry_cache = cache->next;
+            cache->next = new_entry->service_cache_list;
+            new_entry->service_cache_list = cache;
+
+            cache_remove_entry_if_empty(old_entry);
+            return true;
```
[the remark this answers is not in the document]
Siflorite: In current PR where browse is the only consumer of cache, this shall be fixed since browses subscribe to specific services, so no services mean empty. Future ADDR resolvers will be new consumers and allow cache entry with only addresses and no services. 

For now I'll only check services, but I need to mark it here and backport to current implementation when I have implemented ADDR resolver.

Siflorite:
Since PR #1110 is merged and #1125 is to merge, I will rebase this PR to put cache into conditional compilation.

components/mdns/mdns_browser.c
```diff
@@ -877,3 +353,95 @@ esp_err_t mdns_browse_delete(const char *service, const char *proto)
...
+    bool updated = true;
+
+    for (mdns_browse_t *browse = s_browse; browse; browse = browse->next) {
+        if (browse_matches_service_cache(browse, service)) {
+            updated &= browse_build_and_notify_temp_result(browse, entry, service, false);
+        }
+    }
+
+    return updated;
```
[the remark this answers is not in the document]
Siflorite: When receives a PTR TTL=0 record, it means that the entire service is removed. This is also mentioned in `mdns_priv_cache_update_ptr()`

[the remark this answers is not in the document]
david-cermak: I think the bugbot has a point here, it's not about real PTR goodbyes, but about `ttl == 0` because `ptr_ttl` was never set -- see the other link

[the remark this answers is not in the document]
Siflorite: > I think the bugbot has a point here, it's not about real PTR goodbyes, but about `ttl == 0` because `ptr_ttl` was never set -- see the other link

Thanks, I got that point. If SRV and TXT record comes earlier than a PTR record, they will fill in a service cache with not PTR record (ptr_present = false && ptr_ttl = 0). When the parsing is finished, `mdns_priv_cache_process_sync()` -> `mdns_priv_browse_update_from_service_cache()`-> `browse_build_and_notify_temp_result()` -> `mdns_priv_service_cache_to_result()` will generate a `mdns_result_t` with TTL=0, which will be seen as a goodbye by the user.

The solution is to check `ptr_present` in `browse_matches_service()`, which is only used by notifier-related functions.

david-cermak (commented):
I have these 3 high level comments, which could be also perceived as (behavioural) breaking changes. Let's just discuss now -- I'm open to suggestion from just documenting the change (or correcting me), making it more/less compatible up to creating v2.0+ release.

1) Temporary result
**Before**: result was a live node in the browse’s internal list. It stayed valid after the callback (except goodbye, which could free it on return).
**Now**: Projection is freed as soon as the notifier returns.

I like this change, but think of users and how they need to adopt their apps -- e.g. making deep-copies from the callback...

2) result->next always NULL
**Before**: result (goodbye) batches could do like `for (r = result; r; r = r->next)` in the notifier.
**Now**: next is always NULL; one callback = one instance.

I think this is a minor change a could be treatet it as a bugfix...

3) ttl is PTR TTL (goodbye = 0)
**Before**: ttl was the minimum TTL across records (PTR/SRV/TXT/A…)
**Now**: Browse ttl is PTR TTL only; removal is still ttl == 0.

Siflorite:
> I have these 3 high level comments, which could be also perceived as (behavioural) breaking changes. Let's just discuss now -- I'm open to suggestion from just documenting the change (or correcting me), making it more/less compatible up to creating v2.0+ release.
> 
>     1. Temporary result
>        **Before**: result was a live node in the browse’s internal list. It stayed valid after the callback (except goodbye, which could free it on return).
>        **Now**: Projection is freed as soon as the notifier returns.
> 
> 
> I like this change, but think of users and how they need to adopt their apps -- e.g. making deep-copies from the callback...
> 
>     2. result->next always NULL
>        **Before**: result (goodbye) batches could do like `for (r = result; r; r = r->next)` in the notifier.
>        **Now**: next is always NULL; one callback = one instance.
> 
> 
> I think this is a minor change a could be treatet it as a bugfix...
> 
>     3. ttl is PTR TTL (goodbye = 0)
>        **Before**: ttl was the minimum TTL across records (PTR/SRV/TXT/A…)
>        **Now**: Browse ttl is PTR TTL only; removal is still ttl == 0.

Thank you for review @david-cermak ! These behavioural changes are (somehow intentionally) acknowledged and kept during development.

1. Temporary result and 2. `result->next` always `NULL`

Since all cached data are now moved to `mdns_cache`, and all browse instances will be notified, there is no need to keep a result chain. In old `mdns_notify_t` annotation, it is required to copy `result` if it needs to be handled outside the callback.

```c
/**
 * @brief Browse result change notifier
 *
 * Called once per browse result that changed in a given response packet (not
 * once per packet). The @p result argument points at the changed entry, but it
 * remains a live node in the internal browse cache until the callback returns.
 *
 * @warning @p result->next links other cached instances for this browse, not
 *          necessarily other results that changed in the same packet. For
 *          ordinary add/update notifications, use only @p result; do not walk
 *          @c next, because unchanged instances may appear there.
 *
 * @warning Batch PTR TTL=0 ("goodbye") responses (for example when a peer
 *          exits and Bonjour sends many removals in one packet) may invoke the
 *          notifier once while several instances in the cache are updated.
 *          Until this is improved, applications may walk @c next and treat
 *          entries with @c ttl == 0 as removals. Do not assume every node on
 *          @c next changed in the current packet.
 *
 * @warning If handling is deferred outside this callback, copy @p result first
 *          (including strings and addresses). Goodbye entries may be freed when
 *          the callback returns.
 *
 * @warning This callback runs in the mDNS service task and holds the mDNS service lock.
 *          Users must not call APIs that acquire mDNS service lock in this callback.
 *          For example, mdns_browse_new() and mdns_browse_delete().
 *
 * @param result  The browse result that changed. See the warnings above for
 *                use of @c result->next.
 */
typedef void (*mdns_browse_notify_t)(mdns_result_t *result);
```

The new notifier is intended to free the result when the callback returns, so that users can keep their old deep copy implementations without memory leak. Using result outside the callback in old version will cause data race and should not be encouraged.

And the only scenario that users need to traverse result linked list is to track batch TTL=0 responses. Since all browses can be notified independently now, users can receive all TTL=0 responses without traverse, and old traverse code just does nothing as `next` in `result` is always NULL.

3. `ttl` now represents PTR TTL

I think the original `ttl` is a bit ambiguous, since it does not represent any specific TTL. If users need to get the minimum TTL of all records, I think a better way is to query (or use future resolvers) all records to get specific TTLs. But I'm not sure if users rely on this original minimum ttl in browse results. If it is not appropriate to change ttl semantic here, I will make an independent PTR resolver in the future.

I don't think it is necessary to bump to v2, since this browser change is designed to make as few behavioural changes as possible. When v2 is about to release, we can make it more straightforward, e.g. filling up missing components in `mdns_result_t`, transferring the ownership of `mdns_result_t` to users (to avoid unnecessary data clone), etc.

components/mdns/mdns_browser.c
```diff
@@ -206,15 +188,14 @@ static mdns_browse_t *browse_init(const char *service, const char *proto, mdns_b
  */
 static void browse_start(mdns_browse_t *browse)
 {
+    mdns_priv_cache_notify_browse(browse);
```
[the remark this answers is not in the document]
Siflorite: This is a severe issue and involves several changes:
Now browses are set state = `BROWSE_INIT` in `mdns_browse_new()`, and updated to `BROWSE_RUNNING` in `mdns_browse_start()`.
Accordingly, `mdns_browse_new()` and `mdns_browse_delete()` should check `state != BROWSE_OFF` when checking duplicated browses. So that when users call `mdns_browse_new()` -> `mdns_browse_new()` or `mdns_browse_new()` -> `mdns_browse_delete()`  -> `mdns_browse_new()`, it will not generate duplicated browses, or fails to delete browses.
Also, for the call chain: 

`browse_finish()` -> 
`mdns_priv_cache_remove_service_cache_if_unused()` -> 
`mdns_priv_browse_has_service()`

The check in `mdns_priv_browse_has_service()` should also be changed to `state != BROWSE_OFF`, so that it will not clear the cache corresponding to a browse waiting for boot up.

components/mdns/mdns_receive.c
```diff
@@ -874,13 +1000,9 @@ static void mdns_parse_packet(mdns_rx_packet_t *packet)
...
-                    out_sync_browse = mdns_priv_browse_ensure_sync(browse_for_ptr, out_sync_browse);
-                    if (!out_sync_browse) {
-                        goto clear_rx_packet;
-                    }
-                    mdns_priv_browse_result_add_ptr(browse_for_ptr, name->host, browse_for_ptr->service,
-                                                    browse_for_ptr->proto, packet->tcpip_if, packet->ip_protocol,
-                                                    ttl, out_sync_browse);
+                    (void)mdns_priv_cache_update_ptr(mdns_priv_get_esp_netif(packet->tcpip_if), packet->ip_protocol,
+                                                     name->host, browse_for_ptr->service, browse_for_ptr->proto,
+                                                     ttl);
```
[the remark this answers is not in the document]
Siflorite: I didn't move the original check in `mdns_priv_browse_result_add_ptr()` to my new function. Now fixed in `mdns_priv_cache_update_ptr()`.

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,980 @@
...
+        }
+        result->port = service->port;
+    }
+
+    if (service->txt_present) {
+        ESP_GOTO_ON_FALSE(project_txt(service->txt_list, &result->txt, &result->txt_value_len, &result->txt_count),
+                          ESP_ERR_NO_MEM, error, TAG, "Failed to project TXT");
+    }
+    ESP_GOTO_ON_FALSE(project_addr(entry->addr_list, &result->addr), ESP_ERR_NO_MEM, error, TAG,
+                      "Failed to project address list");
```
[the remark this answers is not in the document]
Siflorite: When SRV is no longer available, we should not provide addresses in notified result, since address is bound to hostname. Fixed in `mdns_priv_service_cache_to_result()`

david-cermak:
@Siflorite 

> These behavioural changes are (somehow intentionally) acknowledged and kept during development.


Thanks for explaining and confirming. I'll take one deeper dive into this and hopefully merge this later today (if everything goes smooth)
I agree that we can keep these changes and treat them as improvements in minor version bump.

david-cermak:
@Siflorite Please fix the CI issues: https://github.com/espressif/esp-protocols/actions/runs/33378991086/job/99446787791?pr=1121

david-cermak:
> @Siflorite Please fix the CI issues: https://github.com/espressif/esp-protocols/actions/runs/33378991086/job/99446787791?pr=1121

About the Fuzzer issues, pls fix the harness (it's not a bug in your code)
```patch
b/components/mdns/tests/host_unit_test/stubs/mdns_mem_caps.c
index 7da2681a..814cbff0 100644
--- a/components/mdns/tests/host_unit_test/stubs/mdns_mem_caps.c
+++ b/components/mdns/tests/host_unit_test/stubs/mdns_mem_caps.c
@@ -44,10 +44,7 @@ char *mdns_mem_strndup(const char *s, size_t n)
         return NULL;
     }
 
-    size_t len = strlen(s);
-    if (len > n) {
-        len = n;
-    }
+    size_t len = strnlen(s, n);
```

(Or I'll fix it separately)

Siflorite:
> About the Fuzzer issues, pls fix the harness (it's not a bug in your code)

@david-cermak I have fixed the stub and bugbot issue. Should I squash the first 4 commits? Since they are taken apart for review, I cannot ensure that every single commit functions.

david-cermak:
@Siflorite 

>  Should I squash the first 4 commits?

Yes please, all are related to one feature.

Siflorite:
@david-cermak Done, PTAL

components/mdns/mdns_cache.c
```diff
@@ -0,0 +1,976 @@
...
+
+void mdns_priv_cache_process_sync(void)
+{
+    for (mdns_cache_entry_t *entry = s_cache; entry; entry = entry->next) {
+        for (mdns_service_cache_t *service = entry->service_cache_list; service; service = service->next) {
+            if (service->sync_consumers & MDNS_CACHE_CONSUMER_BROWSE) {
+                mdns_cache_record_mask_t records = service->sync_records;
+                if (mdns_priv_browse_update_from_service_cache(entry, service, records)) {
+                    service_cache_clear_sync_out(service, MDNS_CACHE_CONSUMER_BROWSE, 0);
+                }
```
david-cermak: Similar comment to cursor, here.

it's safer IMO to just log and error and always proceed with clearing...
since if this fails one, it'd be kept dirty...

```suggestion
                if (!mdns_priv_browse_update_from_service_cache(entry, service, records)) {
                    ESP_LOGW(TAG, "Browse notify failed, dropping this sync");
                }
                // Always clear...
                service_cache_clear_sync_out(service, MDNS_CACHE_CONSUMER_BROWSE, 0);
```

david-cermak (commented):
LGTM otherwise

components/mdns/mdns_cache.c
Siflorite: The call chain here is:
`mdns_priv_browse_update_from_service_cache()` ->
`browse_build_and_notify_temp_result()` ->
`mdns_priv_service_cache_to_result()`

The only error that `mdns_priv_service_cache_to_result()` will return is `ESP_ERR_NO_MEM`, when fails to allocate result and its components. I kept this mark so that when memory is freed later, the cache that has been updated can still notify browses.

Is it better to clear the mark when fails to notify, or add a warning log other than `HOOK_MALLOC_FAILED` in `mdns_priv_service_cache_to_result()`?

david-cermak: > Is it better to clear the mark when fails to notify

Yes, I think so, since we give the system a means to recover a bit dropping this sync.

But I understand your reasoning behind the current code (which makes sense too, logically, or if we add some other error code in future)

I don't have a strong opinion here though (if we're out of memory, it's a failure mode already, the only thing is that we mustn't crash, but it's hard to guess what's better -- but dropping a queue of stacking future allocations seems ok to me...)

Siflorite: I got that point, actually current implementation is also not perfect: when future subtype browses are implemented, there may be like 5 browses subscribing to the same service.proto (differ in subtypes). If the memory is not enough, maybe only the first few browses are notified continuously, with the rest not notified and the mark not cleaned, and the program will keep allocating results.

I will always clean the mark, and output a warning log on OOM.

david-cermak:
Thanks for the quick update, this LGTM now.

Please add a few tests for the browse feature now (in the followup PR), before adding subtype support

david-cermak (approved):

PR #1127, opened 2026-08-12
feat(mdns): post event when hostname changes

zwx1995esp:
## Summary

- Add the public `mdns_hostname_changed_cb_t` callback type and `mdns_register_hostname_changed_callback()` API.
- Notify registered callbacks after the hostname changes, including automatic hostname conflict resolution.
- Support multiple callback registrations and ignore duplicate `(cb, arg)` registrations.
- Keep hostname-change notification independent of the system event loop; callbacks receive the new hostname directly and must not call mDNS APIs.

## Tests

- Add Linux host-unit coverage for callback registration, initial hostname assignment, unchanged hostname, and subsequent hostname changes.
- Add ESP target unit coverage for invalid registration arguments and hostname-change notification behavior.

## Validation

- Linux host-unit `test_receiver`: 4 tests passed, including `test_hostname_changed_callback`.
- Built and flashed the M5Stack Thread Border Router example with the local mDNS component. The example now uses the callback API, and hostname updates were confirmed on the connected boards.

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -171,6 +171,21 @@ typedef void (*mdns_query_notify_t)(mdns_search_once_t *search);
 typedef void (*mdns_browse_notify_t)(mdns_result_t *result);
 
 /**
+ * @brief Hostname change notifier
+ *
+ * Called from the mDNS service task after the hostname changes.
+ * The @p hostname argument is the new hostname. It remains valid until the
+ * callback returns.
+ *
+ * @warning Do not call mDNS APIs from this callback. The callback runs while
+ *          the mDNS service is processing an action.
+ *
+ * @param hostname  The new responder hostname.
+ * @param arg       User context supplied when registering the callback.
+ */
+typedef void (*mdns_hostname_changed_cb_t)(const char *hostname, void *arg);
+
+/**
  * @brief  Initialize mDNS on given interface
  *
  * @return
@@ -199,6 +214,24 @@ void mdns_free(void);
  *     - ESP_ERR_NO_MEM memory error
  */
 esp_err_t mdns_hostname_set(const char *hostname);
+
+/**
+ * @brief Register a callback notified when the hostname changes
+ *
+ * Multiple callbacks can be registered. Registering the same callback and
+ * context more than once has no effect. Callbacks cannot be unregistered and
+ * remain registered until mdns_free() is called.
+ *
+ * @param cb   Callback to invoke when the hostname changes.
+ * @param arg  User context passed to @p cb.
+ *
+ * @return
+ *     - ESP_OK on success
+ *     - ESP_ERR_INVALID_STATE when mDNS is not initialized
+ *     - ESP_ERR_INVALID_ARG when @p cb is NULL
+ *     - ESP_ERR_NO_MEM when the callback cannot be registered
+ */
+esp_err_t mdns_register_hostname_changed_callback(mdns_hostname_changed_cb_t cb, void *arg);
 
 /**
  * @brief Get the hostname for mDNS server
--- a/components/mdns/mdns_responder.c
+++ b/components/mdns/mdns_responder.c
@@ -16,8 +16,15 @@
 #include "mdns_pcb.h"
 #include "mdns_service.h"
 
+typedef struct mdns_hostname_changed_callback_s {
+    mdns_hostname_changed_cb_t cb;
+    void *arg;
+    struct mdns_hostname_changed_callback_s *next;
+} mdns_hostname_changed_callback_t;
+
 typedef struct mdns_server_s {
     const char *hostname;
+    mdns_hostname_changed_callback_t *hostname_changed_cbs;
     const char *instance;
     mdns_srv_item_t *services;
     mdns_host_item_t *host_list;
@@ -59,6 +66,17 @@ static void free_delegated_hostnames(void)
     s_server->host_list = NULL;
 }
 
+static void free_hostname_changed_callbacks(void)
+{
+    mdns_hostname_changed_callback_t *callback = s_server->hostname_changed_cbs;
+    while (callback != NULL) {
+        mdns_hostname_changed_callback_t *next = callback->next;
+        mdns_mem_free(callback);
+        callback = next;
+    }
+    s_server->hostname_changed_cbs = NULL;
+}
+
 void mdns_priv_responder_free(void)
 {
     if (s_server->action_sema != NULL) {
@@ -71,6 +89,7 @@ void mdns_priv_responder_free(void)
     mdns_mem_free((char *)s_server->hostname);
     mdns_mem_free((char *)s_server->instance);
     free_delegated_hostnames();
+    free_hostname_changed_callbacks();
     mdns_mem_free(s_server);
     s_server = NULL;
 }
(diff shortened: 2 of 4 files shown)


components/mdns/mdns_responder.c
```diff
@@ -98,11 +98,16 @@ mdns_host_item_t *mdns_priv_get_self_host(void)
...
     if (s_server) {
+        bool hostname_changed = s_server->hostname != hostname &&
+                                (!s_server->hostname || !hostname || strcmp(s_server->hostname, hostname) != 0);
         if (s_server->hostname) {
             mdns_mem_free((void *)s_server->hostname);
         }
         s_server->hostname = hostname;
         s_server->self_host.hostname = hostname;
+        if (hostname_changed) {
+            esp_event_post(MDNS_EVENT, MDNS_EVENT_HOSTNAME_CHANGED, NULL, 0, 0);
```
Siflorite: `esp_event_post()` returns an `esp_err_t` error code. Shall we print error log on failure or use `(void)` to explicitly discard return value?

components/mdns/mdns_responder.c
```diff
@@ -98,11 +98,19 @@ mdns_host_item_t *mdns_priv_get_self_host(void)
...
     if (s_server) {
+        bool hostname_changed = s_server->hostname != hostname &&
+                                (!s_server->hostname || !hostname || strcmp(s_server->hostname, hostname) != 0);
         if (s_server->hostname) {
             mdns_mem_free((void *)s_server->hostname);
         }
         s_server->hostname = hostname;
         s_server->self_host.hostname = hostname;
+        if (hostname_changed) {
+            esp_err_t err = esp_event_post(MDNS_EVENT, MDNS_EVENT_HOSTNAME_CHANGED, NULL, 0, 0);
```
david-cermak: This posts the mdns event into system event loop. I don't think we should have external components use the system loop here. I'll take a deeper look and think of an alternatives. maybe using private event loop or just callbacks for now (these are used in mdns anyway) and we'll add esp-event later as a feature version bump.

zwx1995esp: Hi @david-cermak  I see https://github.com/espressif/esp-protocols/blob/master/components/mdns/mdns_netif.c#L358 Some event handlers are registered when the network interface is initialized, so I thought we might be able to use the component `esp_event`..

david-cermak: Yes, we could use the component, just a bit differently.
The registered event handlers are consumers, that's fine as we're using *system* events and act upon it.
We could also *produce* events (just not via the system event loop), we just create our own/private event loop.
(by producing/posting into system event loop, we're kind of hijacking system resources, we shouln't do that)

david-cermak (changes requested):
LGTM overall, the only thing is posting to the system/defautl event_post

Let's rework it for now to callbacks, something like this:
```cpp
typedef void (*mdns_hostname_changed_cb_t)(const char *hostname, void *arg);
esp_err_t mdns_register_hostname_changed_callback(mdns_hostname_changed_cb_t cb, void *arg);
````
since this method is widely used alrady in the mdns component.

I'll think about the event loop usage and will probably create a private event loop that could be used instead of callbacks. BUt will do that in the next  release

**Comment by david-cermak** [comment]
@zwx1995esp Also please rebase and resolve conflicts

File: components/mdns/mdns_responder.c (hunk 2 above)
Comment (zwx1995esp) [reply components/mdns/mdns_responder.c]: Okay, got it. That makes sense to me.

**Comment by zwx1995esp** [comment]
> LGTM overall, the only thing is posting to the system/defautl event_post
> 
> Let's rework it for now to callbacks, something like this:
> 
> ```c++
> typedef void (*mdns_hostname_changed_cb_t)(const char *hostname, void *arg);
> esp_err_t mdns_register_hostname_changed_callback(mdns_hostname_changed_cb_t cb, void *arg);
> ```
> 

Hi @david-cermak Thanks for your reviewing. I’ll refactor the related logic and rewrite the test cases. Once everything is ready, I’ll push the updated changes and let you know.

**Comment by zwx1995esp** [comment]
@david-cermak Refactor done, PTAL.

File: components/mdns/mdns_responder.c [hunk 3]
```diff
@@ -98,11 +117,19 @@ mdns_host_item_t *mdns_priv_get_self_host(void)
...
             mdns_mem_free((void *)s_server->hostname);
         }
         s_server->hostname = hostname;
         s_server->self_host.hostname = hostname;
+        if (hostname_changed) {
+            for (mdns_hostname_changed_callback_t *callback = s_server->hostname_changed_cbs;
+                    callback != NULL; callback = callback->next) {
+                callback->cb(hostname, callback->arg);
+            }
+        }
```
david-cermak, line 132: LGMT, just one nitpick -- should we service-lock also here? (as we do with the browse notifiers)

components/mdns/include/mdns.h
```diff
@@ -200,6 +215,23 @@ void mdns_free(void);
  */
 esp_err_t mdns_hostname_set(const char *hostname);
 
+/**
+ * @brief Register a callback notified when the hostname changes
+ *
+ * Multiple callbacks can be registered. Registering the same callback and
+ * context more than once has no effect.
```
david-cermak: Maybe just note that the callbacks cannot be unregistered

david-cermak (commented):
LGTM otherwise

components/mdns/mdns_responder.c
zwx1995esp: Hi David, do you mean that we need to add the service-lock like:
```
MDNS_SERVICE_LOCK();  // add lock acquiring here
if (hostname_changed) {
      for (mdns_hostname_changed_callback_t *callback = s_server->hostname_changed_cbs;
              callback != NULL; callback = callback->next) {
          callback->cb(hostname, callback->arg);
     }
}
MDNS_SERVICE_UNLOCK(); // and release lock here.
```
If I did not misunderstand, I don't think we should do like this. There are two scenarios where we will call `mdns_priv_set_global_hostname`

1. Application sets the hostname
```
  mdns_hostname_set()
    -> queues ACTION_HOSTNAME_SET
    -> service_task()
        -> MDNS_SERVICE_LOCK()
        -> execute_action()
            -> mdns_priv_responder_action()
                -> mdns_priv_set_global_hostname()
        -> MDNS_SERVICE_UNLOCK()
```
2.Hostname collision (SRV, A, or AAAA record)
```
  queues ACTION_RX_HANDLE
    -> service_task() [service lock held]
        -> mdns_priv_receive_action()
            -> mdns_parse_packet()
            -> detects an SRV, A, or AAAA collision
                -> mdns_priv_set_global_hostname()
```
In both scenario, we take and give the lock in the upper functions. Since the service-lock is not a recursive mutex, we can not take the lock twice.

I also check the browser notifier, only one place the `browse_sync` called. And we take the lock in the function `service_task`.
```
mdns_priv_browse_sync()
    -> queues ACTION_BROWSE_SYNC
        -> service_task() [service lock held]
            -> execute_action()
                -> mdns_priv_browse_action()
                    -> browse_sync()
                        -> browse->notifier(result)
```

components/mdns/include/mdns.h
zwx1995esp: Done.

components/mdns/mdns_responder.c
david-cermak: I'm sorry, you're right it's been locked already.

david-cermak (approved):

PR #1130, opened 2026-08-19
fix(mdns): handle subtype questions correctly

zwx1995esp:
## Background

A user reported that an `ANY` query containing a DNS-SD subtype could be handled as an instance query. For example, when parsing `_aa._sub._print._udp.local`, the parser stores `_aa` in `host`; subsequent code could therefore mistake it for an instance name.

We reviewed the related subtype handling paths and fixed the issues below.

## Fixes

1. Distinguish subtype `ANY` questions from instance `ANY` questions, so subtype questions are answered without being marked as probes.
2. Ignore `SRV` and `TXT` questions at a subtype owner name. DNS-SD subtypes define `PTR` records at that name only.
3. Share the mDNS debug-name formatting path so record-owner output consistently retains `_sub`.
4. Validate that `_sub` appears only in the DNS-SD subtype position: `<subtype>._sub.<service>.<proto>.<domain>`.

## Tests

Added coverage for:

- A subtype `ANY` question, ensuring it is parsed as a subtype and is not marked as a probe.
- Rejection of a misplaced `_sub` label, while accepting a valid subtype query name.

Verified with the Linux host receiver and sender unit-test suites.

### DIFF
--- a/components/mdns/mdns_debug.c
+++ b/components/mdns/mdns_debug.c
@@ -103,6 +103,38 @@ static void mdns_dbg_printf(const char *fmt, ...)
 #define mdns_dbg_flush()
 #endif
 
+static void dbg_print_name_and_type(const char *prefix, const mdns_name_t *name, uint16_t type)
+{
+    if (type == MDNS_TYPE_A) {
+        dbg_printf("%s%s.%s. A ", prefix, name->host, name->domain);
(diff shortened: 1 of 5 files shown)


david-cermak (approved):
Thanks for the fixes, LGTM!

PR #1153, opened 2026-09-21
feat(mdns): Let mdns_cache expire natrually by TTL

Siflorite:
## Description

Currently the lifetime of cache is bind to consumers: a cache entry is created when receiving a packet matching existing consumers, and removed when receives a TTL=0 or the consumer is deleted. But if users quickly create-delete-create a browse for a service, the old cache is removed when it's still valid and browse need to post a new query, and browses may not get the records notified in time.

This PR implements a natural expiry feature to mdns_cache. The cache will only be removed when:
1. Receives a TTL=0 record matching existing consumers;
2. Cache expires by TTL.

Each record will hold a `mdns_cache_expiry_t` node arranged by expiration time, with head at `s_expiry`. An expiration time flush will rearrange the node, and TTL=0 caches will be moved to head of the chain.

`mdns_service` will check if there are expired records every second, clean up records, and execute `process_sync()`. In this way the lifetime of a cached record is only determined by its TTL.

A stub is added to host unit test to provide `esp_timer_get_time()` for compilation and minimal clocking functions. No tests are added in this PR.

It is worth mentioning that this PR only implements basic cache expiration functions, without the re-query mechanism of a record about to expire suggested by [RFC 6762 §5.2](https://datatracker.ietf.org/doc/html/rfc6762#section-5.2), to prevent being too complicated.

## Testing

The new cache expiry is tested with mdns_browse and functions well.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes browse discovery caching, expiration timing, and notification paths; mis-timed expiry or sync could cause stale services or missed goodbyes, but scope is limited to browse-enabled builds.
> 
> **Overview**
> **mDNS browse cache entries now live until their record TTLs expire**, instead of being torn down when the last browse stops or when TTL=0 is received inline.
> 
> The cache layer adds a **time-ordered expiry queue** (`mdns_cache_expiry_t`) for PTR, SRV, TXT, and A/AAAA records. TTL=0 and natural expiration **schedule removal** rather than deleting service caches immediately; `mdns_priv_cache_remove_expired_records()` performs the actual cleanup, sends PTR goodbyes to browsers, and marks remaining records for sync. Browse teardown **no longer** calls `mdns_priv_cache_remove_service_cache_if_unused`, and `mdns_priv_browse_has_service` only treats **running** browses as subscribers.
> 
> **When expiry and browse sync run** moves to the service task: the mDNS task wakes every **1s** (when browse is enabled) to expire stale cache and run `mdns_priv_cache_process_sync()`; after each RX packet parse, expiry/sync run **after** cache updates. `mdns_priv_cache_process_sync()` is gated on a pending flag. Address updates for existing host entries apply only when an active browse still cares about a service on that host.
> 
> Host unit tests gain a controllable **`esp_timer_get_time()`** stub (`mdns_test_clock.h`) for future TTL tests. `mdns_priv_browse_send_all` is removed (per-protocol browse send remains).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f39768e275fe7a829baa8975f4d6e24d89dd4f0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


components/mdns/mdns_cache.c
```diff
@@ -512,50 +537,65 @@ mdns_cache_update_result_t mdns_priv_cache_update_srv(const esp_netif_t *esp_net
...
     } else if (owner_entry != host_entry) {
         if (!cache_move_service(owner_entry, host_entry, service_entry)) {
+            if (new_srv_expiry) {
+                mdns_mem_free(srv_expiry);
+            }
+            cache_remove_entry_if_empty(host_entry);
             return MDNS_CACHE_ERROR;
         }
         result = MDNS_CACHE_UPDATED;
     }
```
zwx1995esp, line 611: Could we move the `srv_expiry` ceating logic here, then we can remove the local variable `new_srv_expiry`. The only issue maybe `cache_move_service`, we can move back if the `srv_expiry` can not be allocated.

components/mdns/mdns_service.c
```diff
@@ -221,17 +225,39 @@ static void execute_action(mdns_action_t *action)
...
+                mdns_priv_cache_remove_expired_records(esp_timer_get_time());
+                mdns_priv_cache_process_sync();
+            }
+#endif
+            if (received == pdTRUE) {
                 execute_action(a);
-                MDNS_SERVICE_UNLOCK();
             }
+
+            MDNS_SERVICE_UNLOCK();
```
zwx1995esp, line 252: We can remove these calls from the file `components/mdns/mdns_receive.c` and handle the cache entry expiration entirely in the main mdns loop.
```
mdns_priv_cache_remove_expired_records(esp_timer_get_time());
mdns_priv_cache_process_sync();
```
```suggestion
            BaseType_t received = xQueueReceive(s_action_queue, &a, wait_ticks);
            if (received == pdTRUE) {
                assert(a);
                if (a->type == ACTION_TASK_STOP) {
                    break;
                }
            }

            MDNS_SERVICE_LOCK();
#ifdef CONFIG_MDNS_ENABLE_BROWSE
            // Parse the packet first so its records can refresh TTLs in the cache entry.
            // If a record has TTL=0, parsing sets its expiration time to 0, 
            // and the cache cleanup after parsing removes it.
            if (received == pdTRUE && a->type == ACTION_RX_HANDLE) {
                execute_action(a);
            }
            mdns_priv_cache_remove_expired_records(esp_timer_get_time());
            mdns_priv_cache_process_sync();
#endif
            if (received == pdTRUE) {
                execute_action(a);
            }
            MDNS_SERVICE_UNLOCK();
```

components/mdns/mdns_service.c
```diff
@@ -43,6 +43,10 @@
 #define MDNS_TASK_USE_SIMPLE_CREATE 0
 #endif
 
+#ifdef CONFIG_MDNS_ENABLE_BROWSE
+#define MDNS_CACHE_SCAN_INTERVAL_MS 1000
```
zwx1995esp:
```suggestion
#define MDNS_CACHE_EXPIRY_CHECK_TIMEOUT_MS 1000
```

components/mdns/mdns_cache.c
```diff
@@ -781,7 +816,125 @@ mdns_cache_update_result_t mdns_priv_cache_update_addr(const esp_netif_t *esp_ne
...
+void mdns_priv_cache_remove_expired_records(int64_t now_us)
+{
+    if (!s_expiry || s_expiry->expires_at_us > now_us) {
+        return;
+    }
+
+    mdns_cache_expiry_t *head = s_expiry;
+    mdns_cache_expiry_t *end = s_expiry;
+    mdns_cache_expiry_t *merged = NULL;
+    while (end->next && end->next->expires_at_us <= now_us) {
```
zwx1995esp: We currently traverse the expiry list up to three times to handle expired entries. Could we simplify this to a single pass, processing each expired node as we go and stopping at the first unexpired node?

components/mdns/mdns_cache.c
Siflorite: The logic here is re-designed: `srv_expiry` is allocated after settling down entry/service creation and service move. Now if `srv_expiry` fails to allocate, the service cache will be moved back to `owner_entry`, so as to keep `srv_expiry` aligned with service cache item.

components/mdns/mdns_service.c
Siflorite: I find it unnecessary to remove expired records and process sync before non-rx events, since records without stale records will finally notify browses. Now these two processes will be executed after `execute_action()`. Also the two functions in `mdns_receive` are ripped off.

components/mdns/mdns_service.c
Siflorite: Fixed. That's fair: the actual situation is that the maximum time to check expiry is 1000ms, `TIMEOUT` is more adequate than `INTERVAL`.

components/mdns/mdns_cache.c
Siflorite: This part is refactored, now cache records are processed in ONE traverse. The only behavioral change is that notified `result` may contain uncleaned SRV/TXT/ADDR records on PTR TTL=0, but I think ttl=0 means the service is no longer available, so that all other fields should be considered as deprecated.

components/mdns/mdns_browser.c
```diff
@@ -122,7 +115,7 @@ static bool browse_matches_identity(const mdns_browse_t *browse, const char *ser
 bool mdns_priv_browse_has_service(const char *service, const char *proto)
 {
     for (const mdns_browse_t *it = s_browse; it; it = it->next) {
-        if (it->state != BROWSE_OFF && browse_matches_identity(it, service, proto)) {
```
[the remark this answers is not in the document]
Siflorite: Binding cache update to cache items rather than consumers may make the update process too complicated, I prefer the current implementation.

components/mdns/mdns_cache.c
```diff
@@ -714,22 +760,11 @@ static mdns_cache_update_result_t cache_update_addr(const esp_netif_t *esp_netif
...
-                for (mdns_service_cache_t *service = entry->service_cache_list; service; service = service->next) {
-                    service_cache_mark_sync_out(service, MDNS_CACHE_UPDATED, MDNS_CACHE_RECORD_ADDR, MDNS_CACHE_CONSUMER_BROWSE);
-                }
-
-                cache_remove_entry_if_empty(entry);
-
+        for (mdns_cache_addr_t *it = entry->addr_list; it; it = it->next) {
+            if (addr_equal(&it->addr, addr)) {
+                expiry_schedule(&it->expiry, 0);
                 return MDNS_CACHE_REMOVED;
```
david-cermak: shouldn't we say "updated" after `expiry_schedule()` ?
```suggestion
                return MDNS_CACHE_UPDATED;
```

components/mdns/mdns_browser.c
```diff
@@ -141,7 +134,6 @@ static void browse_finish(mdns_browse_t *browse)
     for (mdns_browse_t *it = s_browse; it; it = it->next) {
         if (it == browse) {
             queueDetach(mdns_browse_t, s_browse, it);
-            mdns_priv_cache_remove_service_cache_if_unused(it->service, it->proto);
```
david-cermak, line 144: So now the services are active when we remove browsers? (until TTL even with no browsers)

components/mdns/mdns_cache.c
Siflorite: fixed

components/mdns/mdns_browser.c
Siflorite: Yes, now only receiving TTL=0 when the consumer is active and natrual expiration removes service cache.

components/mdns/tests/host_unit_test/stubs/mdns_test_clock.h
```diff
@@ -0,0 +1,21 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void mdns_test_clock_reset(void);
+
+bool mdns_test_clock_advance_us(int64_t delta_us);
```
david-cermak, line 17: This is good, but do you plan to add unit tests as well? Or in a followup PR? (I understand that you needed this to compile the existing tests)

Siflorite: I find that the original host unit tests on cache and browse will fail since I've modified some inner behaviors. I've fixed these issues, and decided to add one test for expiry extension and notify on expiry in `test_cache` while I'm at it.

Siflorite: I have added a test `test_cache_expiry_entend_and_notify` in `test_cache` for the new expiry feature, utilising the stubs.

Siflorite:
Hi @david-cermak we decide to move the expiration nodes out of cache and will make some modifications later.

components/mdns/private_include/mdns_private.h
```diff
@@ -428,16 +444,16 @@ typedef struct mdns_service_cache_s {
     char *service;
     char *proto;
     // PTR
-    bool ptr_present;   /*!< true if PTR record is present */
+    bool ptr_present;                           /*!< true if PTR record is present */
```
zwx1995esp, line 447: Revert these three unexpected changes?

components/mdns/mdns_cache.c
```diff
@@ -27,6 +30,97 @@ static inline bool nullable_names_equal(const char *a, const char *b)
...
+        }
+        link = &it->next;
+    }
+
+    if (!node) {
+        ESP_LOGE(TAG, "Cannot find expiry node for cache record");
+        return;
+    }
+
+    switch (node->record_mask) {
```
zwx1995esp, line 90: Can we use the variable passed in this function directly? I think here `record_mask` is the same as `node->record_mask`.
```suggestion
    switch (record_mask) {
```
Also use `service` instead of `node->service`

zwx1995esp (approved):
LGTM overall.

Siflorite:
@david-cermak Ready to merge now, PTAL.

